PR #5, opened 2021-09-17
cargo-lock [wip]

MatthewCroughan:


src/translators/rust/pure/cargo-lock/default.nix
```diff
@@ -0,0 +1,50 @@
...
+}:
+
+let
+  translate =
+    {
+      inputPaths,
+      ...
+    }:
+    let
+      cargoLock = builtins.fromTOML (builtins.readFile (builtins.elemAt inputPaths 0));
```
happysalada, line 17: @DavHau just thinking out loud here, but I'm thinking that it might make sense to have inputPaths be a dictionary with name keys ? (rather than a list). Relying on the position of an attribute could be error prone ?
Just a thought.

src/apps/cli.py
```diff
@@ -13,7 +13,8 @@
 
 def strip_hashes_from_lock(lock):
   for source in lock['sources'].values():
-    del source['hash']
+    if 'hash' in source:
```
happysalada, line 16: I think dave just merged that bit

src/translators/rust/pure/cargo-lock/default.nix
```diff
@@ -0,0 +1,50 @@
...
+      parsed = builtins.listToAttrs (map (p: lib.nameValuePair p.name p) cargoLock.package);
```
happysalada, line 18: if you are interested in some of the parsing that nixpkgs does on the cargolock, there was recently a PR about some improvements to it. You might be interested.
https://github.com/NixOS/nixpkgs/pull/137303/files

src/translators/rust/pure/cargo-lock/default.nix
DavHau: Actually I'm currently thinking about removing the path list and instead only allow to pass one path.
One reason is that pure translators should be callable via nix expression, and there it is simple to specify just one source.
If the source needs extra files, then it can be patched inside the nix expression before handing it to the translator

DavHau:
Taking a look at this repo here might be helpful as well: https://github.com/yusdacra/nix-cargo-integration
It might have some bits and pieces we could re-use.

happysalada:
in the helix project it's used. It's very well made!
it integrates nicely all the different nix methods to build rust packages.
It's using crate2nix and buildRustPackage under the hood though.

yusdacra:
Hello, I just came across this and I must say I really like the idea! I would be glad to help with anything. Would love to utilize this in nix-cargo-integration, as I was planning to make crate2nix not use IFD, but this seems like a much better idea :)

DavHau:
@yusdacra We'd love to collaborate with you.
After working on this we discovered that there is [already a parser in nixpkgs](https://github.com/NixOS/nixpkgs/blob/1c1b567985bd1be77601657562ed20299d169529/pkgs/build-support/rust/import-cargo-lock.nix)
Maybe we can re-use parts of it.
If you like we can sync up via a quick call.

DavHau:
@yusdacra Are you interested in working on this?
Maybe it is best to start this from scratch. There is now `utils.simpleTranslate` which should make it simpler to implement a translator.
The flake app `contribute` can be used to generate a template for a new translator with lots of comments inside on how to implement it.

In case this needs some more clarification, we can continue discussing in: https://github.com/nix-community/dream2nix/issues/16

yusdacra:
> @yusdacra Are you interested in working on this?
> Maybe it is best to start this from scratch. There is now `utils.simpleTranslate` which should make it simpler to implement a translator.
> The flake app `contribute` can be used to generate a template for a new translator with lots of comments inside on how to implement it.
> 
> In case this needs some more clarification, we can continue discussing in: https://github.com/nix-community/dream2nix/issues/16

Yes, I would like to work on this. I have some time for this now so I can attempt building something. `contribute` looks nice, will try using it.

yusdacra:
This can be closed now that there is a Cargo lock translator merged.

MatthewCroughan:
@yusdacra And a very very good one at that! Thank you.

PR #11, opened 2021-10-05
Add yarn lock

happysalada:
initial PR on adding a yarn.lock parser.


L-as:
You should change the branch to merge into. @DavHau should probably change the default branch.

happysalada:
you're right, done.

DavHau:
I added some fixes to the dev branch. After testing it seems like the parser is not able to parse all yarn.lock files.
It fails for example on facebook/react. I traced the `parsed.value` attribute:
```
trace: { context = "parsec.eof"; msg = "expected end of input"; offset = <CODE>; str = <CODE>; }
```

Any idea whats going wrong @happysalada ?

happysalada:
just to give you more information, the location of the failure is in `offset`. The location is the position in the string so it's not super helpful, but usually, I remove from the top of the lock file until I get to the failure.
In this case, the failure is at 
```
"eslint-plugin-react-internal@link:./scripts/eslint-rules":
  version "0.0.0"
```

I've verified, and it's the only failure on that lock.
How do we parse this though ?

DavHau:
Ouh that is a link to a local directory. Same thing we discovered today in package-lock.json.
I think we could do the following:
  1. check if the referenced file exists in the upstream repo.
  2. if it does not exist, resolve the package name and version to a nodejs url (hash is still missing)
  3. force the user to override the source hash for that package later.

PR #24, opened 2021-10-13
WIP: Racket translator

thomassdk:
Putting this here to keep track of changes and to have a reference for future work (and all the questions ive littered the code with)

The progress so far is that the translator is creating a list of sources (only for github atm) and creates the dependency graph.

There is one annoying bug in the parser that needs to be fixed before it can be tested properly but ive run the functions on some test data in the repl and they appear to work correctly (with the caveats that are highlighted in the code)

@DavHau If you have time to review the code and that would be fantastic. 

Im aiming to get this translator done by the end of Summer of Nix... but lets see how that goes :smile:


DavHau:
Hey @thomassdk, how are things going here?
Let me know, if you could use some advice.

DavHau:
Ooops, I just re-read your issue description and saw that you already stated some problems there. I will have a look at it soon.

happysalada:
The translator api has changed a bit since you first started the implementation. Here is a more recent example https://github.com/nix-community/dream2nix/blob/main/src/translators/rust/pure/cargo-lock/default.nix
If you are still interested in this, I'm happy to do some pair programming and bring this up to speed. Let me know.

PR #26, opened 2021-10-17
add go translator using gomod2nix

humancalico:
Co-authored-by: DavHau <hsngrmpf+github@gmail.com>


flake.nix
```diff
@@ -1,9 +1,12 @@
 {
   description = "dream2nix: A generic framework for 2nix tools";
 
-  inputs = {
+  inputs = {
```
SuperSandro2000, line 4: Some extra whitespace sneaked in here

src/builders/go/gomod2nix/default.nix
```diff
@@ -0,0 +1,21 @@
...
+let
+  gomod2nixTOML = fetchedSources.mapAttrs
+    lib.replaceStrings ["__slash__"] ["/"] dependencyObject.goName;
+in
+externals.gomod2nixBuilder rec {
+  pname = dreamLock.generic.mainPackage;
+  version = dreamLock.sources."${pname}".version;
+  src = fetchedSources."${pname}";
+  modules = ./gomod2nix.toml;
+}
```
SuperSandro2000, line 21: You should add en extra new line here so that git can generate diffs for the last line.

src/translators/go/impure/gomod2nix/translate.nix
```diff
@@ -0,0 +1,66 @@
...
+          version = getVersion dependencyObject;
+          hash = dependencyObject.fetch.sha256;
+          url = dependencyObject.fetch.url;
+          rev = dependencyObject.fetch.rev;
+        };
+    };
+
+  };
+in
+  b.toJSON translated
```
SuperSandro2000, line 66: Same here

src/builders/go/gomod2nix/default.nix
```diff
@@ -0,0 +1,21 @@
...
+  ...
+}:
+
+{
+  fetchedSources,
+  dreamLock,
+}:
+let
+  gomod2nixTOML = fetchedSources.mapAttrs
+    lib.replaceStrings ["__slash__"] ["/"] dependencyObject.goName;
```
DavHau, line 14: Actually I think we should keep the original names here. It is better if the dream-lock contains the original names, because this will allow us to write override using the upstream names. This should prevent confusion for maintainers.
The builder can later use `utils.sanitizeDerivationName` when creating the derivations.

src/translators/go/impure/gomod2nix/translate.nix
DavHau: Please use `dream2nix.utils.dreamLock.toJSON` here instead, as this will do some optimizations on the outputted data, like compacting dependency tree entries, removing empty entries, etc.

src/translators/go/impure/gomod2nix/default.nix
```diff
@@ -0,0 +1,68 @@
...
+      # read the json input
+      outputFile=$(${jq}/bin/jq '.outputFile' -c -r $jsonInput)
+      inputDirectory=$(${jq}/bin/jq '.inputDirectories | .[0]' -c -r $jsonInput)
+
+      # TODO: run gomod2nix and read the gomod2nix.toml and convert it to dream.lock
+      tmpBuild=$(mktemp -d)
+      cd $tmpBuild
+      cp -r $inputDirectory/* .
+      chmod -R +w .
+      nix run github:tweag/gomod2nix
```
DavHau, line 39: Maybe we should pin the revision of gomod2nix.

DavHau (commented):
@humancalico The json schema verification complaining about the `hash` in git sources is already fixed on master. After a rebase the translator worked for me.

Now just the builder for go is missing, but we can do this in another PR.
If you apply the requests, I can merge it.

flake.nix
humancalico: :+1:

src/builders/go/gomod2nix/default.nix
humancalico: Ohh didn't about this. Thanks!

src/translators/go/impure/gomod2nix/translate.nix
humancalico: Done :+1:

src/translators/go/impure/gomod2nix/default.nix
humancalico: Changed it to `github:tweag/gomod2nix/67f22dd738d092c6ba88e420350ada0ed4992ae8` which is the version in the flake.lock. Is there a better way to do this?

src/builders/go/gomod2nix/default.nix
humancalico: Done!

humancalico:
@DavHau PTAL

DavHau:
Thanks a lot for this @humancalico. Next time please target `main` directly.

PR #50, opened 2021-11-11
nodejs builder/translator and CLI improvements

DavHau:
Done:
  - builder: just skip cyclic dependencies instead of delegating to combined builder
  - builder: fix bug: crash if `bin` field in package.json is empty string
  - package-lock translator: allow `name` to be specified through CLI

### DIFF
--- a/src/apps/cli/commands/add.py
+++ b/src/apps/cli/commands/add.py
@@ -451,4 +451,4 @@ class AddCommand(Command):
     print(f"Created {output}/dream-lock.json")
 
     if config['isRepo']:
-      sp.run(f"git add -N {output}")
+      sp.run(f"git add -N {output}".split())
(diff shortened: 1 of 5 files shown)


src/apps/cli/commands/add.py
```diff
@@ -448,4 +448,7 @@ def update_name(mainPackageDirName):
         defaultNix.write(template)
         print(f"Created {output}/default.nix")
 
-    print(f"Created {output}/dream-lock.json")
\ No newline at end of file
+    print(f"Created {output}/dream-lock.json")
+
+    if config['isRepo']:
+      sp.run(f"git add -N {output}".split())
```
Mic92, line 454: It would be better to pass a list when running commands to avoid quoting issues.

DavHau: Thanks. Will include this into my next PR

Issue #16, opened 2021-10-06
Create translator and builder for rust/cargo


DavHau:
Some work on the translator has already been started but is stalled: https://github.com/DavHau/dream2nix/pull/5
We might be able to re-use some parts of [import-cargo-lock.nix](https://github.com/NixOS/nixpkgs/blob/1c1b567985bd1be77601657562ed20299d169529/pkgs/build-support/rust/import-cargo-lock.nix) from nixpkgs.

happysalada:
One package that would be worth testing is meilisearch github.com/meilisearch/meiliSearch/
currently on nixpkgs, the update to the latest version fails because of a feature override that is needed.
(it doesn't work properly with buildRustPackage, but crate2nix has exactly the right features: granular build and overrides per dependencies.)
Ideally we should model dream2nix after what is done in crate2nix.

A bit more context on meilisearch: I'm the maintainer of it in nixpkgs. Originally I was using crate2nix to build it, but transitioned to a more classic buildRustPackage build to simplify updates. It worked for a bit but now it's problematic again to the point I couldn't do the last update.

DavHau:
Thanks for the info.
Just pinging you @yusdacra as this might be relevant to you.

yusdacra:
I think we can leverage the Cargo.lock / Cargo.toml parsing from `naersk` for the translator and the "builder" from `crate2nix` (granular builds).

yusdacra:
I have started working on the translator here: https://github.com/yusdacra/dream2nix/tree/feat/cargo-translator

My main questions now are:
- What exactly is a "subsystem"? How does this relate to the building process? I see the `nodejs` `package.lock` translator having a `nodejs version` option (?). Shouldn't these be specified in a builder?
- How does `dream2nix` make use of `serializePackages`? The template says it can be "a list of package objects of arbitrary structure", so this only concerns builders that will use the translator? How does it relate to "dream lock"?
- And i guess i would like an overall process of how things go from parsing to building (translator -> builder? what are the steps in between of these?) :slightly_smiling_face:

DavHau:
> And i guess i would like an overall process of how things go from parsing to building (translator -> builder? what are the steps in between of these?)

See this little chart for an [overview of the architecture](https://github.com/nix-community/dream2nix/blob/main/docs/concepts-and-benefits.md#architecture). There is also an example walking through the steps.

[...]

PR #59, opened 2021-11-27
feat: cargo lock translator, rust builder

yusdacra:
Related to #16 

TODOs:
- [x] is the crates-io fetcher okay as is? should decompression also be done there?
- [x] prioritize binary packages while choosing a package to automatically build


src/fetchers/crates-io/default.nix
```diff
@@ -0,0 +1,36 @@
...
+  utils,
+  ...
+}:
+{
+
+  inputs = [
+    "pname"
+    "version"
+    "hash"
+  ];
```
DavHau, line 14:
```suggestion
    "version"
  ];
```

src/fetchers/crates-io/default.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  versionField = "version";
+
+  outputs = { pname, version, hash, ... }@inp:
```
DavHau, line 18:
```suggestion
  outputs = { pname, version, ... }@inp:
```

src/translators/rust/pure/cargo-lock/default.nix
```diff
@@ -0,0 +1,229 @@
...
+      inputFiles,
+    }@args:
+    {
+      inputDirectories = lib.filter 
+        (utils.containsMatchingFile [ ''.*Cargo\.lock'' ])
+        args.inputDirectories;
+      
+      inputFiles = lib.filter
+        (file: lib.hasSuffix "Cargo.toml" file)
+        args.inputFiles;
```
DavHau, line 211: Maybe set this to an empty list, as we ignore inputFiles currently.

DavHau (commented):
Thanks a lot for that.

Can't we just assume that the `Cargo.toml` is always at the root of the project?
Also as of now, you can assume to have only one input directory.

> is the crates-io fetcher okay as is? should decompression also be done there?

Yes, decompression should be done inside the fetcher.

> prioritize binary packages while choosing a package to automatically build

Not sure what exactly you mean by that. Maybe you could provide an example?

yusdacra:
> Not sure what exactly you mean by that. Maybe you could provide an example?

What i meant by that was to prioritize Cargo packages that have binaries. When there is multiple Cargo packages, usually most of them are libraries and there is one or two binary packages. Pretty much everyone will want the binary package, so it should prioritize choosing those over library packages (which i pushed now!).

> Yes, decompression should be done inside the fetcher.

I implemented this, though not sure if it's the right way. Would like a look at it :slightly_smiling_face:

DavHau:
You can test the creates-io fetcher by using it with the `add` command:
```
nix run . -- add crates-io:ripgrep/13.0.0
```

I think the tar unpack command should use `--strip-components 1`

DavHau:
To test the whole integration nicely we would also need to add a builder for rust. Do you have an idea what we could base this off?  
I'd assume we could just use `buildRustPackage` from nixpkgs and pass our dependencies via `cargoVendorDir`?

yusdacra:
> To test the whole integration nicely we would also need to add a builder for rust. Do you have an idea what we could base this off? I'd assume we could just use `buildRustPackage` from nixpkgs and pass our dependencies via `cargoVendorDir`?

Yeah, i think that would be the simplest way. I can implement a granular builder like crate2nix later on.

yusdacra:
I implemented a simple builder. It still misses some stuff (like exporting all Cargo packages, instead of only the main package). I'm not sure how to implement "packageOverrides", do i just implement it like how `node2nix` builder does it?

DavHau:
Cool! After rebasing your branch on main I could build ripgrep from `github:BurntSushi/ripgrep/13.0.0`.

To enable overriding, just wrap the `buildRustPackage` call in `makeDerivation` like seen in `src/builders/nodejs/granular/default.nix`

One problem: If I build any of the other contained packages, for example:
```
nix build -f ./ripgrep/default.nix packages.grep-cli
```
It fails with:
```
error: value is a string with context while a set was expected

       at /nix/store/7vak2420mdwdipjvbkznp4vibgz36dag-dream2nix-full-src/builders/rust/build-rust-package/default.nix:62:83:

           61|       postUnpack = ''
           62|         ln -s ${vendorPackageDependencies mainPackageName mainPackageVersion} ./${src.name}/nix-vendor
             |                                                                                   ^
           63|       '';
```

Also, could you add a test for the translator under `src/tests/impure`. You can just copy one of the existing ones and modify slightly.

yusdacra:
Hmm, that's weird. I rebased on main and now it complains about an error in lock. It seems that the generated lockfile doesn't have `version` fields for crates-io sources. Not sure why that would happen because i'm adding `version` in translator, so that didn't change. Maybe a regression?

yusdacra:
It seems that the last commit introduces that behavior, the offending line is `argsKeep = b.removeAttrs args [ "reComputeHash" "version" ];` in `src/fetchers/default.nix`. What's the reason for that change? And how can I fix this? Removing the `version` from there fixes it, but that's probably not the way to go. In the lock file, there is already a version that is used as a key for the source, so maybe that was supposed to be used somehow?

DavHau:
I refactored the interfaces recently. Version is a special field as it is always present in the lock file to key the package. We should probably not remove the version from the arguments and pass it to all fetchers by default. But it should not be stored as an extra field, as it's already in the key.

I can work on this, but probably won't make it before tomorrow.

yusdacra:
I pushed a fix for that. It seems to work fine. The way I did it is the version is still removed from `constructSource`, but it is passed to all fetchers.

It seems that since `name` and `version` is used as *keys*, technically npm, pypi and crates-io fetchers don't need to pass their own `name` and `version`s in their `sourceConstructor`. What do you think about passing `name` and `version` to fetchers by default? This would simplify for example the crates-io source constructor in rust translator to just `hash = dependencyObject.hash;` and for others as well, since you no longer have to care about passing `name` and `version`.

DavHau:
Thanks for the amazing work!

DavHau:
Yep, we should probably pass pname and version to each fetcher. Let's do that in another PR.

PR #60, opened 2021-11-29
feat: pass version and pname to fetchers

yusdacra:
This PR changes fetcher interface so that both version and pname are now passed to fetchers. This is useful since we can drop `pname` and `version` from source constructors, and it makes the lock have less duplicated data. It also adds warnings if a source constructor uses `pname` and `version` fields, since they will be overrided by dream2nix.

### DIFF
--- a/src/fetchers/default.nix
+++ b/src/fetchers/default.nix
@@ -42,7 +42,7 @@ rec {
           '' false
           else true
         ) args;
-      argsKeep = overrideWarning [ "name" "version" ] args;
+      argsKeep = overrideWarning [ "pname" "version" ] args;
       fetcherOutputs = fetcher.outputs (argsKeep // dependencyInfo);
     in
       argsKeep
--- a/src/fetchers/default.nix
+++ b/src/fetchers/default.nix
@@ -25,27 +25,15 @@ rec {
   constructSource =
     {
       type,
-      dependencyInfo ? { },
       reComputeHash ? false,
       ...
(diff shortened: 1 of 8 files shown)


src/fetchers/default.nix
```diff
@@ -25,15 +25,25 @@ rec {
   constructSource =
     {
       type,
-      sourceVersion ? null,
+      dependencyInfo ? { },
```
DavHau, line 28: Maybe instead of `dependencyInfo` we should directly pass `pname` and `version` here. That makes the interface more clear and simple I believe.

That won't allow us to raise the warning anymore, but maybe the warning should instead be triggered by `simpleTranslate` in `utils/translator.nix` instead.

yusdacra: Hmm yeah, that's a good idea. I see that `constructSource` is also used in other places, so issuing the warning in `simpleTranslate` sounds better.

yusdacra: Should be done in latest commit.

yusdacra:
Unrelated to this PR, but it seems that you can no longer use `crates-io:ripgrep/13.0.0` for example, dream2nix complains it can't be parsed.

DavHau:
Oh yes, the regex doesn't allow for the `-` symbol. I'll fix that up. Thanks

DavHau:
fixed the parsing problems via 133b2d0093b22bb3367f8792f2dcd2f235e394ba

PR #64, opened 2021-12-03
fix(translator/cargo-lock): use cargoPackages instead of cargoTomls while searching for crates

yusdacra:
This fixes an issue with `Cargo.toml` files that don't contain packages be picked up by the `findCratePath` function for the `path` source constructor.


yusdacra:
It might be useful to have a test case for this. The https://github.com/helix-editor/helix repo could be good to test as well.

PR #67, opened 2021-12-21
fix(rust): handle git source vendoring properly

yusdacra:
Before, git sources weren't vendored properly. This PR:
- writes vendor entries to .cargo/config before building
- writes .cargo-checksum in vendored sources for git deps
- handles workspace git dependencies where the crates are in subdirectories
- copies sources instead of symlinking them so we can write the cargo checksums

### DIFF
--- a/src/translators/rust/pure/cargo-lock/default.nix
+++ b/src/translators/rust/pure/cargo-lock/default.nix
@@ -24,17 +24,19 @@
           (n: v: if v == "directory" then recurseFiles "${path}/${n}" else "${path}/${n}")
           (l.readDir path)
         );
+      getAllFiles = dirs: l.flatten (l.map recurseFiles dirs);
+
+      getCargoTomlPaths = l.filter (path: l.baseNameOf path == "Cargo.toml");
+      getCargoTomls = l.map (path: { inherit path; value = l.fromTOML (l.readFile path); });
+      getCargoPackages = l.filter (toml: l.hasAttrByPath [ "package" "name" ] toml.value);
 
       # Find all Cargo.toml files and parse them
-      allFiles = l.flatten (l.map recurseFiles inputDirectories);
-      cargoTomlPaths = l.filter (path: l.baseNameOf path == "Cargo.toml") allFiles;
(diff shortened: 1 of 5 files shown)


src/utils/dream-lock.nix
```diff
@@ -77,6 +77,9 @@ let
 
       cyclicDependencies = lock.cyclicDependencies;
 
+      getSourceSpec = pname: version:
+        sources."${pname}"."${version}" or {};
```
DavHau, line 81: I think we should not use `{}`as a default here. If an element is accessed which doesn't exist, then it's a bug in the code and it should crash, right?

src/translators/rust/pure/cargo-lock/default.nix
```diff
@@ -180,31 +201,24 @@
...
               let
                 findCratePath = name:
-                  l.dirOf (
+                  l.baseNameOf (l.dirOf (
                     l.findFirst
                     (toml: toml.value.package.name == name)
                     (throw "could not find crate ${name}")
                     cargoPackages
-                  ).path;
+                  ).path);
```
DavHau, line 209: This will be inefficient if we have many such path-like dependencies as it re-iterates over all possible packages. At some point we probably should generate an attrset containing all the packages to allow for cheap lookups here.

DavHau (commented):
Thanks a lot for this. If you fix the `getSourceSpec`, we can merge it.

src/utils/dream-lock.nix
yusdacra: Yeah, that makes sense. Would a warning message like in `getSource` suffice?

src/translators/rust/pure/cargo-lock/default.nix
yusdacra: Yes, that would be nice. Ideally I would want to only do a one time pass for all the packages, and extract all info we need at once. But for now I just wanted to get it right :D

src/utils/dream-lock.nix
yusdacra: Should be done in 405e63e941d6dd7ad21393a0751462363dfd215f

PR #83, opened 2022-02-13
feat(rust): implement crane builder

yusdacra:
This PR adds a `crane` builder. Crane builds a dependencies only derivation first, and then the actual crate. For this reason, I have made the deps only derivation use `${pname}-deps` as override name and the actual crate just uses `pname` like normal.


flake.nix
```diff
@@ -66,6 +72,42 @@
...
+          "lib/writeTOML.nix"
+          "lib/cleanCargoToml.nix"
+          "pkgs/configureCargoCommonVarsHook.sh"
+          "pkgs/configureCargoVendoredDepsHook.sh"
+          "pkgs/inheritCargoArtifactsHook.sh"
+          "pkgs/installCargoArtifactsHook.sh"
+          "pkgs/remapSourcePathPrefixHook.sh"
+          "LICENSE"
+        ];
+        nix-std = [
```
yusdacra, line 89: I personally don't like that crane requires this entire repo just for a `toTOML` function (afaik), maybe we could vendor our own `toTOML`...?

src/builders/default.nix
```diff
@@ -21,8 +21,10 @@
   };
   
   rust = rec {
-    default = buildRustPackage;
+    default = crane;
```
yusdacra, line 24: Should the default be made crane?

src/default.nix
```diff
@@ -238,6 +298,7 @@ let
   # build a dream lock via a specific builder
   callBuilder =
     {
+      source,
```
yusdacra, line 301: This is needed to pass the original source the user passed to `riseAndShine` so that it can be used in the `crane` builder, because `crane` expects a Cargo.lock and a Cargo.toml for a builds only dep, which will only be available at workspace root for workspaces. dream2nix dependency sources won't have this information if it's a workspace, of course.

Maybe the name could be changed to something else though?

flake.nix
DavHau: I just tried in a repl and if I delete all files from nix-std except `{default,flake,list,serde,set,string}.nix`, the toTOML function still works.

So maybe we can just keep a limited number of files.

But yeah, obviously it would be better if we just had less dependencies.

src/builders/default.nix
DavHau: I leave that decision up to you.

src/default.nix
```diff
@@ -100,6 +100,66 @@ let
...
+            "inheritCargoArtifactsHook"
+            "installCargoArtifactsHook"
+            "remapSourcePathPrefixHook"
+          ] makeHook;
+      in rec {
+        # These aren't used by dream2nix
+        crateNameFromCargoToml = null;
+        vendorCargoDeps = null;
+
+        writeTOML = importLibFile "writeTOML" {
```
DavHau, line 134: Oh, I see what you mean. We could just put our own `writeTOML` here. That would probably be the best option.

src/default.nix
DavHau: Hm, but this will only be available if `riseAndShine` is called with a source tree as `source`. But it must also work if a dream-lock.json is passed as `source`. Therefore all necessary information for building must be in the dream-lock.

Actually since the dream-lock now supports more than one top-level package, we need to extend the `path` source type with a field that points to the root package source it resides in.
We could add for example a field named `root` pointing to a package name from `.generic.packages`.
Then we could add a builder function `getSourceRoot` or something like that.

What do you think?

DavHau (commented):
Thanks, please see comments below.

src/default.nix
yusdacra: Yes, it is kind of hacky right now. It does try to handle it if it's a dream-lock but it's not good with that.

A `getSourceRoot` function that I would need should essentially just "resolve" path dependencies to their root, so without the relative path prepended. Everything else is a "root" on their own, so they don't matter.

DavHau: Would the `getSourceRoot` function even be sufficient for this? Or would there still be the problem that we don't know the exact location of the toml file and have to search for it again?

Another potential problem: Does crane read the cargo toml/lock during eval time?
If we build from dream-lock, then reading contents of files within the source would require IFD, because teh source has to be fetched first, for evaluation to continue. We should prevent that.

In case we really need the Cargo.toml content during eval time, we can embed the toml content inside `subsystemAttrs` and then pass it to the builder. Not nice, but the architecture would be fine with it.

yusdacra: `getSourceRoot` should be enough, it will always contain a `Cargo.lock`. For path dependencies this means a workspace, for others they are either fetched from crates-io which will always have a Cargo.lock or they are fetched from github / gitlab but aren't a workspace.

IIRC the only place crane does read the Cargo.toml is for the `cleanCargoToml` function which is called when building the dummy source. I believe that could be moved to the dummy source derivation itself, can make an issue for it on the crane repo. I think it also does IFD while generating some dummy source files. It would be appreciated if you could check `lib/mkDummySource.nix` and `lib/cleanCargoToml.nix` on crane repo for IFDs.

src/builders/default.nix
yusdacra: I made it the default builder since it allows for more caching than `buildRustPackage`. It should be a sane default if we fix the IFD woes. It also seemed much simpler to me than how `buildRustPackage` reads / works, so I think it would be more easy to override.

src/default.nix
DavHau: If you invoke via flake, then the input source will be fetched before evaluation and therefore doesn't trigger IFD.
But if you render the dream-lock.json and then call dream2nix without flakes, it will throw an error:
```
> nix run . -- add github:BurntSushi/ripgrep/13.0.0
> nix build -f ./ripgrep/default.nix --show-trace -L --option allow-import-from-derivation false
error: cannot build '/nix/store/xpyh6a730rkk7kirah3v6swi9m1lr1jj-source.drv' during evaluation because the option 'allow-import-from-derivation' is disabled

       … while realising the context of path '/nix/store/rps32shpcxjz0siavrrqziapsf9ynhb1-source/Cargo.toml'

       at /nix/store/la5wa5mxlr6jn6819ii670fa209glgwc-dream2nix-external-dir/crane/lib/buildDepsOnly.nix:33:8:

           32|   dummySrc =
           33|     if builtins.pathExists cargoToml && builtins.pathExists cargoLock
             |        ^
           34|     then mkDummySrc args

       … while evaluating the attribute 'src' of the derivation 'ripgrep-deps-13.0.0'

       at /nix/store/96y8li2047jfsg5b6zasdaz9gvdw5ywg-source/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let

       … while evaluating the attribute 'cargoArtifacts' of the derivation 'ripgrep-13.0.0'

       at /nix/store/96y8li2047jfsg5b6zasdaz9gvdw5ywg-source/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let
```

DavHau: I also see the option that we make the framework aware of the fact that a builder requires IFD, so we can disable some of the builders whenever dream2nix is used in a non IFD enabled environment.

yusdacra: > But if you render the dream-lock.json and then call dream2nix without flakes, it will throw an error
Oh, I haven't considered that :sweat_smile:

> I also see the option that we make the framework aware of the fact that a builder requires IFD, so we can disable some of the builders whenever dream2nix is used in a non IFD enabled environment.
Yeah, i think this would be a good way to go with things. How do you think should this be done? Maybe we could make the builders in `builders/default.nix` be attrsets with a `requiresIfd` attribute, and the actual builder. I'm not aware of a way to detect IFDs, though `tryEval` + something that triggers an IFD could work fine?

DavHau: `tryEval` won't work. It only catches `throw` and `assert` errors. Also `tryEval` should not be used for exception handling, as eelco explained here: https://github.com/NixOS/nix/pull/5564#issuecomment-968750546

If we did introduce an IFD flag, it probably would have to be hardcoded by a developer.

But I think if possible somehow we should avoid the builder doing any IFD to not add another source of possible confusion for the user by having different `types` of builders.

Maybe we just copy the whole Cargo.toml file into `subsystemAttrs` and make the builder read it from there?

yusdacra: We can't really do that I think. The way crane builds the dependency-only derivation, it creates a dummy source only retaining the most important information, as to avoid frequent rebuilds just because flake.nix changed or project source code changed. The way it does this, it finds all `Cargo.toml` files, and then "cleans" them (stripping them of unnecessary information). So I don't think we can avoid that, sadly. We could put all `Cargo.toml`s in the dream-lock, but even then it uses `Cargo.lock` and reads some other files too, so not sure if it's feasible.

DavHau: The core idea of separating translation and building in dream2nix is that after the translation we have extracted all information from the source we ever need for constructing the build. Now passing through the `source` kind of defeats that principle. I'm not sure if we should adapt the framework to allow for this.

I think the only clean way of integrating crane is to change crane to get some of the information it needs via parameters instead of reading directly from the Cargo.toml o Cargo.lock files.

I see that this might involve some work. So as a workaround for now we could just introduce a little patch in crane, so it allows passing the toml/lock as attrsets instead of file paths. Then we can just store the cargo lock and toml for each package in the dream lock.

Does that make sense for you?

DavHau: Ah, I see crane execute `findCargoFiles` inside `mkDummySrc.nix`, so that would have to be changed to receive a list of already loaded cargo files (attrsets).

yusdacra: > So as a workaround for now we could just introduce a little patch in crane,

This would work, but it sounds hacky and I'm not sure if i want to maintain something like that. IMHO if we don't want to include crane due to IFD concerns, I can just go for a granular builder using `buildRustCrate`, if that's fine by you? This PR was mostly just testing the waters, so I'm fine with going for something else :slightly_smiling_face:

I was also thinking of doing a `naersk` builder, but `naersk` also does the same thing as `crane` does, ie. reading `Cargo.toml`s / `.cargo/config`s in eval time, so that also wouldn't really work either. So the only feasible option if we don't want to have it be too hacky by patching stuff around is just implementing a granular builder.

DavHau: To be honest I think I would prefer the custom solution. Part of the idea behind dream2nix was to go away from the monolithic `lock-file -> nix derivations` model which most other `2nix` tools use, and instead separate the process in distinct phases (translation, fetching, building), so we can have generic (user) interfaces for all of these phases.

I'm worried now that if we circumvent that model and allow exceptions, that later we might run into trouble keeping our interfaces uniform and shift more complexity to the end user.
But also I'm currently unable to come up with a good example that underlines this point.

So maybe we should just go ahead and merge this if you think this already adds value to the project. I leave the decision up to you.

If we decide to keep this, then we should at least remove the `source` attribute being passed to the builder and add the `getSourceRoot` function we talked about.

yusdacra: I'm not sure if I want to merge it since unlike `buildRustPackage` it doesn't use env variables in it's build / check / install phases which makes it hard to override those phases or some other stuff. `buildRustPackage` is fine for now IMO, until a granular builder lands.

I'll still keep the branch around though, in case someone has a better idea on how to do all this.

yusdacra: I think we can still merge it though, I feel it would be better if people saw something like this is actually implemented, some people might still want a deps only drv for better caching.

I can add a warning that this requires IFD, and not make it the default builder. And also introduce `getSourceRoot`. How does that sound?

DavHau: Sounds great! Let's do it.

yusdacra: I implemented what you said on the matrix room, and it seems to be working fine now! Both `buildRustPackage` and `crane` builder seem to be able to build things fine. `buildRustPackage` is still the default. I added a comment in `src/builders/default.nix` for crane that says it requires IFD. Should be ready to go now!

PR #84, opened 2022-02-14
style: format with alejandra

kamadorueda:


overrides/nodejs/electron/generic.nix
```diff
@@ -1,52 +1,53 @@
...
+  mesa,
+  libxkbcommon,
+  libappindicator-gtk3,
+  libxshmfence,
+  libXdamage,
+  nss,
 }:
-
-version: hashes:
-let
```
DavHau, line 28: Here the `let` is pulled up

src/builders/nodejs/granular/default.nix
```diff
@@ -7,65 +7,64 @@
...
-
+  standalonePackageNames ? [ ],
   nodejs ? null,
   ...
-}@args:
-
+}
+@ args:
 let
-
```
DavHau, line 52: ... and here the `let` is not pulled up

src/builders/nodejs/granular/default.nix
```diff
@@ -148,51 +151,51 @@ let
...
-        else
-          pkgs."electron_${electronVersionMajor}".headers;
-
-
-      pkg =
-        produceDerivation name (stdenv.mkDerivation rec {
-
+  makePackage = name: version: let
+    deps = getDependencies name version;
+
```
DavHau, line 156: ...and here it is pulled up again

DavHau: Not a big issue, just noticing the inconsistency. Maybe it's for a good reason, maybe not...

kamadorueda: Thanks for noticing, this kind of feedback is very useful!

src/builders/nodejs/granular/default.nix
```diff
@@ -7,65 +7,64 @@
...
-  buildPackageWithOtherBuilder, # { builder, name, version }: -> drv
-
+  getCyclicDependencies,
+  # name: version: -> [ {name=; version=; } ]
+  getDependencies,
+  # name: version: -> [ {name=; version=; } ]
+  getSource,
+  # name: version: -> store-path
+  buildPackageWithOtherBuilder,
+  # { builder, name, version }: -> drv
```
DavHau, line 26: Inline comments are moved to the next line and now describe the wrong line.
Those should either stay there or moved into the upper line.

DavHau: Oh, I just noticed now that there is already an open issue for this

kamadorueda: solved

kamadorueda: oh, no, it was not solved for patterns. Just list, attr-sets, let-ins

src/builders/go/gomod2nix/default.nix
```diff
@@ -4,13 +4,13 @@
...
-
 {
   fetchedSources,
   dreamLock,
 }:
 let
-  gomod2nixTOML = fetchedSources.mapAttrs
+  gomod2nixTOML =
+    fetchedSources.mapAttrs
     dependencyObject.goName;
```
DavHau, line 14: For what reason is that function call pushed into the next line why others are not?

kamadorueda: It's because there was previously a line between the function and the argument:

```nix
  gomod2nixTOML = fetchedSources.mapAttrs
    dependencyObject.goName;
```

Alejandra interprets that you don't want a long line, and formats it vertically

If you do this:


```nix
  gomod2nixTOML = fetchedSources.mapAttrs dependencyObject.goName;
```

Alejandra will interpret that you want a long line, and would format horizontally

There is no concept of max-line-length, so the user has the freedom of introducing newlines where a break would be more readable

PR #89, opened 2022-02-21
Fix(cli): process a default value that returns empty

tgunnoe:
In the CLI, if no value is provided for the input the default is supposed to be selected. But if the default value returned from the translator is an empty string, the ask will infinitely recurse. I'm a bit rusty on Python so apologies if this method isn't sufficient! 

Example:
```
Please specify 'npmArgs': Additional arguments for npm
Example values: --force
Leave empty for default ()
npmArgs: 
npmArgs: 
npmArgs: 
npmArgs:
```


DavHau:
The whole interactivity part is removed by now. Thanks anyways/

PR #94, opened 2022-03-04
use one nixpkgs instance

zimbatm:
Push the nix configuration into the CLI so we don't need the overlay,
and so we don't need to create another instance of nixpkgs.

This also means that the python CLI can be executed directly as long as
the user has a recent-enough version of nix and make iterations a bit
faster.

For more details, see
https://zimbatm.com/notes/1000-instances-of-nixpkgs

TODO: the PR needs testing

### DIFF
--- a/src/apps/cli/nix_ffi.py
+++ b/src/apps/cli/nix_ffi.py
@@ -6,6 +6,8 @@ import tempfile
 
 dream2nix_src = os.environ.get("dream2nixSrc")
 
+def nix(*args, **kwargs):
+  return sp.run(["nix", "--option", "experimental-features", "nix-command flakes"] + list(args), capture_output=True, **kwargs)
 
 def callNixFunction(function_path, **kwargs):
   with tempfile.NamedTemporaryFile("w") as input_json_file:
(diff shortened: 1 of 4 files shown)


src/apps/cli/nix_ffi.py
```diff
@@ -6,6 +6,8 @@
 
 dream2nix_src = os.environ.get("dream2nixSrc")
 
+def nix(*args, **kwargs):
+  return sp.run(["nix", "--option experimental-features", "nix-command flakes"] + args, capture_output=True, **kwargs)
```
DavHau, line 10:
```suggestion
  return sp.run(["nix", "--option experimental-features", "nix-command flakes"] + list(args), capture_output=True, **kwargs)
```

DavHau:
> TODO: the PR needs testing

If `nix run.#tests-impure` succeeds, the CLI can be considered working.

src/apps/cli/nix_ffi.py
DavHau: Oh, that doesn't seem to fix the issue. I'm confused

DavHau:
Thanks!

PR #99, opened 2022-03-12
fix: parse resolved keys with git+https correctly

tgunnoe:
There's an edge case where a dependency in a yarn.lock file has a resolved address from `"package@git+https://..."`:
`git+https:...`

I am not sure why  `l.strings.sanitizeDerivationName` was commented out.

This is what fixed the issue for me.

Thanks @tomberek for the help & walkthrough.

### DIFF
--- a/src/translators/nodejs/pure/yarn-lock/default.nix
+++ b/src/translators/nodejs/pure/yarn-lock/default.nix
@@ -100,7 +100,8 @@
           || (dObj
             ? resolved
             && lib.hasInfix "codeload.github.com/" dObj.resolved)
-          || lib.hasInfix "@git+" dObj.yarnName
+          || (lib.hasInfix "@git+" dObj.yarnName
+            || lib.hasPrefix "git+" dObj.resolved)
           # example:
           # "jest-image-snapshot@https://github.com/machard/jest-image-snapshot#machard-patch-1":
           #   version "4.2.0"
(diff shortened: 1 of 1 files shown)


DavHau:
Could you provide an example lock file that triggers the issue and didn't work before this PR? I might add a test for this, so we won't break it in the future.

tgunnoe:
Here's a simple lockfile i made where i only included the one library in the same manner I was having trouble with in a monorepo.

https://gist.github.com/tgunnoe/7b39ae9c2819966cb812f3c12469a797#file-yarn-lock-L156-L161

that was made with:
```
  {
    "name": "test",
    "version": "1.0.0",
    "main": "index.js",
    "license": "MIT",
    "devDependencies": {
      "ethereumjs-abi": "git+https://github.com/ethereumjs/ethereumjs-abi.git"
    }
  }
```

However, there seems to be another error on latest, i think unrelated.

DavHau:
> However, there seems to be another error on latest, i think unrelated.

Sorry for that. If you open an issue, I will try to fix it.

DavHau:
Or wait. I will just test with the hardhat project myself. No need to open an issue

tgunnoe:
I've added a test repository with only dependency `eth-waffle` for the `git+https` here. 
This works fine at tip: https://github.com/tgunnoe/ethereum-waffle-yarnlock/blob/master/yarn.lock#L2454-L2459

I don't know why the smaller gist i provided above does not, as i just did:
```
yarn init
yarn add -D ethereumjs-abi@git+https://github.com/ethereumjs/ethereumjs-abi.git
```
instead for that one.

PR #100, opened 2022-03-13
Add aarch64-linux support

max-privatevoid:
I'd like to use dream2nix on an ARM64 machine. Passes `nix flake check`.


tgunnoe:
have you tried setting `dream2nix.lib.init { systems = [ "aarch64" ] }`, or `dream2nix.lib.init { pkgs = nixpkgs }` (nixpkgs with aarch64 as the system argument) ? support is provided by `nixpkgs` fyi.

PR #109, opened 2022-04-05
overrides: fix electron override for darwin

happysalada:
fixing the electron override for darwin.

Testing this right now.

### DIFF
--- a/overrides/nodejs/default.nix
+++ b/overrides/nodejs/default.nix
@@ -288,17 +288,25 @@ in
     in {
       add-binary = {
         overrideAttrs = old: {
-          postPatch = ''
-            cp -r ${getElectronFor "${old.version}"}/lib/electron ./dist
-            chmod -R +w ./dist
-            echo -n $version > ./dist/version
-            echo -n "electron" > ./path.txt
-          '';
+          postPatch =
+            if pkgs.stdenv.isLinux
+            then ''
+              cp -r ${getElectronFor "${old.version}"}/lib/electron ./dist
+              chmod -R +w ./dist
+              echo -n $version > ./dist/version
+              echo -n "electron" > ./path.txt
+            ''
+            else "";
 
-          postFixup = ''
-            mkdir -p $out/lib
-            ln -s $(realpath ./dist) $out/lib/electron
-          '';
+          postFixup =
+            if pkgs.stdenv.isLinux
+            then ''
+              mkdir -p $out/lib
+              ln -s $(realpath ./dist) $out/lib/electron
+            ''
+            else ''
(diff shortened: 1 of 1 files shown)


happysalada:
Just tested, this still doesn't work 
```
> chmod: changing permissions of './dist': Operation not permitted
```

happysalada:
ok the build is working now at least.
is this piece used ?
```
 echo -n "electron" > ./path.txt
```
because it doesn't appear to be used anywhere in the later steps.

happysalada:
Alright I just tried something else. I didn't like that you have to `dontBuild = true` on darwin.
What is happening in the postinstall script, is that where something is being downloaded from the internet ?
when you have some time, we have to go through this. No big deal, it won't take long and it's not very important.

happysalada:
alright I've got something that seems to work, I just need to confirm that the postinstall step is ok.

happysalada:
@DavHau friendly ping on this.

DavHau:
Thanks for pinging and sorry for the delay.
Why is the behavior different on darwin? Which command fails?
After your change the version won't be written anymore to ./dist/version

DavHau:
> ok the build is working now at least. is this piece used ?
> 
> ```
>  echo -n "electron" > ./path.txt
> ```
> 
> because it doesn't appear to be used anywhere in the later steps.

I don't think this is needed anymore. You can remove it.

happysalada:
here are the details I gathered at the time
```
Regarding "electron" package build for darwin. The build fails with cp: cannot stat '/nix/store/q4yvhf9wy56zxzcmikb8ixaimink5gy7-electron-12.2.3/lib/electron': No such file or directory . I check in the resulting derivation and there is nothing under /lib, there is a /bin/electron though. I was just looking at the electron override inside the repo, and it seems the build step for darwin is putting something inside /bin and not /lib.
```

DavHau:
OK, could you remove the path.txt thing?
And did you verify if electron apps actually work with this?

happysalada:
I never managed to build the app I was working on.
Let me test this on a working app and confirm to you.

DavHau:
Looks like something went wrong while rebasing. Maybe reset to main and cherry-pick a715898693a5b3e08421b69fa64f3ae0933de16b onto it?

happysalada:
yep, mistake on my side, I still need to leave this one open because I need to test it.
I'm close. :-)
the esbuild 0.13 is ready though!

happysalada:
of course this doesn't work.
I'm going to close this for now just to reduce the noise for you, since I'll push a lot to this branch to make it work properly.

PR #114, opened 2022-04-10
feat(translator/cargo-lock): use discoveredProjects to find dependencies that are outside of a packages source

### LINKED ISSUE #113: Rust builds - referencing path dependencies outside of a workspace.
Hi!

While playing around using `dream2nix` for our monorepo, I ran into some trouble with crate dependencies that aren't in a workspace.

The monorepo contains projects in many languages, and we currently using nix for
   - rust builds - https://github.com/cargo2nix/cargo2nix
   - python builds - poetry2nix 
The monorepo directory structure is something like:

```
root
   |
   | service
         |
         | -- > serviceA/Cargo.toml // imports ../../lib/libraryA
         | -- > serviceB/Cargo.toml // imports libraryA, but will never import serviceA
         | -- > pythonServiceA/pyproject.toml // imports pythonLibraryA
   | lib
         | -- > libraryA/Cargo.toml // may import libraryB
         | -- > libraryB/Cargo.toml
         | -- > pythonLibraryA/pyproject.toml
```
The python builds work well - the overlay being:
```
        overlay = nixpkgs.lib.composeManyExtensions [
          poetry2nix.overlay
          (final: prev: {
            devEnv = prev.poetry2nix.mkPoetryEnv {
              python = final.pkgs.python38;
              projectDir = ./.;
              overrides = prev.poetry2nix.overrides.withDefaults
                (import ../../tools/nix/python-nix-overlay.nix);
            };
          })
        ];
```

Cargo2nix also works well - but keeping the `Cargo.nix` files in sync with `Cargo.lock` files is somewhat of a hassle.

When trying to build a service using `dream2nix` it would fail with
`error: could not find crate libraryA`

The flake looks like
```
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs?ref=release-21.11";
    dream2nix.url = "github:nix-community/dream2nix";
  };

  outputs = { self, dream2nix, ... }@inputs:
    let
        systems = [ "x86_64-linux" "aarch64-darwin" ];
        config.projectRoot = ./.;
      };
    in dream2nix.makeFlakeOutputs {
      source = ./.;
    };
}
```

I also tried changing `config.projectRoot` to `../..` and using:
```
      sourceOverrides = oldSources: {
        capture_storage_client = builtins.path {
          path = ../../lib/libraryA;
        };
      };
```

But since poetry2nix manages to work around this I reckoned that the rust builder can do it as well 😄 

Thanks!

yusdacra:
Resolves #113. This uses `discoveredProjects` to find dependencies that are outside of the `inputDir` passed to the `cargo-lock` translator.

This also reworks how the `cargo-lock` translator finds Cargo packages. Since packages / workspaces discovery is handled by the discoverer now, that is replaced with just finding packages for workspace members if the translator is translating a workspace. This makes everything a lot simpler, and generally should make everything faster (no `recurseFiles`...).

It is currently a draft because there is a bug preventing this from fully working:
```
error: cannot coerce null to a string

       at /nix/store/0ziix5kkds3rkqsjc48bzlnyf5a1y2kw-source/src/fetchers/default-fetcher.nix:30:38:

           29|           then throw "source for ${name}@${version} is referencing itself"
           30|           else "${overriddenSources."${source.rootName}"."${source.rootVersion}"}/${source.path}"
               |                                                        ^
           31|         else if fetchers.fetchers ? "${source.type}"
``` 
this is fixed now. Now it is replaced with:
```
error: value is a string with context while a set was expected
```
for reproducing, run `nix build github:yusdacra/dream2nix-bugs/pr-114#basic-bin`.

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -537,7 +537,7 @@
       (proj: let
         translator = getTranslator proj.subsystem proj.translator;
         dreamLock' = translator.translate {
-          inherit source tree;
+          inherit source tree discoveredProjects;
           project = proj;
         };
         dreamLock =
--- a/src/fetchers/default-fetcher.nix
+++ b/src/fetchers/default-fetcher.nix
@@ -12,18 +12,18 @@
   sources,
   ...
 }: let
-  b = builtins;
+  l = lib // builtins;
 
   fetchedSources =
-    lib.mapAttrs
+    l.mapAttrs
     (name: versions:
-      lib.mapAttrs
+      l.mapAttrs
       (version: source:
         if source.type == "unknown"
         then "unknown"
         else if source.type == "path"
         then
-          if lib.isStorePath source.path
+          if l.isStorePath (l.concatStringsSep "/" (l.take 4 (l.splitString "/" source.path)))
           then source.path
           else if name == source.rootName && version == source.rootVersion
           then throw "source for ${name}@${version} is referencing itself"
@@ -43,7 +43,7 @@
     sources;
 
   overriddenSources =
-    lib.recursiveUpdateUntil
+    l.recursiveUpdateUntil
     (path: l: r: lib.isDerivation l)
     fetchedSources
     (sourceOverrides fetchedSources);
--- a/src/translators/rust/pure/cargo-lock/default.nix
+++ b/src/translators/rust/pure/cargo-lock/default.nix
@@ -10,22 +10,22 @@ in {
     utils,
     ...
   }: {
-    source,
+    project,
+    tree,
     packageName,
     discoveredProjects,
     ...
   } @ args: let
-    inputDir = source;
-
-    # Function to read and parse a TOML file and return an attrset
-    # containing its path and parsed TOML value
-    readToml = path: {
-      inherit path;
-      value = l.fromTOML (l.readFile path);
-    };
+    # get the root source and project source
+    rootSource = tree.fullPath;
+    projectSource = "${tree.fullPath}/${project.relPath}";
+    projectTree = tree.getNodeFromPath project.relPath;
 
     # Get the root toml
-    rootToml = readToml "${inputDir}/Cargo.toml";
+    rootToml = {
+      path = "${projectSource}/Cargo.toml";
+      value = projectTree.files."Cargo.toml".tomlContent;
+    };
 
     # Get all workspace members
     workspaceMembers =
@@ -38,7 +38,7 @@ in {
           if l.last components == "*"
           then let
             parentDirRel = l.concatStringsSep "/" (l.init components);
-            parentDir = "${inputDir}/${parentDirRel}";
+            parentDir = "${projectSource}/${parentDirRel}";
             dirs = l.readDir parentDir;
           in
             l.mapAttrsToList
@@ -50,7 +50,10 @@ in {
     # Get cargo packages (for workspace members)
     workspaceCargoPackages =
       l.map
-      (relPath: readToml "${inputDir}/${relPath}/Cargo.toml")
+      (relPath: {
+        path = "${projectSource}/${relPath}/Cargo.toml";
+        value = (projectTree.getNodeFromPath "${relPath}/Cargo.toml").tomlContent;
+      })
       # Filter root referencing member, we already parsed this (rootToml)
       (l.filter (relPath: relPath != ".") workspaceMembers);
 
@@ -70,22 +73,11 @@ in {
       if args.packageName == "{automatic}"
       then packageToml.value.package.name
       else args.packageName;
-
-    # Find the base input directory, aka the root source
-    baseInputDir = let
-      # Find the package we are translating in discovered projects
-      thisProject =
-        l.findFirst (project: project.name == packageName) null discoveredProjects;
-      # Default to an no suffix, since if we can't find our package in
-      # discoveredProjects, it means that we are in a workspace and our
-      # package will be in this workspace, so root source is inputDir
-    in
-      l.removeSuffix (thisProject.relPath or "") inputDir;
     # Map the list of discovered Cargo projects to cargo tomls
     discoveredCargoTomls =
       l.map (project: rec {
         value = l.fromTOML (l.readFile path);
-        path = "${baseInputDir}/${project.relPath}/Cargo.toml";
+        path = "${rootSource}/${project.relPath}/Cargo.toml";
       })
       discoveredProjects;
     # Filter cargo-tomls to for files that actually contain packages
@@ -99,7 +91,7 @@ in {
       discoveredCargoTomls;
 
     # Parse Cargo.lock and extract dependencies
-    parsedLock = l.fromTOML (l.readFile "${inputDir}/Cargo.lock");
+    parsedLock = l.fromTOML (l.readFile "${projectSource}/Cargo.lock");
     parsedDeps = parsedLock.package;
     # This parses a "package-name version" entry in the "dependencies"
     # field of a dependency in Cargo.lock
@@ -129,8 +121,6 @@ in {
 
     package = rec {
       toml = packageToml.value;
-      tomlPath = packageToml.path;
-
       name = toml.package.name;
       version =
         toml.package.version
@@ -245,7 +235,7 @@ in {
             null;
           toml = findToml cargoPackages;
           discoveredToml = findToml discoveredCargoPackages;
-          relDir = lib.removePrefix "${inputDir}/" (l.dirOf toml.path);
+          relDir = lib.removePrefix "${projectSource}/" (l.dirOf toml.path);
         in
           if
             package.name
@@ -253,7 +243,7 @@ in {
             && package.version == dependencyObject.version
           then
             dlib.construct.pathSource {
-              path = source;
+              path = projectSource;
               rootName = null;
               rootVersion = null;
             }
@@ -285,6 +275,8 @@ in {
         };
       };
     });
+
+  version = 2;
 
   projectName = {source}: let
     cargoToml = "${source}/Cargo.toml";
--- a/src/translators/rust/pure/cargo-lock/default.nix
+++ b/src/translators/rust/pure/cargo-lock/default.nix
@@ -12,61 +12,90 @@ in {
   }: {
     source,
     packageName,
+    discoveredProjects,
     ...
   } @ args: let
     inputDir = source;
 
-    recurseFiles = path:
-      l.flatten (
-        l.mapAttrsToList
-        (n: v:
-          if v == "directory"
-          then recurseFiles "${path}/${n}"
-          else "${path}/${n}")
-        (l.readDir path)
-      );
+    # Function to read and parse a TOML file and return an attrset
+    # containing its path and parsed TOML value
+    readToml = path: {
+      inherit path;
+      value = l.fromTOML (l.readFile path);
+    };
 
-    # Find all Cargo.toml files and parse them
-    allFiles = l.flatten (l.map recurseFiles [inputDir]);
-    cargoTomlPaths = l.filter (path: l.baseNameOf path == "Cargo.toml") allFiles;
-    cargoTomls =
+    # Get the root toml
+    rootToml = readToml "${inputDir}/Cargo.toml";
+
+    # Get all workspace members
+    workspaceMembers =
       l.map
-      (path: {
-        inherit path;
-        value = l.fromTOML (l.readFile path);
-      })
-      cargoTomlPaths;
+      (
+        memberName: let
+          components = l.splitString "/" memberName;
+        in
+          # Resolve globs if there are any
+          if l.last components == "*"
+          then let
+            parentDirRel = l.concatStringsSep "/" (l.init components);
+            parentDir = "${inputDir}/${parentDirRel}";
+            dirs = l.readDir parentDir;
+          in
+            l.mapAttrsToList
+            (name: _: "${parentDirRel}/${name}")
+            (l.filterAttrs (_: type: type == "directory") dirs)
+          else memberName
+      )
+      (rootToml.value.workspace.members or []);
+    # Get cargo packages (for workspace members)
+    workspaceCargoPackages =
+      l.map
+      (relPath: readToml "${inputDir}/${relPath}/Cargo.toml")
+      workspaceMembers;
 
-    # Filter cargo-tomls to for files that actually contain packages
+    # All cargo packages that we will output
     cargoPackages =
-      l.filter
-      (toml: l.hasAttrByPath ["package" "name"] toml.value)
-      cargoTomls;
+      if l.hasAttrByPath ["package" "name"] rootToml.value
+      # Note: the ordering is important here, since packageToml assumes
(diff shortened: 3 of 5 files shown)


omerbenamram:
@yusdacra Thanks for picking this up!

I did try this out and I'm still getting the same error (could not find crate ...) and not the one you're describing.
```
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs?ref=release-21.11";
    dream2nix = {
      type = "github";
      owner = "yusdacra";
      repo = "dream2nix";
      ref = "feat/cargo-lock-subpackages";
    };
  };

  outputs = { self, dream2nix, ... }@inputs:
    let
        systems = [ "x86_64-linux" "aarch64-darwin" ];
        config.projectRoot = ./.;
      };
    in dream2nix.makeFlakeOutputs {
      source = ./.;
    };
}
```

yusdacra:
> @yusdacra Thanks for picking this up!
> 
> I did try this out and I'm still getting the same error (could not find crate ...) and not the one you're describing. Note that a workspace is _not_ being used - just path deps like this:
> 
> ```
> [dependencies]
> serde = { version = "^1.0", features = ["derive"] }
> serde_json = { version = "^1.0" }
> private_lib = { path = "../../lib/private_lib" }
> ```
> 
> ```
> {
>   inputs = {
>     nixpkgs.url = "github:nixos/nixpkgs?ref=release-21.11";
>     dream2nix = {
>       type = "github";
>       owner = "yusdacra";
>       repo = "dream2nix";
>       ref = "feat/cargo-lock-subpackages";
>     };
>   };
> 
>   outputs = { self, dream2nix, ... }@inputs:
>     let
>       dream2nix = inputs.dream2nix.lib.init {
>         # modify according to your supported systems
>         systems = [ "x86_64-linux" "aarch64-darwin" ];
>         config.projectRoot = ./.;
>       };
>     in dream2nix.makeFlakeOutputs {
>       source = ./.;
>     };
> }
> ```

Could you create a repo that showcases the error?

omerbenamram:
@yusdacra https://github.com/omerbenamram/dream2nix-issue-113 

😄

yusdacra:
> @yusdacra [omerbenamram/dream2nix-issue-113](https://github.com/omerbenamram/dream2nix-issue-113)

The reason this doesn't work is because the crate directories don't have `Cargo.lock` files (or are ignored by git), so they aren't picked up by `dream2nix`. If you really want to ignore `Cargo.lock` files, ie. not stage them but still have them tracked, you can use `git add --intent-to-add`.

yusdacra:
> Hmm - I've added all `Cargo.lock` files - but I still can't seem to build it ([omerbenamram/dream2nix-issue-113](https://github.com/omerbenamram/dream2nix-issue-113)). Are there any other changes maybe I'm missing?

I've added them to a fork and pushed here https://github.com/yusdacra/dream2nix-issue-113. You can try that. It should not build, but it should output the error I mentioned in the OP.

omerbenamram:
Maybe it's something else in my environment 😅 ? from the fork checked out i'm still getting the missing libraries.

CleanShot 2022-04-12 at 11 13 31

omerbenamram:
I understand - but I thought this is what  `config.projectRoot` allows to modify, by setting it to `../..` it stores all of the repository in the /nix/store.

yusdacra:
> I understand - but I thought this is what  `config.projectRoot` allows to modify, by setting it to `../..` it stores all of the repository in the /nix/store.
> 

The problem is that your crates aren't being discovered by `dream2nix`, so they aren't known by the translator. For them to be discovered you need to have both Cargo.lock and Cargo.toml in the crate source. What `dream2nix` could do is perhaps look only for `Cargo.toml`, but the translator outputs an error if there isn't a `Cargo.lock` for the crate that is being built. I will look into that.

DavHau:
@yusdacra I think it would be better if we would not discover workspace projects as separate projects in the first place.
Everything that shares a lock file should be detected as a single project and therefor map to a single translator call. If I understand correctly, with this PR's change we would call translation for each workspace which is unnecessary.

For nodejs I also exclude workspace projects from discovered projects. Since each translator call will produce exactly one dream-lock and the dream-lock can output more than one package, all packages will show up in the final output. It doesn't make a difference for the user but it is more efficient. Also we don't need to pass discovered projects to the translator with this strategy.

DavHau:
Just an example.
On this project you can see how the output of `discoverProjects` only consists of one single project, yet the `nix flake show` lists all packages.

```sh
> clone https://github.com/ethereum-optimism/optimism --depth 1
> cd optimism
> nix flake init -t github:nix-community/dream2nix#simple
> nix flake show
warning: Git tree '/home/grmpf/synced/projects/github/optimism' is dirty
git+file:///home/grmpf/synced/projects/github/optimism?shallow=1
├───apps
│   └───x86_64-linux
│       └───resolveImpure: app
├───packages
│   └───x86_64-linux
│       ├───"@eth-optimism/batch-submitter-service": package '__at__eth-optimism__slash__batch-submitter-service-0.1.10'
│       ├───"@eth-optimism/common-ts": package '__at__eth-optimism__slash__common-ts-0.2.6'
│       ├───"@eth-optimism/contracts": package '__at__eth-optimism__slash__contracts-0.5.21'
│       ├───"@eth-optimism/core-utils": package '__at__eth-optimism__slash__core-utils-0.8.3'
│       ├───"@eth-optimism/data-transport-layer": package '__at__eth-optimism__slash__data-transport-layer-0.5.27'
│       ├───"@eth-optimism/gas-oracle": package '__at__eth-optimism__slash__gas-oracle-0.1.11'
│       ├───"@eth-optimism/go-builder": package '__at__eth-optimism__slash__go-builder-0.0.3'
│       ├───"@eth-optimism/hardhat-node": package '__at__eth-optimism__slash__hardhat-node-0.1.4'
│       ├───"@eth-optimism/indexer": package '__at__eth-optimism__slash__indexer-0.0.4'
│       ├───"@eth-optimism/integration-tests": package '__at__eth-optimism__slash__integration-tests-0.5.9'
│       ├───"@eth-optimism/js-builder": package '__at__eth-optimism__slash__js-builder-0.0.3'
│       ├───"@eth-optimism/l2geth": package '__at__eth-optimism__slash__l2geth-0.5.17'
│       ├───"@eth-optimism/l2geth-exporter": package '__at__eth-optimism__slash__l2geth-exporter-0.0.4'
│       ├───"@eth-optimism/message-relayer": package '__at__eth-optimism__slash__message-relayer-0.4.5'
│       ├───"@eth-optimism/op-exporter": package '__at__eth-optimism__slash__op-exporter-0.5.4'
│       ├───"@eth-optimism/proxyd": package '__at__eth-optimism__slash__proxyd-3.8.4'
│       ├───"@eth-optimism/replica-healthcheck": package '__at__eth-optimism__slash__replica-healthcheck-1.0.4'
│       ├───"@eth-optimism/sdk": package '__at__eth-optimism__slash__sdk-1.1.0'
│       ├───"@eth-optimism/teleportr": package '__at__eth-optimism__slash__teleportr-0.0.3'
│       └───optimism: package 'optimism-1.0.0'
└───projectsJson: unknown

> nix eval .#projectsJson --raw | jq
[
  {
    "dreamLockPath": "packages/optimism/dream-lock.json",
    "name": "optimism",
    "relPath": "",
    "subsystem": "nodejs",
    "subsystemInfo": {
      "workspaces": [
        "packages/common-ts",
        "packages/contracts",
        "packages/core-utils",
        "packages/data-transport-layer",
        "packages/message-relayer",
        "packages/replica-healthcheck",
        "packages/sdk",
        "l2geth",
        "integration-tests",
        "go/batch-submitter",
        "go/gas-oracle",
        "go/indexer",
        "go/l2geth-exporter",
        "go/op-exporter",
        "go/proxyd",
        "go/teleportr",
        "ops/docker/hardhat",
        "ops/docker/go-builder",
        "ops/docker/js-builder"
      ]
    },
    "translator": "yarn-lock",
    "translators": [
      "yarn-lock",
      "package-json"
    ]
  }
]
```

yusdacra:
> @yusdacra I think it would be better if we would not discover workspace projects as separate projects in the first place.
> Everything that shares a lock file should be detected as a single project and therefor map to a single translator call.

I'm not sure what you meant by "discovering workspace projects as seperate projects"? The current behavior is that if a directory has a `Cargo.toml` and a `Cargo.lock`, then it is counted as a "project", which is correct. That can either be a Cargo workspace or a Cargo package. So everything that shares a lockfile will already be detected as a single project, this PR doesn't change anything with that.

> If I understand correctly, with this PR's change we would call translation for each workspace which is unnecessary.

But this is already what it does? As i explained above, the discoverer will discover Cargo workspaces and Cargo packages as "projects". These don't share the same lockfile. If a repository has two seperate Cargo workspaces, then they'll be translated seperately, there is no way around that.

> For nodejs I also exclude workspace projects from discovered projects. Since each translator call will produce exactly one dream-lock and the dream-lock can output more than one package, all packages will show up in the final output. It doesn't make a difference for the user but it is more efficient. Also we don't need to pass discovered projects to the translator with this strategy.

Again I'm not sure how this would help with the original issue. You have to somehow let the cargo-lock translator know of projects. Since the cargo-lock translator will only be aware of the input path given to it, it won't be able to work correctly if a dependency for the package / workspace we are translating is outside that input path, since it won't be able to find it and add it as a path dependency. For this passing discovered projects makes it easy, the translator can just filter that to get Cargo packages and use those to get the paths for dependencies outside the input path it was passed.

DavHau:
> I'm not sure what you meant by "discovering workspace projects as seperate projects"? The current behavior is that if a directory has a Cargo.toml and a Cargo.lock, then it is counted as a "project", which is correct. That can either be a Cargo workspace or a Cargo package. So everything that shares a lockfile will already be detected as a single project, this PR doesn't change anything with that.

Oh, absolutely right, I made some wrong assumptions here. So the discovery already behaves as it should.

Still I'm not sure if passing all discovered projects to the translator is a good idea.
The problem I currently see is that the new translator interface is not documented and I will do that right now and come back to you. This will make it simpler to discuss about this. I might just hack on this PR a bit as well and see if I can find another way of solving that with the new interface.

DavHau:
I updated the translator templates via 8eebb32a52ce49a1bd9327f197243a4f0e077722.
Most of the changes are there because it now uses `simpleTranslate2`.
Though, the cargo-lock translator can continue using the old helper `simpleTranslate`, no need to change it now.

Also future translators should only receive `tree` and `project` as arguments instead of `source`.

I started refactoring the current PR a bit here: https://github.com/nix-community/dream2nix/tree/cargo-lock-subpackages

I won't have time tomorrow to continue this. If you want you can take it from there. Otherwise I will finish it another time.

yusdacra:
> I updated the translator templates via [8eebb32](https://github.com/nix-community/dream2nix/commit/8eebb32a52ce49a1bd9327f197243a4f0e077722). Most of the changes are there because it now uses `simpleTranslate2`. Though, the cargo-lock translator can continue using the old helper `simpleTranslate`, no need to change it now.
> 
> Also future translators should only receive `tree` and `project` as arguments instead of `source`.
> 
> I started refactoring the current PR a bit here: [https://github.com/nix-community/dream2nix/tree/cargo-lock-subpackages](https://github.com/nix-community/dream2nix/tree/cargo-lock-subpackages?rgh-link-date=2022-04-14T23%3A21%3A22Z)
> 
> I won't have time tomorrow to continue this. If you want you can take it from there. Otherwise I will finish it another time.

I have gone through the translator and made it use `simpleTranslate2`, it's a much better interface than the old one. This PR specifically still doesn't work, I still get the same error as I mentioned in the OP.

DavHau:
Fixed the error via: 2331a0aa02fd1f4ae8f0508fbf1a20b9d1721793
It seems like I did not understand the behavior of `recursiveUpdateUntil` correctly.
I opened an issue upstream. https://github.com/NixOS/nixpkgs/issues/169044

DavHau:
Though the original problem demonstrated by https://github.com/omerbenamram/dream2nix-issue-113 remains unsolved.

yusdacra:
> Fixed the error via: [2331a0a](https://github.com/nix-community/dream2nix/commit/2331a0aa02fd1f4ae8f0508fbf1a20b9d1721793) It seems like I did not understand the behavior of `recursiveUpdateUntil` correctly. I opened an issue upstream. [NixOS/nixpkgs#169044](https://github.com/NixOS/nixpkgs/issues/169044)

Thanks for looking into this! I've noticed the error was coming from there but couldn't figure out why exactly it was happening.

> Though the original problem demonstrated by [omerbenamram/dream2nix-issue-113](https://github.com/omerbenamram/dream2nix-issue-113) remains unsolved.

Yeah, it looks like there is still some stuff that needs to be done. Can you reopen #113, since its still not fixed? I'll look into this today.

DavHau:
I think the problem with the current logic is, that it searches `discoveredProjects` for dependenceis, but `discoveredProjects` only contains top-level projects identified by having a Cargo.lock file.

I think we should search for dependencies in the discover phase, and write the dependencies into a project's `.subsystemData.dependencies`, or something like that.

Then the translator knows immediately about all the deps and doesn't need to search again.

yusdacra:
> I think the problem with the current logic is, that it searches discoveredProjects for dependenceis, but discoveredProjects only contains top-level projects identified by having a Cargo.lock file.

This is not really an issue I believe. The current logic will detect those with a `Cargo.lock` file. A `Cargo.lock` file is essential if you do anything at all with the projects, so developing them in any way will create it. The reason it doesn't currently build is that because vendoring doesn't play well with path dependencies. So with the example repo, the hierarchy is basically like this:

```
lib/another_library
lib/library -> depends on another_library
service/binary -> depends on library
```

When dream2nix tries to build `service/binary`, what basically happens is since the `lib/library` is a path dependency, Cargo expects it to be at that relative path. But since dream2nix uses the `service/binary` path, this fails.

I think what we might have to do here is make it use the root source, instead of the `service/binary` source. Then we would need to modify the builders so that they change directory to `service/binary` and run their commands there. There is probably a better way to do this, but that's what just came to my mind. Note that we can't use source replacement for path dependencies, Cargo doesn't support that (although that would make it way easier...).

DavHau:
What stop us from just patching the Cargo.toml at build time (or vendoring time) and replacing the relative paths with absolute store paths of the dependency source?
Or if absolute paths are not allowed, we can generate a `./deps` directory with symlinks to absolute paths.

yusdacra:
> What stop us from just patching the Cargo.toml at build time (or vendoring time) and replacing the relative paths with absolute store paths of the dependency source? Or if absolute paths are not allowed, we can generate a `./deps` directory with symlinks to absolute paths.

Patching should be fine! Just didn't think of that when I wrote that :sweat_smile: I think we can patch the `Cargo.toml` easily in `preConfigure`. The translator can write a list of paths that will need to be patched to `subsystemAttrs`, and then we can replace the relative paths with the absolute paths as provided in the `subsystemAttrs`, it should just be a simple matter of `substituteInPlace` for each `rel path -> absolute path` in that list.

PR #115, opened 2022-04-22
Attempt to fix #113

### LINKED ISSUE #113: Rust builds - referencing path dependencies outside of a workspace.

yusdacra:
This WIP PR implements a way to fix #113. As mentioned in #114's last comments. Note this doesn't work yet, I'm just putting this here so it is seen.

@DavHau would be nice if you could take a look at this, and see the general gist of what I meant in my comment on #114.

### DIFF
--- a/src/discoverers/rust/default.nix
+++ b/src/discoverers/rust/default.nix
@@ -8,22 +8,24 @@
   discoverCrates = {tree}: let
     cargoToml = tree.files."Cargo.toml".tomlContent or {};
 
-    subdirProjects =
+    subdirCrates =
       l.flatten
       (l.mapAttrsToList
         (dirName: dir: discoverCrates {tree = dir;})
         (tree.directories or {}));
   in
     if cargoToml ? package.name
-    then [
-      {
-        inherit (cargoToml.package) name version;
-        inherit (tree) relPath fullPath;
-      }
-    ]
+    then
+      [
+        {
+          inherit (cargoToml.package) name version;
+          inherit (tree) relPath fullPath;
+        }
+      ]
+      ++ subdirCrates
     else subdirCrates;
 
-  _discover = {
+  discoverProjects = {
     tree,
     crates,
   }: let
@@ -33,7 +35,7 @@
       l.flatten
       (l.mapAttrsToList
         (dirName: dir:
-          _discover {
+          discoverProjects {
             inherit crates;
             tree = dir;
           })
@@ -59,7 +61,7 @@
     else subdirProjects;
 
   discover = {tree}:
-    _discover {
+    discoverProjects {
       inherit tree;
       crates = discoverCrates {inherit tree;};
     };
--- a/src/discoverers/rust/default.nix
+++ b/src/discoverers/rust/default.nix
@@ -19,7 +19,7 @@
       [
         {
           inherit (cargoToml.package) name version;
-          inherit (tree) relPath fullPath;
+          inherit (tree) relPath;
         }
       ]
       ++ subdirCrates
--- a/src/discoverers/rust/default.nix
+++ b/src/discoverers/rust/default.nix
@@ -5,13 +5,38 @@
 }: let
   l = lib // builtins;
 
-  discover = {tree}: let
+  discoverCrates = {tree}: let
+    cargoToml = tree.files."Cargo.toml".tomlContent or {};
+
+    subdirProjects =
+      l.flatten
+      (l.mapAttrsToList
+        (dirName: dir: discoverCrates {tree = dir;})
+        (tree.directories or {}));
+  in
+    if cargoToml ? package.name
+    then [
+      {
+        inherit (cargoToml.package) name version;
+        inherit (tree) relPath fullPath;
+      }
+    ]
+    else subdirCrates;
+
+  _discover = {
+    tree,
+    crates,
+  }: let
     cargoToml = tree.files."Cargo.toml".tomlContent;
 
     subdirProjects =
       l.flatten
       (l.mapAttrsToList
-        (dirName: dir: discover {tree = dir;})
+        (dirName: dir:
+          _discover {
+            inherit crates;
+            tree = dir;
+          })
         (tree.directories or {}));
   in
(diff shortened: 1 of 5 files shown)


src/translators/rust/pure/cargo-lock/default.nix
```diff
@@ -171,6 +172,13 @@ in {
       # The structure of this should be defined in:
       #   ./src/specifications/{subsystem}
       subsystemAttrs = rec {
+        replacePathsWithAbsolute = l.listToAttrs (l.map (
+            project: {
+              name = project.relPath;
+              value = "${rootSource}/${project.relPath}";
+            }
+          )
+          discoveredProjects);
```
DavHau, line 181: I guess you might want to filter for project with subsystem `rust` only.
Also I think this will not always work. discoveredProjects only contains projects with a Cargo.lock file. So whenever a project without Cargo.lock is referenced it will be missing here.

Why don't we move some of that logic to the discoverer, so the discovered projects will look like that:
```json
[
  {
    "name": "foo",
    "subsystem": "rust",
    "subsystemInfo": {
      "referredWorkspaces": [
        "rel/path/to/arbitrary/directory/within/source"
      ],
    },
  }
]
```

With this, you will get `referredWorkspaces` as an input to the translator, and you will have all you need, right?

DavHau (commented):
The general approach looks good, but I still think we can get rid of `discoveredProjects`. See comment below.

src/translators/rust/pure/cargo-lock/default.nix
yusdacra: Yeah, I was doing that actually, but didn't push that yet.

Something i noticed with this approach is that for patching the paths we need to create the proper relative path. So for example, in `Cargo.toml`, a path dependency will have the path `../../library`. So we need to be able to create the `../../library` from our list of crates, which means we need to figure out the relative paths to the project we are translating. Then we put those in the `subsystemAttrs`, relative path as the key, and the absolute path as the value, then we just run substitute on those. Do you happen to have any idea on how resolve those relative paths efficiently?

DavHau: I just added a library function `sanitizePath` via b68a9ee61c733aa8aec6059f439e184eb6cf85df.
Now you can do this:
```command
# from the dream2nix source
nix-repl> :lf ./.
Added 14 variables.

nix-repl> tree = lib.dlib.prepareSourceTree {source = ./.;}

nix-repl> subTree = tree.getNodeFromPath "src/fetchers"

nix-repl> lib.dlib.sanitizePath "${subTree.fullPath}/../"
"/nix/store/26k11das6x5n4449sywhdkqrjw3nfzif-dream2nix/src"
```

Does that solve the problem?
There is also `sanitizeRelativePath` to do the same for relative paths.

yusdacra: Yep! I was able to use that to fix how I do things. It seems to be working now. I have updated https://github.com/yusdacra/dream2nix-issue-113, so you can check if it works for you too there.

src/discoverers/rust/default.nix
```diff
@@ -5,13 +5,40 @@
...
+      (l.mapAttrsToList
+        (dirName: dir: discoverCrates {tree = dir;})
+        (tree.directories or {}));
+  in
+    if cargoToml ? package.name
+    then
+      [
+        {
+          inherit (cargoToml.package) name version;
+          inherit (tree) relPath fullPath;
```
DavHau, line 22: No need to put the `fullPath` in the project. you can always re-calculate it via `"${tree.fullPath}/${relPath}"`

DavHau: Oh this isn't actually the project spec. Sorry I was confused. Forget about the previous comment

yusdacra:
I think this should be ready to merge now, if there aren't any other issues! https://github.com/yusdacra/dream2nix-issue-113 works fine so I think the issue should be resolved.

DavHau:
Thanks a lot

PR #116, opened 2022-04-23
ebuild  0.14

happysalada:
Just adds an override for esbuild 0.14.
I'll add more of those if that works for you.

### DIFF
--- a/overrides/nodejs/default.nix
+++ b/overrides/nodejs/default.nix
@@ -370,6 +370,24 @@ in
           };
         in "${esbuild}/bin/esbuild";
       };
+      "add-binary-0.14" = {
+        _condition = satisfiesSemver "~0.14";
+        ESBUILD_BINARY_PATH = let
+          esbuild = pkgs.buildGoModule rec {
+            pname = "esbuild";
+            version = "0.14.38";
+
+            src = pkgs.fetchFromGitHub {
+              owner = "evanw";
+              repo = "esbuild";
+              rev = "v${version}";
+              sha256 = "sha256-rvMi1oC7qGidvi4zrm9KCMMntu6LJGVOGN6VmU2ivQE=";
+            };
+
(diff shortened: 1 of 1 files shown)


overrides/nodejs/default.nix
```diff
@@ -370,6 +370,24 @@ in
           };
         in "${esbuild}/bin/esbuild";
       };
+      "add-binary-0.14" = {
+        _condition = pkg: satisfiesSemver "~0.14" pkg;
```
DavHau, line 374:
```suggestion
        _condition = satisfiesSemver "~0.14";
```

happysalada:
@DavHau let me know if that is what you had in mind.
I'll be adding more as I go along.

DavHau (commented):
currying is king

happysalada:
haha, will remember that one.
All good now.

DavHau:
Wonderful. Thank you very much. This solution looks good. I think we can solve similar problems like that as well.
At some point we should probably start using some input management for these extra sources. Otherwise there is a risk of those being forgotten and never updated.

happysalada:
I completely agree.
I don't know if you've seen the patttern in nixpkgs that uses a pin.json
ideally we should have the same, a json file that stores data like version, sha256 and the like.
So that a script can update those very easily.

DavHau:
I'm a fan of using a flake for managing inputs. We could have a separate flake.nix just for the overrides.
But then again flakes doesn't support custom update logic.
But we have already custom update logic implemented in `./src/fetchers` and `./src/updaters`. Maybe we can glue that together with with flakes somehow.
Instead of `nix flake update` you would run `nix run .#update` which is our custom updater implementation.

happysalada:
that sounds like a good idea.
I haven't seen it where you can update to the next version for example. This sounds like it would exist though.

PR #126, opened 2022-04-24
readme: add more involved override

happysalada:
Adds a more involved example to show what is possible in the readme.

One very weird thing though.
I get this error
```
error: value is a set while a string was expected

       at /nix/store/jrsj7h4ibxlfiv99i5sxmiid3ixmjc8c-source/semver.nix:60:23:

           59|     let
           60|       constraintStr = builtins.replaceStrings [ " " ] [ "" ] constraint;
             |                       ^
           61|       # The common prefix operators
```
when trying to use it as is.
(coming from the utils.semver part).
is the exported version different from the version inside the repo ?

also I would understand if you are uncomfortable exposing things out of the internals. If you don't like that, I could just replace the semver part with a simple version check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,34 @@ Extensive Example `flake.nix`:
             # update attributes
             buildInputs = old: old ++ [ pkgs.hello ];
           };
+          # a more complicated override like esbuild in the repo
+          "add-binary-0.13" = let 
+            # some additional functions are available in utils
+            # run `nix repl flake.nix` and `:lf .` to find out about
+            # all of them
+            utils = inputs.dream2nix.lib.x86_64-linux.utils;
+            # in case you need to use nixpkgs
+            pkgs = inputs.dream2nix.inputs.nixpkgs.legacyPackages.x86_64-linux;
+            in {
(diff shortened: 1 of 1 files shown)


DavHau:
I think this bloats the readme too much. Maybe this should better go into `docs/override-system.md`?

Yeah, the `utils.satisfiesSemver` is different. It is the upstream variant coming from poetry2nix. In the overrides file, it is wrapped so it accepts a `pkg` as an argument instead of a version.
We should definitely fix that and make a clear distinction between those two variants.

Also I'd like to refactor the overrides so each override is in a single file. That will allow us to implement a `nix run .#upstreamOverrides` which opens a PR with the users overrides. I feel like managing everything in a big file makes it harder to merge overrides from different sources together and therefore users don;t upstream their valuable work.

happysalada:
definitely adds a bit too much to the main readme.
It might belong more into the contributing doc part.
if the api is going to change, then I'll just wait until the api changes before making a change to the doc.

happysalada:
of forgot to say, putting each overrides into their own file sounds interesting. Happy to test the upstreaming.

PR #127, opened 2022-04-24
electron: fix darwin build

happysalada:
I'm not sure how to disable it, but it looks like by default the evaluation on darwin is impure for electron.
the the resulting directory looks exactly right, and the only thing to do is change the electron wrapper.
Let me know if that works for you.

electron is working, just showing a `Electron could not be found. No hard resets for you!` warning before running

### DIFF
--- a/overrides/nodejs/default.nix
+++ b/overrides/nodejs/default.nix
@@ -288,17 +288,23 @@ in
     in {
       add-binary = {
         overrideAttrs = old: {
-          postPatch = ''
-            cp -r ${getElectronFor "${old.version}"}/lib/electron ./dist
-            chmod -R +w ./dist
-            echo -n $version > ./dist/version
-            echo -n "electron" > ./path.txt
-          '';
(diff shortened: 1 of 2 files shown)


happysalada:
This for sure improves the situation but I'm not sure it completely fixes it.
The app starts with a blank screen and it tries to connect to the server
```
(node:58093) electron: Failed to load URL: http://localhost:3333/ with error: ERR_CONNECTION_REFUSED
(Use `Electron --trace-warnings ...` to show where the warning was created)
VITE NOT READY, WILL TRY AGAIN IN 1000ms 3333
```
the server is probably using vite. It seems something is not getting started.
I think it's safe to merge that as a first step, but some more things might be needed.

DavHau:
Sounds good. We can do further optimizations later.
Maybe try `mattermost-desktop-5.0.1`. This version once worked in dreampkgs on linux. This version should just build with out overrides.

PR #133, opened 2022-04-28
don't mv ../.cargo/config if it doesn't exist

lelongg:
I have dream2nix complaining about missing `../.cargo/config` and I don't know if it's the right fix but it seems ok


yusdacra (approved):
Looks fine to me! I was considering doing this, but I was thinking the file would always be available. It would be neat if you could show with what project it's not found, but it's not necessary.

PR #135, opened 2022-05-01
Initial stab at extensible subsystems

jaen:
Hi, I come bearing actual discussion material regarding allowing `dream2nix` to be extended from outside.
This is the least amount of changes that allowed me to specify new discoverers, translators, fetchers and builders using the `config` attribute. Right now this is pretty un-fancy – for example it would be tedious to combine multiple extensions together, as you would have to manually merge attrsets together. An overlay-based mechanism, with each extension providing an overlay adding language-subsystem-specific functionality would probably make more sense, but wanted to keep it as simple as possible and have really no idea how much changes would be required to accommodate that.

Now, I don't quite understand what I'm doing here, so some changes might've been unnecessary or better accomplished another way – for example I had to push down the `dlib` like that and thread `config` through more places, because otherwise I ended up with translators that alternatively had the specified config or not, probably closing over different things – but hopefully at least better explains what I thought `dream2nix` being extensible might be useful and could be a starting port for a discussion.
 
Here's a repo using it to get a barebones Ruby subystem running to show it actually works:
https://github.com/jaen/iwai

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -63,8 +63,8 @@
 
     lib = nixpkgs.lib;
 
-    # dream2nix lib (system independent utils)
-    dlib = import ./src/lib {inherit lib;};
+    # # dream2nix lib (system independent utils)
+    # dlib = import ./src/lib {inherit lib;};
 
     supportedSystems = ["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"];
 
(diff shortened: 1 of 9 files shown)


DavHau:
Some thoughts:
  - Currently still a lot of things are changing in dream2nix. Therefore I cannot guarantee a stable interface for external extensions.
  - Maybe better have an argument `extensions` for dream2nix.lib.init, like:
    ```nix
    dream2nix.lib.init {
      extensions.translators.ruby.pure.bundler-impure = ...;
      extensions.builders.ruby = ...;
    }
    ```
 - Currently the structure of translators and builders is different in dream2nix. For translator the translator name is derived from the file name of the nix file and for builders the names are defined in builders/default.nix. We should probably unify that structure to be the same for every `module` kind.

jaen:
Here's the module version – https://github.com/nix-community/dream2nix/pull/153

jaen:
Closing in favour of #155

PR #137, opened 2022-05-03
feat: allow ref to be specified in git source

### LINKED ISSUE #66: feat: let users be able to specify a `ref` alongside a `rev` for the `git` fetcher
This is good for not using `allRefs`, which can slowdown the fetch if there are many refs. Translators that have access to the `ref` alongside a `rev` can use this to not require `allRefs` usage (for example, `Cargo.lock` files can include a `branch` or a `tag` for git dependencies, alongside the usual git revision).

yusdacra:
This resolves #66.

### DIFF
--- a/src/fetchers/git/default.nix
+++ b/src/fetchers/git/default.nix
@@ -5,6 +5,11 @@
   ...
 }: let
   b = builtins;
+
+  # check if a string is a git ref
+  isGitRef = b.match "refs/(heads|tags)/.*";
+  # check if a string is a git rev
+  isGitRev = b.match "[a-f0-9]*";
 in {
   inputs = [
     "url"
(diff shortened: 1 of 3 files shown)


src/fetchers/git/default.nix
```diff
@@ -18,14 +23,21 @@ in {
...
     else let
       b = builtins;
 
       refAndRev =
-        if b.match "refs/(heads|tags)/.*" inp.rev != null
+        # if the source specifies a ref, then we add both the ref and rev
+        if inp ? ref
+        then {inherit (inp) rev ref;}
+        # otherwise check if the rev is a ref, if it is add to ref
+        else if b.match "refs/(heads|tags)/.*" inp.rev != null
```
DavHau, line 38: `isGitRef inp.rev` is computed a second time here. Maybe use let..in to store the result?

src/fetchers/git/default.nix
```diff
@@ -34,7 +46,8 @@ in {
           (refAndRev
             // {
               inherit url;
-              allRefs = true;
+              # disable fetching all refs if the source specifies a ref
+              allRefs = !(inp ? ref);
```
DavHau, line 50: Maybe check for `ref == null` as well? Just in case.

DavHau, line 50:
```suggestion
              allRefs = inp.ref or null == null;
```

src/fetchers/git/default.nix
```diff
@@ -17,15 +22,24 @@ in {
...
     then throw ''rev must either be a sha1 revision or "refs/heads/branch-name" or "refs/tags/tag-name"''
+    else if isGitRef (inp.ref or "") == null
+    then throw ''ref must be in either "refs/heads/branch-name" or "refs/tags/tag-name" format''
     else let
       b = builtins;
 
       refAndRev =
-        if b.match "refs/(heads|tags)/.*" inp.rev != null
+        # if the source specifies a ref, then we add both the ref and rev
+        if inp ? ref
```
DavHau, line 37: null check as well?

PR #143, opened 2022-05-08
Resolve `path` sources relative to the source root

jaen:
Changes the built-in path fetcher to resolve relative paths relative to the source root, allowing to build packages from the same repository the lockfile resides in. A test package that doesn't build without those changes and starts building with them was added.

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -200,7 +200,7 @@
   # fetch only sources and do not build
   fetchSources = {
     dreamLock,
-    sourceRoot,
+    sourceRoot ? null,
     fetcher ? null,
     extract ? false,
     sourceOverrides ? oldSources: {},
@@ -342,7 +342,7 @@
 
   makeOutputsForDreamLock = {
     dreamLock,
-    sourceRoot,
+    sourceRoot ? null,
     builder ? null,
     fetcher ? null,
     inject ? {},
--- a/src/default.nix
+++ b/src/default.nix
@@ -200,6 +200,7 @@
   # fetch only sources and do not build
   fetchSources = {
     dreamLock,
+    sourceRoot,
     fetcher ? null,
     extract ? false,
     sourceOverrides ? oldSources: {},
@@ -213,7 +214,7 @@
       else args.fetcher;
 
     fetched = fetcher rec {
-      inherit sourceOverrides;
+      inherit sourceOverrides sourceRoot;
       defaultPackage = dreamLock._generic.defaultPackage;
       defaultPackageVersion = dreamLock._generic.packages."${defaultPackage}";
       sources = dreamLock'.sources;
@@ -341,6 +342,7 @@
 
   makeOutputsForDreamLock = {
     dreamLock,
+    sourceRoot,
     builder ? null,
     fetcher ? null,
     inject ? {},
@@ -369,7 +371,7 @@
 
     fetchedSources =
       (fetchSources {
-        inherit dreamLock sourceOverrides;
+        inherit dreamLock sourceOverrides sourceRoot;
         fetcher = fetcher';
       })
       .fetchedSources;
@@ -552,6 +554,7 @@
     outputsForProject = proj: let
       outputs = makeOutputsForDreamLock {
         inherit inject packageOverrides;
+        sourceRoot = source;
         builder = proj.builder or null;
         dreamLock = proj.dreamLock;
         sourceOverrides = oldSources:
--- a/src/fetchers/default-fetcher.nix
+++ b/src/fetchers/default-fetcher.nix
@@ -11,7 +11,7 @@
   defaultPackageVersion,
   sourceOverrides,
   sources,
-  sourceRoot,
+  sourceRoot ? null,
   ...
 }: let
   l = lib // builtins;
@@ -27,11 +27,13 @@
         then
           if l.isStorePath (l.concatStringsSep "/" (l.take 4 (l.splitString "/" source.path)))
           then source.path
-          else if l.pathExists (l.traceVal "${sourceRoot}/${source.path}")
-          then "${sourceRoot}/${source.path}"
           else if name == source.rootName && version == source.rootVersion
           then throw "source for ${name}@${version} is referencing itself"
-          else "${overriddenSources."${source.rootName}"."${source.rootVersion}"}/${source.path}"
+          else if source.rootName != null && source.rootVersion != null
+          then "${overriddenSources."${source.rootName}"."${source.rootVersion}"}/${source.path}"
+          else if sourceRoot != null
+          then "${sourceOverrides}/${source.path}"
+          else throw "${name}-${version}: cannot determine path source"
         else if fetchers.fetchers ? "${source.type}"
         then
           fetchSource {
--- a/src/fetchers/default-fetcher.nix
+++ b/src/fetchers/default-fetcher.nix
@@ -11,6 +11,7 @@
   defaultPackageVersion,
   sourceOverrides,
   sources,
+  sourceRoot,
   ...
 }: let
   l = lib // builtins;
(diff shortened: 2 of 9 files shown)


src/fetchers/default-fetcher.nix
```diff
@@ -26,6 +27,8 @@
         then
           if l.isStorePath (l.concatStringsSep "/" (l.take 4 (l.splitString "/" source.path)))
           then source.path
+          else if l.pathExists (l.traceVal "${sourceRoot}/${source.path}")
```
yusdacra, line 30: Why is the path checked if exists here? The sourceRoot should be used if `rootName` and `rootVersion` are `null` (so in the `else` branch, you can add a check for it there).

src/fetchers/default-fetcher.nix
```diff
@@ -11,6 +11,7 @@
   defaultPackageVersion,
   sourceOverrides,
   sources,
+  sourceRoot,
```
yusdacra, line 14: This should be optional rather than required.

```suggestion
  sourceRoot ? null,
```
an error should be thrown if `sourceRoot` needs to be used but it is `null`.

src/default.nix
```diff
@@ -341,6 +342,7 @@
 
   makeOutputsForDreamLock = {
     dreamLock,
+    sourceRoot,
```
yusdacra, line 345: This should be optional.

```suggestion
    sourceRoot ? null,
```

src/default.nix
```diff
@@ -200,6 +200,7 @@
   # fetch only sources and do not build
   fetchSources = {
     dreamLock,
+    sourceRoot,
```
yusdacra, line 203: This should be optional.

```suggestion
    sourceRoot ? null,
```

yusdacra (changes requested):

src/fetchers/default-fetcher.nix
jaen: Okay, then I guess I have misunderstood you, I'll change the conditions like you suggest.

IIUC in the case when `rootName` is set, the relative path should be working already (relativised to the `rootName` source) and if that then is relative, then only then it should be handled by what I added.

src/fetchers/default-fetcher.nix
jaen: I wanted to make it optional at first, but then I thought I've seen something downstream that has it required, so I figured it's better to make it required as well.

But maybe I confused it with a single package's source. I'll adjust it how you suggest.

src/fetchers/default-fetcher.nix
yusdacra: > IIUC in the case when rootName is set, the relative path should be working already (relativised to the rootName source) and if that then is relative, then only then it should be handled by what I added. Right?

Basically yeah. If the `rootName` exists, then the `"${overriddenSources."${source.rootName}"."${source.rootVersion}"}/${source.path}"` code should be used. If not, the `sourceRoot` can be used for figuring out the absolute path.

src/fetchers/default-fetcher.nix
yusdacra: These should be optional mostly because this argument didn't exist before (to prevent code from breaking). And a `sourceRoot` may not always exist (for example, when building from a dream-lock directly).

yusdacra:
@jaen I pushed my changes to the branch. I checked to see if it works or not on some Rust projects, and it seems to work fine. After @DavHau reviews this, and if it also works with your Ruby stuff, we can merge it.

jaen:
@yusdacra ah, sorry, I actually did those changes you requested, but then I kind of ran away with better validation errors for when sources are incorrectly specified and didn't have enough time to finish it during the week. I actually tested my changes with nodejs mostly (as included with the tests) and I'm not quite sure Ruby gem built cleanly with just those changes (but it's an issue for the Ruby builder itself anyway), but I can doublecheck,

DavHau:
Currently the idea behind path sources is that those are always relative to the source of another package. There always needs to be a `parent` for a path source which we express via `rootName` and `rootVersion`.
The mistake for the nodejs translators was that those didn't set these variables.
I fixed this via https://github.com/nix-community/dream2nix/pull/146.
After this, the path sources will point to a parent source as expected by the framework. That parent source might not exist in the dream lock. But this is OK, as it can be added via source override.
`makeFlakeOutputs` adds the main source via sourceOverrides automatically:
https://github.com/nix-community/dream2nix/blob/9003aed816a3237342d85250c347622647f6cfb3/src/default.nix#L539-L548

jaen:
I'll have to test this, from cursory look at the PR I don't quite understand how this fixes my use-case (package that comes from a place relative to the source root). That override _maybe_ looks like it's what would do this, but I would have to look at it more closely first.

jaen:
@DavHau yeah, but that's not where I came across the issue initially – the problem turned up in my Ruby subsystem, where I try to bootstrap the Ruby impure translator with dream2nix alongside the subsystem. The sources sit next to the lock and by default I end up with a in-store path to the sources when I run the translator – which won't fly, because someone else won't have my store paths on their PC. So I relativise the path to the project source and then I resolve it against the `sourceRoot` as you can see here. The difference here is that my package is the top-level one, why those are nested due to nodejs workspace support – this is the closest I could get to my Ruby example with existing translators. Maybe I'm doing something wrong in my translator or maybe this case is different (I'm not quite sure why would it be, but I don't understand the internals of dream2nix well enough yet to be sure) and I would need to test it first.

From what I understand of your changes, it's quite probable that the default override you're adding will solve that case, but I just want to make sure before deciding that indeed your PR is the way to go — for sure you know the codebase better and nix is hard to wrap your head around when using as a general-purpose language, but I want to be certain that we are on the same page regarding what this needs to support.

Hope that makes sense, I have a feeling this commend is kinda disjointed.

yusdacra:
The problem this solves for the `cargo-lock` translator was this:

While implementing the `cargo-toml` impure translator, I found out that source specs were being removed because they contained path sources which referenced the nix store (which makes sense). This causes the builds to fail, because the vendoring code for Rust builders expect there to be a source spec for every crate/version. So to fix this I would change the `cargo-lock` translator to not output nix store paths, but there is a problem with this for the "default package" crate and crates which are outside the root of this crate. `dream2nix` adds a source override for this "default package" crate in `realizeProjects`, but firstly this only adds a source, not a source spec and secondly this doesn't include crates that are outside that path. To fix this in a simple way I thought it would be a okay idea to basically just let relative path sources that have a relative path and `rootName` / `rootVersion` set to null just have use the source root as a base when making the relative path absolute. This way we also make sure information for all of the path dependencies are stored properly, instead of just omitting them.

DavHau:
OK, got it. Let's better have the sourceRoot as the one single undefined variable of the dream-lock instead of multiple missing sources.

src/fetchers/default-fetcher.nix
```diff
@@ -28,7 +29,11 @@
           then source.path
           else if name == source.rootName && version == source.rootVersion
           then throw "source for ${name}@${version} is referencing itself"
-          else "${overriddenSources."${source.rootName}"."${source.rootVersion}"}/${source.path}"
+          else if source.rootName != null && source.rootVersion != null
+          then "${overriddenSources."${source.rootName}"."${source.rootVersion}"}/${source.path}"
+          else if sourceRoot != null
+          then "${sourceOverrides}/${source.path}"
```
DavHau, line 35:
```suggestion
          then "${sourceRoot}/${source.path}"
```

src/fetchers/default-fetcher.nix
```diff
@@ -11,6 +11,7 @@
   defaultPackageVersion,
   sourceOverrides,
   sources,
+  sourceRoot ? null,
```
DavHau, line 14: I don't think we should put much effort into maintaining downward compatibility. The dream2nix project is in early stage development and marked as `WIP`. Anyone depending on it should count with breaking changes on any update.
It unnecessarily increases the complexity of our code and we have no clear policy on when to remove the extra code.

But if you prefer to have it like this, I'm fine with it as well.

yusdacra: It being optional should allow people to call default fetcher without a `sourceRoot`. This should help if someone only uses the `dream-lock.json` file to build it, without passing a "source root" (because they know there is none). For example `dream2nix-index-lib` relies on being able to only build from the dream-lock files, so if this was required it wouldn't work.

DavHau:
Alright, I will rebase this then on https://github.com/nix-community/dream2nix/pull/146 and merge it.

DavHau:
Thanks for your effort and patience,

PR #144, opened 2022-05-09
feat: implement cargo-toml impure translator for rust subsystem

yusdacra:
This implements a new impure `cargo-toml` translator that can translate Cargo projects that don't have a `Cargo.lock` file. In addition, it makes the `cargo-lock` translator not put absolute paths for fixing up path dependency paths in a `Cargo.toml`, instead relative paths are used now, because absolute paths break the generated dream-lock. Rust discoverer is also updated to support the new translator.

~~This is a draft because it doesn't seem to be able to build stuff properly (`ripgrep` doesn't build).~~ `ripgrep` builds fine now.

### DIFF
--- a/src/builders/rust/utils.nix
+++ b/src/builders/rust/utils.nix
@@ -16,17 +16,14 @@ in rec {
 
   # Generates a script that replaces relative path dependency paths with absolute
   # ones, if the path dependency isn't in the source dream2nix provides
-  replaceRelativePathsWithAbsolute = {
-    paths,
-    src,
-  }: let
+  replaceRelativePathsWithAbsolute = {paths}: let
     replacements =
       l.concatStringsSep
       " \\\n"
       (
         l.mapAttrsToList
         (
-          from: rel: "--replace '\"${from}\"' '\"${dlib.sanitizePath "${src}/${rel}"}\"'"
+          from: rel: ''--replace "\"${from}\"" "\"$TEMPDIR/$sourceRoot/${rel}\""''
         )
         paths
       );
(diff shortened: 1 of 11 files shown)


yusdacra:
I'm marking as draft again because this branch breaks nix-cargo-integration's tests. Will see to that soon.

yusdacra:
Turns out the issue was related to path sources with nix store paths being removed from the dream-lock which makes sense, but I'm not sure why it hasn't popped up before. Anyhow to (properly) fix this #143 is needed, because otherwise path source specs with nix store paths will be removed, which breaks Rust vendoring. It will also break for crates outside of project source, #143 allows fixing that too.

DavHau:
Could you rebase this on main to see if all tests succeed?
Also please add an example under `./examples` which uses this translator.

DavHau:
Could you rebase this on main to see if all tests succeed?
Also please add an example under `./examples` which uses this translator.

src/builders/rust/utils.nix
```diff
@@ -15,17 +18,83 @@ in rec {
...
       );
   in ''
     substituteInPlace ./Cargo.toml \
       ${replacements}
   '';
+
+  # Script to write the Cargo.lock if it doesn't already exist.
+  writeCargoLock = ''
+    if [ ! -f "$PWD/Cargo.lock" ]; then
+      cp -v ${cargoLock} "$PWD/Cargo.lock"
```
DavHau, line 40: This triggers the build of `cargoLock` to produce the file in the nix store. This will be counted as import from derivation, because nix has to build this derivation (via builtins.toFile) in order to continue the evaluation.

Alternatively you could insert the lock file as a string here, and therefore not forcing nix to create a store path during eval time.

PR #146, opened 2022-05-17
Path improvements

DavHau:
Alternative fix to https://github.com/nix-community/dream2nix/pull/143

The problem here was that we changed the path source some time ago to expect a `rootName` and `rootVersion`. The package-lock and yarn-lock translators simply didn't set these values, which is fixed by this PR

@jaen @yusdacra

### DIFF
--- a/examples/path-example/flake.lock
+++ b/examples/path-example/flake.lock
@@ -1,0 +1,210 @@
+{
+  "nodes": {
+    "alejandra": {
+      "inputs": {
+        "flakeCompat": "flakeCompat",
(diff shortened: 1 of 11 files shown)


tgunnoe:
Will this fix an issue I've been experiencing with trying to filter out flake/nix files so changes to them wouldn't update the nix hash of a package that hasn't changed?  The problem is that the package builds from dream2nix have the source directory as an input twice -- once with the flitered items (eg, just the entire repo), and again without the files that are filtered.

PR #151, opened 2022-05-20
feat: categorize builders like translators

yusdacra:
This gives builders types like translators, namely `ifd` and `pure`. I implemented this similar to how translators are done.


DavHau:
> Yeah, I agree sweat_smile. Do you want me to try taking a shot at that for both translators and builders?

Yes, that would be amazing. Eventually we should load all our module types the same way, so we can have a sane extension interface for all types of modules.

DavHau:
related: https://github.com/nix-community/dream2nix/pull/135

jaen:
Also related: https://github.com/nix-community/dream2nix/pull/153

Issue #152, opened 2022-05-20
yarn-lock: versions like `npm:package-name@1.2.3` not supported


DavHau:
see https://github.com/nix-community/dream2nix/pull/149#issuecomment-1132199197

PR #159, opened 2022-05-24
yarn-lock: handle `npm:` directive in version

nrdxp:
Fixes #152

PR #160, opened 2022-05-25
Fix: #152 parse npm:package-name@ver correctly

tgunnoe:
Builds on #159
@DavHau I know everything is still a WIP, but please add any suggestions if desired. 

I am assuming that for all cases we should treat `"package": "npm:package-in-npm@version-in-npm"` rewritten as `"package-in-npm": "version-in-npm"` for dream2nix's purposes.

Fixes https://github.com/nix-community/dream2nix/issues/152

### DIFF
--- a/src/translators/nodejs/pure/yarn-lock/default.nix
+++ b/src/translators/nodejs/pure/yarn-lock/default.nix
@@ -148,6 +148,13 @@
           # @matrix-org/olm@https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.3.tgz
           else if lib.hasInfix "@https://" rawObj.yarnName
           then lib.head (lib.splitString "@https://" rawObj.yarnName)
+          else if lib.hasInfix "npm:" rawObj.version
+          then let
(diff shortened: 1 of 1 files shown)


nrdxp:
Can confirm that this fully resolves the lockfile generation problems I was having.

DavHau:
Thanks. Is there an example we can add to `generateUnitTestsForProjects` which didn't work before that change and now works?

tgunnoe:
@DavHau I will get back to you with the change; It looks like this is only the case were scoped dependencies in `package.json` use the `npm:` version tag.  I assumed it was for any dependency and not just scoped, so let me figure out why that is so first.

DavHau:
The file was moved by #155. I applied this patch to the new file. Still, adding a project to the unit tests that requires this fix would be nice.

tgunnoe:
@DavHau this is a minimal project that would break it before this fix:  https://github.com/tgunnoe/dream2nix-yarn-npm-test
but the original comes from https://github.com/input-output-hk/cardano-js-sdk/tree/flakify (may not exist in the future after a merge) with `nix run .#cardano-sdk.resolve`

PR #161, opened 2022-05-27
feat: implement makeOutputs that outputs `system`less structure 

yusdacra:
This implements a `makeDream2nix` function for creating a dream2nix instance with `externalSources` and `externalPaths` already provided. This is useful because otherwise you have to create these yourself, and since this is a flake lib function it's more convenient than `import`ing the `src/default.nix` (`dream2nix.lib.makeDream2nix {inherit pkgs;}` as opposed to `import "${dream2nix}/src" {inherit pkgs;}`). It is also useful for users of dream2nix that don't want dream2nix to "take over" the flake outputs, but still want to take advantage of the flake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,12 +27,12 @@ The intention is to integrate many existing 2nix converters into this framework,
 (Currently only nodejs and rust packaging is supported)
 
 1. Make sure you use a nix version >= 2.4 and have `experimental-features = "nix-command flakes"` set.
-1. Navigate to to the project intended to be packaged and initialize a dream2nix flake:
+2. Navigate to to the project intended to be packaged and initialize a dream2nix flake:
     ```command
       cd ./my-project
       nix flake init -t github:nix-community/dream2nix#simple
     ```
-1. List the packages that can be built
+3. List the packages that can be built
     ```command
       nix flake show
     ```
@@ -43,13 +43,9 @@ Minimal Example `flake.nix`:
 {
   inputs.dream2nix.url = "github:nix-community/dream2nix";
   outputs = { self, dream2nix }@inputs:
-    let
-      dream2nix = inputs.dream2nix.lib.init {
-        # modify according to your supported systems
-        systems = [ "x86_64-linux" ];
-        config.projectRoot = ./. ;
-      };
-    in dream2nix.makeFlakeOutputs {
+    dream2nix.lib.makeFlakeOutputs {
+      systems = ["x86_64-linux"];
+      # config.projectRoot defaults to source
       source = ./.;
     };
 }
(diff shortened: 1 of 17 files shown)


tgunnoe:
Oh nice, I've been wanting to look at more customization of flake outputs! I will test it out.  Thank you.

DavHau:
I'm not sure I understand why this is needed? Is initializing the framework too complex?
I feel the current interface might be too messy. Maybe we should re-factor it completely and make it easy to use by default instead of adding a helper?
I'm really open for changing the whole interface completely. Maybe outputting flake-like structure by default was a mistake as well and we should go back to only ever initialize dream2nix with one system eg. one nixpkgs instance.

yusdacra:
> I'm not sure I understand why this is needed? Is initializing the framework too complex?

The `init` interface doesn't work well when you have some `pkgs` already and just want to use that. And as I mentioned, `import`ing the `src/default.nix` isn't really a solution because then you run into the issue of `externalPaths` and `externalSources` not being available, so this also fixes that.

> I'm really open for changing the whole interface completely. Maybe outputting flake-like structure by default was a mistake as well and we should go back to only ever initialize dream2nix with one system eg. one nixpkgs instance.

I personally like initializing stuff with just a pkgs instance and any other configuration arguments, because usually I will just have a `forPkgs` or similar. And that is also more extensible than dream2nix taking over the entire flake outputs, because otherwise you'll have to change basically all of your `flake.nix` to add new stuff to outputs (*without* hardcoding systems and importing new nixpkgs, you could just do that of course but its messy imo). So I would be fine with that, not sure what everyone else thinks though.

DavHau:
> The init interface doesn't work well when you have some pkgs already and just want to use that

DavHau:
> The init interface doesn't work well when you have some pkgs already and just want to use that

The current init interface supports passing `pkgs` instead of `system`.
```nix
dream2nix.init {
  inherit pkgs;
  config = ...;
}
```

Does that solve the current issue?

yusdacra:
> > The init interface doesn't work well when you have some pkgs already and just want to use that
> 
> The current init interface supports passing `pkgs` instead of `system`.
> ```nix
> dream2nix.init {
>   inherit pkgs;
>   config = ...;
> }
> ```
> 
> Does that solve the current issue?

Doesn't that only expose `makeFlakeOutputs`? I feel it would be better if it exposed the dream2nix instance so people can use `realizeProjects` directly (probably also alias that to `makeOutputs`). `makeFlakeOutputs` outputs with systems which doesn't make sense if you are passing a `pkgs` I think.

DavHau:
> I personally like initializing stuff with just a pkgs instance and any other configuration arguments, because usually I will just have a forPkgs or similar. And that is also more extensible than dream2nix taking over the entire flake outputs, because otherwise you'll have to change basically all of your flake.nix to add new stuff to outputs (without hardcoding systems and importing new nixpkgs, you could just do that of course but its messy imo). So I would be fine with that, not sure what everyone else thinks though.

I totally agree. I think I went way to far trying to simplify things but actually making it more complicated. If people don't want to deal with creating flake outputs themselves they can just use flake-utils. No need to have that logic in dream2nix.

DavHau:
> Doesn't that only expose `makeFlakeOutputs`? I feel it would be better if it exposed the dream2nix instance so people can use `realizeProjects` directly (probably also alias that to `makeOutputs`). `makeFlakeOutputs` outputs with systems which doesn't make sense if you are passing a `pkgs` I think.

Oh true. But still I don't like the idea that we have a second `init` function called `makeDream2nix`. We should fix our current init function to have desirable behavior.

yusdacra:
> > Doesn't that only expose `makeFlakeOutputs`? I feel it would be better if it exposed the dream2nix instance so people can use `realizeProjects` directly (probably also alias that to `makeOutputs`). `makeFlakeOutputs` outputs with systems which doesn't make sense if you are passing a `pkgs` I think.
> 
> Oh true. But still I don't like the idea that we have a second `init` function called `makeDream2nix`. We should fix our current init function to have desirable behavior.

Yeah, that's fine. I went with a seperate function because I forgot that behaviour existed (is it documented anywhere?).

DavHau:
> is it documented anywhere?

Nope ;)

yusdacra:
I just added a readme documentation for the init usage with pkgs, but we should probably also add a full example into `examples` showcasing the same thing...

yusdacra:
So I changed quite a lot of stuff now, I realized `realizeProjects` doesn't add `resolveImpure` so I moved most outputs stuff from `src/lib.nix` to `src/default.nix` and called it `makeOutputs`, which adds `resolveImpure` too. I also added an example that uses it.

tgunnoe:
> The current init interface supports passing `pkgs` instead of `system`.
> 
> ```nix
> dream2nix.init {
>   inherit pkgs;
>   config = ...;
> }
> ```
> 
> Does that solve the current issue?

Using init in this way with a multiarch pkgs did not work for me.  At least a few months ago.  It could have been a bug? but indeed it was complex to find the culprit.



> Oh true. But still I don't like the idea that we have a second `init` function called `makeDream2nix`. We should fix our current init function to have desirable behavior.

I also agree!

tgunnoe:
So if you use `pkgs` in `init`, only `makeOutputs` is available. And when you use `systems` in `init`, `makeFlakeOutputs` works as it does now. This works for me

src/lib.nix
```diff
@@ -69,26 +69,29 @@
...
-
-    defaultApp =
-      forAllSystems
-      (system: pkgs:
-        dream2nixFor."${system}".apps.flakeApps.dream2nix);
-  };
+  in
+    if pkgs != null
+    then dream2nixFor."${makePkgsKey pkgs}"
+    else {
```
DavHau, line 75: I think it is confusing if init results in different structure, depending on the input.
Maybe `makeFlakeOutputs` should be removed entirely, or if kept, moved to the same level as `makeOutputs`.
Implementation wise I think makeFlakeOutputs should just be a wrapper around `makeOutputs`.

DavHau: One argument for keeping makeFlakeOutputs is, that it allows us to have a very very simple example flake in the readme.
Also, having the kapability to quickly generate flake structure can come in handy for future CLI tools, where one wants to quickly `flakify` some arbitrary project via one command.

yusdacra: `makeFlakeOutputs` being on the same level as `makeOutputs` doesn't really make sense because `makeOutputs` is in `src/default.nix`, which requires a nixpkgs pkgs set. I think we should change `makeFlakeOutputs` so that it initializes the dream2nix instance(s) and outputs in one call (instead of needing to call init and then makeFlakeOutputs). This would make it so that `init` takes only a pkgs (or one system too maybe, for ease of use). I think this is better than the current `makeFlakeOutputs` usage; it already "takes over" the flake outputs, but you need to go to the trouble of initializing dream2nix and then using that, which doesn't make sense to me if it already takes over the outputs (meaning you wont use the instance again, only once), you should just be able to do it in one call.

DavHau: OK, makes sense. I'm fine with all of what you just said.

README.md
```diff
@@ -132,6 +132,22 @@ Extensive Example `flake.nix`:
...
+Initializing a dream2nix from a nixpkgs set:
+```nix
+let
+  system = "x86_64-linux";
+  # in this case, 'nixpkgs' is a nixpkgs flake
+  pkgs = nixpkgs.legacyPackages.${system};
+  # 'dream2nix' is a dream2nix flake
+  d2n = dream2nix.lib.init {inherit pkgs;};
+in
+  d2n.realizeProjects {
```
tgunnoe, line 144: Should this be accessible, if `makeOutputs` is `realizedProjects // apps.resolveImpure`? Is there a case where you'd use realizeProjects and not makeOutputs?

src/lib.nix
yusdacra: I pushed these changes to the PR just now.

README.md
yusdacra: I think we should keep it just to offer more options. We may add more stuff to `makeOutputs` or do some changes with the outputs, so it's probably better to offer a way to only get the builder outputs and nothing else.

DavHau (commented):
Thanks. That's some lovely refactoring that was overdue since quite some time.
I think eventually, we should get rid of `default.nix` completely, because it gives the wrong impression that the project can be used without flakes, which isn't actually the case. Instead we should put a default.nix which uses the flake via `flake-compat`. But this can be done some other time.

README.md
```diff
@@ -59,76 +55,96 @@ Extensive Example `flake.nix`:
 ```nix
 {
-  outputs = { self, dream2nix }@inputs:
+  outputs = { self, dream2nix }:
```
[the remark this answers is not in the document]
yusdacra: Yeah, that makes sense, I changed it back to use `makeFlakeOutputs` and added a link for the example.

README.md
```diff
@@ -43,13 +43,9 @@ Minimal Example `flake.nix`:
...
-    let
-      dream2nix = inputs.dream2nix.lib.init {
-        # modify according to your supported systems
-        systems = [ "x86_64-linux" ];
-        config.projectRoot = ./. ;
-      };
-    in dream2nix.makeFlakeOutputs {
+    dream2nix.lib.makeFlakeOutputs {
+      systems = ["x86_64-linux"];
+      # config.projectRoot defaults to source
```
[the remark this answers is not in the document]
yusdacra: Yeah got it. I was thinking it would be convenient for most repos, but yeah it is probably best to explicitly set that so users don't get confused later on.

PR #162, opened 2022-05-29
fix: expose all builder outputs

yusdacra:
This makes a change to expose all builder outputs in `realizeProjects`.

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -556,20 +556,9 @@
           (outputs.packages or {});
       };
 
-    projectOutputs =
-      l.map
-      (proj: outputsForProject proj)
-      translatedProjects;
+    projectOutputs = l.map outputsForProject translatedProjects;
 
-    mergedOutputs =
-      l.foldl'
-      (all: outputs:
-        all
-        // {
-          packages = all.packages or {} // outputs.packages;
-        })
-      {}
-      projectOutputs;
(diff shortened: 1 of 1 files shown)


src/default.nix
```diff
@@ -556,20 +556,9 @@
...
-    mergedOutputs =
-      l.foldl'
-      (all: outputs:
-        all
-        // {
-          packages = all.packages or {} // outputs.packages;
-        })
-      {}
-      projectOutputs;
+    mergedOutputs = l.foldl' l.recursiveUpdate {} projectOutputs;
```
DavHau, line 561: recursiveUpdate on attrsets that contain derivations can be dangerous because derivations are attrsets as well and might end up being merged to a weird result.

yusdacra: Oh right, I keep forgetting about that :sweat_smile: would the latest commit work then?

src/default.nix
```diff
@@ -556,20 +556,16 @@
...
-        // {
-          packages = all.packages or {} // outputs.packages;
-        })
-      {}
-      projectOutputs;
+    projectOutputs = l.map outputsForProject translatedProjects;
+
+    mergedOutputs = let
+      isNotDrvAttrs = val:
+        l.isAttrs val && (val.type or "") != "derivation";
```
DavHau, line 563: `l.isDerivation` ?

DavHau: nevermind. Maybe what you did is more efficient :+1:

PR #164, opened 2022-05-31
stack-lock translator & haskell builder

DavHau:
Implements a builder for haskell and a translator for stack.yaml.lock files.

This is a rough POC which needs to be improved in many places.


src/subsystems/haskell/translators/stack-lock/utils.nix
```diff
@@ -0,0 +1,76 @@
...
+        echo "all-cabal-hashes" seems to be outdated
+        exit 1
+      fi
+    '';
+  in
+    pkgs.runCommand "cabal-json-files" {}
+    (l.concatStringsSep "\n"
+      (l.map (c: convertOne c.name c.version) candidates));
+
+  batchCabalData = candidates: let
```
m1-s, line 67: we could add a doc string to each of the functions in this file

src/subsystems/haskell/builders/default/default.nix
```diff
@@ -0,0 +1,122 @@
...
+          doBenchmark = false;
+
+          libraryToolDepends = libraryHaskellDepends;
+          executableHaskellDepends = libraryHaskellDepends;
+          testHaskellDepends = libraryHaskellDepends;
+          testToolDepends = libraryHaskellDepends;
+
+          libraryHaskellDepends =
+            (with pkgs.haskell.packages.ghc8107; [
+              # Cabal_3_2_1_0
```
m1-s, line 89: whats the meaning of this comment?

src/subsystems/haskell/translators/stack-lock/default.nix
```diff
@@ -0,0 +1,376 @@
...
+      projectSource = "${tree.fullPath}/${project.relPath}";
+      projectTree = tree.getNodeFromPath project.relPath;
+
+      # parse the cabal file
+      cabalFiles =
+        l.filter
+        (l.hasSuffix ".cabal")
+        (l.attrNames projectTree.files);
+
+      cabalFile = projectTree.getNodeFromPath (l.head cabalFiles);
```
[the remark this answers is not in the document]
DavHau: If there is no cabal file, the translator cannot work. Failing in this case is correct. Anyways, the discoverer will not even call this translator if the cabal file is missing.

DavHau:
I will merge this for now. I think maintenance should continue on main branch. This seems usable enough. I'm happy to continue to incorporate any feedback from this thread.

PR #166, opened 2022-06-02
fix(builder/crane): remove code that triggers IFD, remove dependency on full project source for deps only drv

yusdacra:
This makes some changes to make it so that the dependencies-only derivation (when using the crane builder) doesn't capture the full project source path as a dependency in it's `inputSrcs`. This causes very frequent rebuilds, and the full source path is *not* needed (crane already creates a dummy source).

@DavHau it would be nice if you could review the usage of `builtins.unsafeDiscardStringContext` here.


src/subsystems/rust/builders/utils.nix
```diff
@@ -96,7 +97,10 @@ in rec {
       )
       dreamLock.dependencies
     );
-    lock = {inherit package;};
+    lock = utils.toTOML {inherit package;};
   in
-    utils.toTOML lock;
+    # on Nix 2.9.0pre20220527_ec07a70> builtins.toFile doesn't trigger IFD
+    if l.compareVersions l.nixVersion "2.9.0pre20220527_ec07a70" >= 0
```
yusdacra, line 103: I chose this version because it was the one on my system :stuck_out_tongue: using `2.9.0` here does not work for me (as in `compareVersions` then deems my system Nix version too old, which does make sense I guess). I think it should be close enough to when reading `toFile` paths was made non-IFD, but I can get a more accurate version if it's needed.

src/subsystems/rust/translators/cargo-lock/default.nix
```diff
@@ -25,8 +25,23 @@ in {
...
+    # and passes an unrealized source.
+    #
+    # After this we can discard the string context and proceed as usual.
+    #
+    # If this is not done, when using the crane builder, the dependency
+    # derivation will capture the fullPath in it's `inputSrcs`, which causes
+    # frequent rebuilds (and the source is not needed anyways).
+    rootSource = l.unsafeDiscardStringContext (
+      l.seq (l.readFile "${tree.fullPath}/Cargo.toml") tree.fullPath
+    );
```
DavHau, line 43: I think it's better to use `builtins.path {path = "${tree.fullPath}/Cargo.toml", name = "Cargo.toml"}`.
This will make this path content addressed, like if it was coming from a FOD, and therefore will not cause rebuilds, despite the source changing.

src/subsystems/rust/builders/utils.nix
DavHau: In general `nixVersion` branching code should be discouraged. It can lead to reproducibility issues between two different nix versions. imagine you build a cache, but some people will need different store paths because of a different nix version.
Especially in this situation where we don't need this, I would prefer to have this condition removed.
If `nixVersion` is ever used in a condition, then it should be to decide about an `abort` or `throw`, not to change the result  of the evaluation.

src/lib/default.nix
```diff
@@ -287,7 +293,10 @@
     sanitizedRelPath = l.removePrefix "/" (l.toString (l.toPath "/${path}"));
   in
     if absolute
-    then "/${sanitizedRelPath}"
+    then
+      l.path {
+        path = "/${sanitizedRelPath}";
+      }
```
DavHau, line 299: I think there is no need to use `builtins.path` in sanitizePaths. sanitizePaths is just a string operation that normalizes `//` to `/` etc.

src/subsystems/rust/translators/cargo-lock/default.nix
yusdacra: Okay, I made some changes to use `builtins.path` in some places and it seems to fix it for me.

src/subsystems/rust/builders/utils.nix
yusdacra: Hmm yeah that makes sense; I changed it to just use `toFile`.

src/lib/default.nix
```diff
@@ -104,7 +104,10 @@
   # }
   prepareSourceTreeInternal = sourceRoot: relPath: name: depth: let
     relPath' = relPath;
-    fullPath' = "${sourceRoot}/${relPath}";
+    fullPath' = l.path {
+      path = "${sourceRoot}/${relPath}";
+      inherit name;
+    };
```
DavHau, line 110: Do we really need `builtins.path` here? Is there any context inherited from here.
I think we should be careful using `builtins.path`, and make sure that we only use it in places were we really need it.
It is an expensive operation. AFAIU nix copied the whole path each time this is called.

If we really need this on the main source, then it should be done only once, not several times.

yusdacra: Yeah, I was just looking at that. This and the one at sanitizePath doesn't seem to be needed, so I removed them.

DavHau:
It still seems to do IFD:
```command
> nix run .#tests-examples ripgrep -- --no-allow-import-from-derivation --show-trace
error: cannot build '/nix/store/fjngmiq90a1p6kgp79pkmbzcbhd5i726-Cargo.lock.drv' during evaluation because the option 'allow-import-from-derivation' is disabled

       … while realising the context of path '/nix/store/kc4dnq04nnbd72gfymd4z03icnqgcyzk-Cargo.lock'

       at /nix/store/20lp1y524zyr5dflc4wv79ncashgy16k-source/lib/mkDummySrc.nix:176:13:

          175|     then ""
          176|     else if pathExists cargoLock
             |             ^
          177|     then "cp ${cargoLock} $out/Cargo.lock"

       … while evaluating the attribute 'buildCommand' of the derivation 'dummy-src'
```

yusdacra:
Oh, that makes sense. I made a PR to crane to remove that check https://github.com/ipetkov/crane/pull/40.

PR #168, opened 2022-06-05
refactor: use new flake output schema for default package and devshells

yusdacra:
Changes builder outputs so that they fit the new flake outputs schema (nix 2.7+). Current nixos stable nix version is 2.8, so this should be an OK change.

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -279,6 +279,7 @@
     # Makes the packages tree compatible with flakes schema.
     # For each package the attr `{pname}` will link to the latest release.
     # Other package versions will be inside: `{pname}.versions`
+    # Adds a `default` package by using `defaultPackageName` and `defaultPackageVersion`.
     formattedOutputs =
       outputs
       // {
@@ -297,8 +298,13 @@
                   versions = releases;
                 })))
             allPackages;
+
+          defaultPackage =
+            allPackages
+            ."${dreamLockInterface.defaultPackageName}"
+            ."${dreamLockInterface.defaultPackageVersion}";
         in
-          latestPackages;
+          latestPackages // {default = defaultPackage;};
       };
   in
     formattedOutputs;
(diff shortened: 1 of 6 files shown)


DavHau:
Should we keep it downward compatible for a while by still having `defaultPackage` as an alias to `packages.default`, etc..?

yusdacra:
> Should we keep it downward compatible for a while by still having `defaultPackage` as an alias to `packages.default`, etc..?

I don't think we should. `nix flake check` warns about those outputs, so I'd rather not have that clutter. And since flakes are an experimental feature, it should be fine for us to break compatibility here, I think pretty much everyone that use flakes are going to use newer versions anyways...

yusdacra:
> I don't think we should. `nix flake check` warns about those outputs, so I'd rather not have that clutter. And since flakes are an experimental feature, it should be fine for us to break compatibility here, I think pretty much everyone that use flakes are going to use newer versions anyways (nix 2.9 also just released, so that makes 2.7 not too recent)...

Maybe we should start keeping a CHANGELOG.md file even if we don't do releases? Feels like it could be helpful.

DavHau:
> Maybe we should start keeping a CHANGELOG.md file even if we don't do releases? Feels like it could be helpful.

Not sure by which rules you would add an entry to the changelog. On every breaking change to the API?
I'm not sure if this will quickly spiral out into overhead that is equivalent like when doing real releases. For me writing the changelog always was the most time consuming part of `releasing` with mach-nix. ;) But maybe I was doing something wrong.

We could do releases without semantic versioning, to at least save us the time worrying about backward incompatibility. We just start from 0000 and count up, so it's followed by 0001 and 0002 and we use the github feature which generates a changelog from PRs (see example of the generated `What's Changed` [here](https://github.com/DavHau/mach-nix/releases/tag/3.5.0)). But then again I don't always do PRs when changing stuff as of now.

yusdacra:
Maybe something like https://github.com/yusdacra/nix-cargo-integration/blob/master/CHANGELOG.md where we only keep breaking changes? We don't really need a version number or anything, I think just using dates (calendar versioning) is fine.

PR #173, opened 2022-06-11
docs: setup mdbook book for documentation

yusdacra:
Adds an `mdbook` book to `docs/` and moves existing docs under the book. Adds subsystem documentation for `rust` subsystem. CI job is added for building the docs, and deploying to github pages (we need to let it run once on main branch and choose `gh-pages` branch in pages settings).

Also adds a CLI app for showing these docs in the terminal. It can be used with `nix run .#docs`. It also shows suggestions of docs if a doc with the name passed doesn't exist, but it is really just a list of files that have the passed name in their content.


docs/src/subsystems/rust.md
```diff
@@ -0,0 +1,32 @@
...
+### crane (ifd)
+
+Builds a package using [`crane`](https://github.com/ipetkov/crane).
+
+#### Override gotchas
+
+This builder builds two separate derivations, one for your crate's dependencies
+and another for your crate. This means that if you want to override stuff for
+the dependencies, you need to use the `<crate-name>-deps` key for your override
+where `<crate-name>` is the name of the crate you are building.
```
yusdacra, line 32: An example would be useful here

docs/src/intro.md
```diff
@@ -0,0 +1,28 @@
...
+- Simplified updating of packages
+
+The goal of this project is to create a standardized, generic, modular
+framework for automated packaging solutions, aiming for better flexibility,
+maintainability and usability.
+
+The intention is to integrate many existing 2nix converters into this framework,
+thereby improving many of the previously named aspects and providing a unified
+UX for all 2nix solutions.
+
```
yusdacra, line 27: It would probably be useful to add the "further reading" section here, with links to the various documentation

docs/src/subsystems/rust.md
```diff
@@ -0,0 +1,32 @@
...
+
+### cargo-toml (impure)
+
+Translates a `Cargo.toml` file to a dream2nix lockfile by generating a
+`Cargo.lock` file first and then using `cargo-lock` translator on the
+generated lockfile.
+
+## Builders
+
+### build-cargo-package (pure)
```
Kranzes, line 19: it's `build-rust-package`, not `build-cargo-package`.
```suggestion
### build-rust-package (pure)
```

README.md
```diff
@@ -4,7 +4,9 @@
 
 ## [WIP] dream2nix - A framework for automated nix packaging
 
-dream2nix is a framework for automatically converting packages from other build systems to nix.
+dream2nix is a framework for automatically converting packages from other build
+systems to nix.
```
DavHau, line 8: What is the reason for putting in new lines here and in many other places? I think the markdown renderer should decide where to wrap the lines. Otherwise we end up with weird line wrappings on smaller screens.

flake.nix
```diff
@@ -235,6 +236,50 @@
               export PATH="${alejandra.defaultPackage.${system}}/bin"
               ${pkgs.treefmt}/bin/treefmt --clear-cache "$@"
             '');
+
+          docs.type = "app";
+          docs.program =
+            l.toString
+            (pkgs.writeScript "docs" ''
+              # convert to lowercase, all of our doc paths are in lowercase
```
DavHau, line 244: Can we move that script into another file? And maybe we should expose this via devshell menu

README.md
yusdacra: Doesn't it already decide that (on GitHub / mdbook at least)? The reason I did that was so that it doesn't go over 80ish columns so it's easier to view and edit the raw markdown (especially if you don't have soft wrapping or don't like soft wrapping in your editor...). But if it is already a problem we can change it back (on some other docs as well).

flake.nix
yusdacra: Yeah that sounds alright, I'll do that :+1:

README.md
Kranzes: Sounds like something *you* have to set in your editor.

DavHau: Oh actually the .md renderer ignores single new lines, so it wouldn't lead to rendering issues. But anyways, I think it is better to not manually wrap in text files containing large text blocks. Otherwise you'd always have to adjust the line breaks after editing content. Let's change that back.

yusdacra: Okay, I'll do that later (if you don't do it first). Looks like maybe I should look into soft wrapping for Helix xD

PR #175, opened 2022-06-14
feat: implement overrideRustToolchain for crane and build-rust-package

yusdacra:
This implements an override for the Rust builders called `overrideRustToolchain` that can be used to override the Rust toolchain packages used by the builders. This also enables regex overrides for packages so that overriding the toolchain globally is made much more convenient.

### DIFF
--- a/docs/src/subsystems/rust.md
+++ b/docs/src/subsystems/rust.md
@@ -106,3 +106,23 @@ takes `cargo` packages like so:
 
 where `cargoHostTarget` has the same meaning as coming from a `pkgsHostTarget`
 and `cargoBuildBuild` has the same meaning as coming from a `pkgsBuildBuild`.
+
+Also keep in mind that if you want to override the toolchain for a package
+and you are using the `crane` builder, you would need to set an override
+for both the dependencies and the main package derivation:
+
+```nix
+let
+  toolchainOverride = old: { /* ... */ };
+in
+{
+  # ...
+  packageOverrides = {
+    # ...
+    crate-name.set-toolchain.overrideRustToolchain = toolchainOverride;
+    crate-name-deps.set-toolchain.overrideRustToolchain = toolchainOverride;
+    # ...
+  };
+  # ...
+}
+```
--- a/docs/src/subsystems/rust.md
+++ b/docs/src/subsystems/rust.md
@@ -16,7 +16,7 @@ generated lockfile.
 
 ## Builders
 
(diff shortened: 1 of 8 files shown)


Kranzes (approved):

yusdacra:
@DavHau do you think using regex overrides like this is alright? I was going to implement them but it looks like they were already there so I just uncommented the code.

DavHau:
> @DavHau do you think using regex overrides like this is alright? I was going to implement them but it looks like they were already there so I just uncommented the code.

I would prefer to not do that now, because the override interface is subject to changes and any features we add now will make it more difficult to transition.
I'd really like to try a more module system approach here. @infinisil Already helped me out building a [POC](https://github.com/infinisil/dream2nix-modules)

DavHau:
In general, I'd argue, that the toolchain settings should be a project level setting, not a per dependency setting. But in order to enable that configuration, we'd need a better interface, which brings us back to the module system :D

Anyways, we can add the regex overrides if you like, because eventually we'd have to break backward compatibility anyways if we used something nixos module based.

yusdacra:
> In general, I'd argue, that the toolchain settings should be a project level setting, not a per dependency setting. But in order to enable that configuration, we'd need a better interface, which brings us back to the module system :D

Yeah that's mostly true, but if we want to have centralized overrides then I think it's still nice to be able to override per dependency. It may also be useful for testing out patches that require a specific toolchain (although that's going to be a really rare of a use case so I imagine it may not be very useful).

I wanted to add a project level setting but indeed with how dream2nix `settings` work right now it's hard to do that, which means we'd need to expose another settings attribute (`builderArgs`) which would just make things worse. So I am for modules in this case, but I also think we should keep the ability to override per package.

But I think if we get modules, this code might be unnecessary since we could most likely implement an override function for packages that take module options like the user interface would do and just let the user set it there, this would reuse the module options for a builder.

> Anyways, we can add the regex overrides if you like, because eventually we'd have to break backward compatibility anyways if we used something nixos module based.

We can probably merge this then after I fix something with the crane builder override.

DavHau:
Sure, we should keep the ability to override per dependency attributes, no matter if the module system is used or not. It wasn't my intention to remove that.

yusdacra:
Seems there is something wrong with the crane builder when setting the toolchain (specifically in the main package derivation in installPhase), so the CI fails... I'll look into that further later.

DavHau (approved):

PR #176, opened 2022-06-14
Using `cabal2json` from nixpkgs instead that from a flake

aciceri:
@DavHau as promised


src/subsystems/haskell/translators/stack-lock/utils.nix
```diff
@@ -16,16 +16,29 @@ in rec {
...
-
-  cabal2json = cabal2jsonFlake.defaultNix.packages.${pkgs.system}.cabal2json;
+  cabal2json = let
+    haskellPackages = pkgs.haskell.packages.ghc8107.override {
+      overrides = curr: prev: {
+        autodocodec = prev.autodocodec.overrideAttrs (old: {
+          meta = old.meta // {broken = false;};
+        });
+        validity-aeson = prev.validity-aeson.overrideAttrs (old: {
+          meta = old.meta // {broken = false;};
```
m1-s, line 26: There is a function called `markUnbroken` https://github.com/NixOS/nixpkgs/blob/93d3ee74a184a568b5a506ef76e6ffa00866440f/pkgs/development/haskell-modules/lib/compose.nix#L178

Never used it with flakes but it should work the same, right?

src/subsystems/haskell/translators/stack-lock/utils.nix
```diff
@@ -16,16 +16,29 @@ in rec {
...
+          patches = [];
+        });
+      };
+    };
+    cabal2json' = haskellPackages.cabal2json;
+    cabal2json'' = cabal2json'.override {
+      Cabal = haskellPackages.Cabal_3_2_1_0;
+    };
+    cabal2json = cabal2json''.overrideAttrs (old: {
+      doCheck = false;
```
m1-s, line 38: There also is a `dontCheck` https://github.com/NixOS/nixpkgs/blob/93d3ee74a184a568b5a506ef76e6ffa00866440f/pkgs/development/haskell-modules/lib/compose.nix#L110

src/subsystems/haskell/translators/stack-lock/utils.nix
aciceri: Hi Michael, didn't know about these ones, nice!

aciceri: I see no reason why they shouldn't work (and they seems to work). Do you know if it's normal that one test failed?

DavHau:
c2a8067aada2fdae22f80e8ef2ea0cdbe5ca7331 should fix the failing build. Just rebase onto the upstream haskell branch

aciceri:
It doesn't seem to have worked.

DavHau:
I think for now it's just the format not matching. You can format your code by running `nix run .#format`.
An other thing is, that the example tests are running super slow now. This is not related to your change. I will try to fix that.

DavHau:
> If you prefer I can squash the commits however.

Not necessary. I will use the `squash and merge` button ;)

PR #186, opened 2022-07-03
add devShell outputs for nodejs ad python

DavHau:
This adds `devShell` and `devShells` outputs to the python and nodejs default builders
That means, flake outputs generated by `makeFlakeOutputs` will be suitable for use with `nix develop` or `direnv`.

### DIFF
--- a/src/subsystems/nodejs/builders/granular/default.nix
+++ b/src/subsystems/nodejs/builders/granular/default.nix
@@ -62,7 +62,7 @@
           makePackage name version))
       packageVersions;
 
-    outputs = {
+    outputs = rec {
       # select only the packages listed in dreamLock as main packages
       packages =
         b.foldl'
@@ -73,6 +73,16 @@
             "${name}"."${version}" = allPackages."${name}"."${version}";
           })
           args.packages);
+
+      devShell = devShells.default;
+
+      devShells =
+        {default = devShells.${defaultPackageName};}
+        // (
+          l.mapAttrs
+          (name: version: allPackages.${name}.${version}.devShell)
+          args.packages
+        );
     };
 
     # This is only executed for electron based packages.
@@ -201,6 +211,17 @@
         pname = lib.replaceStrings ["@" "/"] ["__at__" "__slash__"] name;
 
         passthru.dependencies = passthruDeps;
+
+        passthru.devShell = pkgs.mkShell {
+          NODE_PATH = l.concatStringsSep ":" (
+            l.map
+            (dep: "${allPackages.${dep.name}.${dep.version}}/lib/node_modules")
+            deps
+          );
+          buildInputs = [
+            nodejs
+          ];
+        };
 
         installMethod = "symlink";
 
--- a/src/subsystems/nodejs/builders/granular/default.nix
+++ b/src/subsystems/nodejs/builders/granular/default.nix
@@ -250,8 +250,8 @@
           shellHook = ''
             # create the ./node_modules directory
             if [ -e ./node_modules ] && [ ! -L ./node_modules ]; then
-              echo -e "\nFailed creating the ./node_modules symlink."
-              echo -e "\n./node_modules already exists and is a directory, which means it is managed by anaother program. Please delete ./node_modules first and re-enter the dev shell."
+              echo -e "\nFailed creating the ./node_modules symlink to ${nodeModulesDir}"
+              echo -e "\n./node_modules already exists and is a directory, which means it is managed by another program. Please delete ./node_modules first and re-enter the dev shell."
             else
               rm -f ./node_modules
               ln -s ${nodeModulesDir} ./node_modules
--- a/src/subsystems/nodejs/builders/granular/default.nix
+++ b/src/subsystems/nodejs/builders/granular/default.nix
@@ -158,12 +158,33 @@
 
     # Generates a derivation for a specific package name + version
     makePackage = name: version: let
+      pname = lib.replaceStrings ["@" "/"] ["__at__" "__slash__"] name;
+
       deps = getDependencies name version;
 
       nodeDeps =
         lib.forEach
         deps
         (dep: allPackages."${dep.name}"."${dep.version}");
+
+      nodeModulesDir = pkgs.runCommand "node_modules-${pname}" {} ''
+        mkdir $out
+        ${l.concatStringsSep "\n" (
+          l.forEach nodeDeps
+          (pkg: ''
+            ln -s ${pkg}/lib/node_modules/* $out/
+          '')
+        )}
+      '';
+
+      nodeModulesBinDirs = pkgs.runCommand "node_modules_bin-${pname}" {} ''
+        touch $out
+        for dep in ${l.toString nodeDeps}; do
+          for bin in $(${pkgs.findutils}/bin/find $dep -name .bin); do
+            echo "$bin" >> $out
+          done
+        done
+      '';
 
       passthruDeps =
         l.listToAttrs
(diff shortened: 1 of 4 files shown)


wmertens:
So cool that I only needed to change the input url to point to this branch to test :-)

The devshell part is working, but:

- `NODE_PATH` is very long and I worry it might get too long in big projects - I think it'd be better to point to a composed directory
- the `PATH` doesn't include the node_modules binaries, so running e.g. `eslint` doesn't work, and since there's no composed node_modules I can't point a symlink to that, so tools like `nps` don't find the binary to call
- (sidenote) am I correct in that the default app actually also includes the devDependencies?

I tried building the package with `nix build .` but that complained about not having `defaultPackage` defined so instead I ran `nix build .#default` which did make the package. I then tried to symlink the modules with `ln -s result/lib/node_modules/MY_PKG/node_modules` and then I noticed that there's no `.bin` directory in there either so `./node_modules/.bin/eslint` still didn't work.

wmertens:
Ah - another issue: [ESM apps don't support NODE_PATH](https://nodejs.org/api/esm.html#esm_no_node_path).

So setting NODE_PATH in devShell is not a 100% solution, instead the only way is to symlink node_modules.

wmertens:
another issue: When I try to specify a nodejs version, the devshell attributes disappear.

I can't find how to specify the version, the below is just a guess

flake.nix:
```nix
{
  inputs.dream2nix.url = "github:nix-community/dream2nix/devShells";
  outputs = inp:
      systems = [ "x86_64-linux" ];
      config.projectRoot = ./.;
      source = ./.;
      settings = [{
        subsystemInfo.nodejsVersion = 18;
      }];
    };
}
```

wmertens:
Also, devShell should ignore the `systems` attribute and only build for the current system, right? I don't see that in the PR but I don't know how that works with mkShell.

DavHau:
@wmertens 
Could you check if my last commit 9901487fc5c616799ad2de7451af1823fc97065f resolves the issues? I stopped exporting NODE_PATH and instead create a node_modules directory now.

> I tried building the package with `nix build .` but that complained about not having `defaultPackage` defined so instead I ran `nix build .#default` which did make the package.

`defaultPackage` is a deprected flake attribute. If you use a newer nix version, it will automatically default to `packages.${system}.default`.

> another issue: When I try to specify a nodejs version, the devshell attributes disappear.

That is because the dream-lock.json file which you earlier created by executing `nix run .#resolveImpure` or `nix run .#default.resolve` is now invalid, because of the nodejsVersions change. Just re run the resolve command and the devShell attribute should re-appear.

src/subsystems/nodejs/builders/granular/default.nix
```diff
@@ -148,13 +158,34 @@
...
         lib.forEach
         deps
         (dep: allPackages."${dep.name}"."${dep.version}");
 
+      nodeModulesDir = pkgs.runCommand "node_modules-${pname}" {} ''
+        mkdir $out
+        ${l.concatStringsSep "\n" (
+          l.forEach nodeDeps
+          (pkg: ''
+            ln -s ${pkg}/lib/node_modules/* $out/
```
wmertens, line 175: this fails when linking multiple namespaced modules like @pnp

src/subsystems/nodejs/builders/granular/default.nix
```diff
@@ -148,13 +158,34 @@
...
+          '')
+        )}
+      '';
+
+      nodeModulesBinDirs = pkgs.runCommand "node_modules_bin-${pname}" {} ''
+        touch $out
+        for dep in ${l.toString nodeDeps}; do
+          for bin in $(${pkgs.findutils}/bin/find $dep -name .bin); do
```
wmertens, line 183: Isn't there only one .bin folder at the top of the node_modules? so compose .bin in the node_modules directory with something like
```sh
if [ -d ${dep}/lib/node_modules/.bin ]; then
  ln -s ${dep}/lib/node_modules/.bin/* $out/.bin
fi
```
?

src/subsystems/nodejs/builders/granular/default.nix
wmertens: yes but if there's two, the second one tries to overwrite the first

DavHau:
@wmertens Thanks for your detailed review. I think the last commit should solve all your points.
Could you test it once again? Thanks.

wmertens:
@DavHau Nice, node_modules works :-)

Now I'm struggling to get the correct Node.js version. I am trying
```nix
      settings = [{
        nodejs = 18;
        subsystemInfo.nodejsVersion = 18;
      }]
```
but they don't work, the resolution is version 14.

Then, the devShell doesn't have nodejs in the path, but that's just a matter of

```nix
        passthru.devShell = pkgs.mkShell {
          path = [
            nodejs
          ];
          buildInputs = [
            nodejs
          ];
```

in the builder. I wonder if electron should also be in there for electron projects?

wmertens:
BTW, would it not be best to then make the package with the composed node_modules, now that we have it split out and working including .bin?

wmertens:
Ok, so when I use https://github.com/wmertens/dream2nix/commit/f30dce390b9fa30ca744703082701ff3b4c46503 I can get node in my devShell. I'm using this flake:

```nix
{
  inputs = {
    dream2nix.url = "github:nix-community/dream2nix/devShells";
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-22.05";
  };

  outputs = inp:
      systems = [ "x86_64-linux" "aarch64-darwin" "x86_64-darwin" ];
      config.projectRoot = ./.;
      source = ./.;
      settings = [{
        subsystemInfo.nodejs = 18;

        # This also removes dev deps from the devShell :-(
        # subsystemInfo.noDev = true;
      }];
    };
}
```


I now found another issue: I'd like to keep the dev dependencies out of the production bundle. I tried setting noDev but then the dream-lock doesn't include the dev dependencies at all.

Perhaps the dream-lock should put dev dependencies under the devDependencies key, and then the package doesn't use them but the devShell does?

DavHau:
@wmertens f30dce390b9fa30ca744703082701ff3b4c46503 does not seem to make any difference for me. Packages which are inside `buildInputs` are automatically added to the path.
With the current PR, I observe this behavior:

```command
$ nix develop
$ ls $(which node)
/nix/store/liq2m5v995wh5kzbvw9il4p4iwniwsjq-nodejs-14.19.3/bin/node
ls /nix/store/liq2m5v995wh5kzbvw9il4p4iwniwsjq-nodejs-14.19.3/bin
corepack  node  npm  npx
```

> BTW, would it not be best to then make the package with the composed node_modules, now that we have it split out and working including .bin?

We could do that but its still a trade-off. It would add +1 derivation for each dependency. Launching many small builds does introduce more overhead. But in terms of caching it might be better, for sure. I currently need to focus on a few other things. If you want to work on a PR, feel free.

I'll try to have a look at the nodejs version problem though and the peerDependency handling mentioned in another thread.

> I now found another issue: I'd like to keep the dev dependencies out of the production bundle. I tried setting noDev but then the dream-lock doesn't include the dev dependencies at all.
> 
> Perhaps the dream-lock should put dev dependencies under the devDependencies key, and then the package doesn't use them but the devShell does?

Currently the dream-lock does not have a concept of different kind of dependencies. The schema should probably be extended at some point. For now we could store that information under the `_subsystem` key which is a free form that can be utilized by each subsystem differently.

DavHau:
> I now found another issue: I'd like to keep the dev dependencies out of the production bundle. I tried setting noDev but then the dream-lock doesn't include the dev dependencies at all.
> 
> Perhaps the dream-lock should put dev dependencies under the devDependencies key, and then the package doesn't use them but the devShell does?

Actually devDendencies are required to build the production bundle. Therefore we cannot remove them as build inputs. But we can purge them from the final closure after the build, maybe inside `fixupPhase` ?

wmertens:
Ooh I missed all this, I was having too much fun changing the translation and builder layers :sweat_smile:

I'll have a PR soon, still fixing some build issues. Might make a broken draft PR for comments.

DavHau:
> Might make a broken draft PR for comments.

Sounds good

wmertens:
BTW, I needed that `path = [nodejs];` for devShell to work in direnv. Since direnv doesn't start a subshell, I presume it doesn't have the same environment and it probably uses the path only.

DavHau:
If I use direnv with `use flake` in .envrc it works completely fine and it adds nodejs/bin to my PATH. I'd really like to understand what is different on your side.


direnv: loading ~/synced/projects/github/dream2nix/examples/eslint/.envrc
direnv: using flake
direnv: nix-direnv: using cached dev shell
direnv: export +AR +AS +CC +CONFIG_SHELL +CXX +HOST_PATH +IN_NIX_SHELL +LD +NIX_BINTOOLS +NIX_BINTOOLS_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu +NIX_BUILD_CORES +NIX_CC +NIX_CC_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu +NIX_CFLAGS_COMPILE +NIX_ENFORCE_NO_NATIVE +NIX_HARDENING_ENABLE +NIX_INDENT_MAKE +NIX_LDFLAGS +NIX_STORE +NM +NODE_PATH +OBJCOPY +OBJDUMP +RANLIB +READELF +SIZE +SOURCE_DATE_EPOCH +STRINGS +STRIP +buildInputs +buildPhase +builder +configureFlags +depsBuildBuild +depsBuildBuildPropagated +depsBuildTarget +depsBuildTargetPropagated +depsHostHost +depsHostHostPropagated +depsTargetTarget +depsTargetTargetPropagated +doCheck +doInstallCheck +dontAddDisableDepTrack +name +nativeBuildInputs +out +outputs +patches +phases +propagatedBuildInputs +propagatedNativeBuildInputs +shell +shellHook +stdenv +strictDeps +system ~PATH ~XDG_DATA_DIRS

PR #187, opened 2022-07-03
imp: only uses the typescript patch if typescript is used

blaggacao:
- this is a dev-loop optimization
- it means that when working on the d2n, not every iteration on the
  ts patch triggers a gigantic rebuilding loop as no inputs change
  on non-ts packages


DavHau:
Thanks a lot for this!

PR #191, opened 2022-07-08
nodejs: upgrade esbuild 0.14 override

wmertens:
I wonder if there's no way to automatically generate this in the impure phase? The module now has a check for the binary version and doesn't like it when there's a mismatch.

### DIFF
--- a/overrides/nodejs/default.nix
+++ b/overrides/nodejs/default.nix
@@ -409,40 +409,7 @@ in
           '';
         };
       };
-      "add-binary-0.14" = let
-        version = "0.14.38";
-      in {
-        _condition = satisfiesSemver "~0.14";
-        ESBUILD_BINARY_PATH = let
-          esbuild = pkgs.buildGoModule {
-            pname = "esbuild";
-            inherit version;
-
-            src = pkgs.fetchFromGitHub {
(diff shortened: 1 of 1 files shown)


wmertens:
actually, 0.14 now uses optionalDependencies and so an override is no longer needed

wmertens:
Yes, 0.14 works fine without the override, I have a commit in my other branch that only removes that.

wmertens:
FYI now this PR just removes the override for 0.14

DavHau:
> FYI now this PR just removes the override for 0.14

esbuild is a GO program, I really wonder how it can work without a binary. Maybe it just works in your case, because you depend on esbuild but don't make use of it during build?

wmertens:
It's very clever: It has optional dependencies on static binaries directly packaged as NPM packages, and then the `os` filter only installs the correct platform.

For example, https://unpkg.com/browse/esbuild-linux-64@0.14.53/

DavHau:
Wow unpkg.com is an incredible tool .Thanks for that. OK, it seems like esbuild has this mechanism since 0.13.0.
Though currently most of our translator seem to ignore optionalDependencies. At least on the top-level. That should probably be fixed.

PR #197, opened 2022-07-14
indexer: add libraries-io

DavHau:
Indexer which utilizes the libraries.io api.
Given a platform (like npm or crates-io), it returns a list of packages sorted
by the highest number of dependents
(the number of other packages which depend on that package).

This indexer requires an api key that must first be created on libraries.io.
Github login is supported, so this is very straight forward.

More options could be added.
libraries.io also supports other interesting popularity metrics:
  - dependent_repos_count
  - stars
  - forks
  - rank (source rank, see https://libraries.io/api#project-sourcerank)


yusdacra (approved):
Looks fine to me

PR #201, opened 2022-07-16
python improvements, docs and guide.

DavHau:
- add warning to readme and docs stating that dream2nix is unstable software
- refactor structure of docs: remove title .md pages, instead add titles via (# title) to SUMMARY
- change order of doc topics. move `guides` and `subsystems` to the front
- add docs for python subsystem
- add getting started guide for python
- extend ./examples/python to make use of translator options
- add new function to dlib `systemsFromFile` that allows to load the `supportedSystems` from a file. This alows changing the supported systems without touching the flake.
- move pip translator args to extra file `args.nix` to allow embedding in docs
- fix pip translator
- add a .gitignore to the basic flake template
- use new systemsFromFile function in basic flake template


src/lib/default.nix
```diff
@@ -295,6 +296,17 @@
...
     else sanitizedRelPath;
 
+  systemsFromFile = file:
+    if ! l.pathExists file
+    then
+      throw ''
+        The system for your flake.nix is not initialized yet.
+        Please execute the following command to initialize it:
+
+        nix eval --impure --raw --expr 'builtins.currentSystem' > ./${l.baseNameOf file} && git add ./${l.baseNameOf file}
```
yusdacra, line 306: This won't work if the user passes a path relative to the `flake.nix` like `./nix/supported-systems`. Maybe we should instead use `config.projectRoot` to strip the prefix path?

templates/simple/flake.nix
```diff
@@ -3,7 +3,7 @@
   outputs = inp:
     inp.dream2nix.lib.makeFlakeOutputs {
       # modify according to your supported systems
```
yusdacra, line 5: Maybe we should remove / change the comment here? "append / remove lines to the file for modifying supported systems" or something?

yusdacra (changes requested):

templates/simple/flake.nix
DavHau: I removed the comment. I think a comment is not necessary because an error will be thrown on first use explaining what to do.

yusdacra (approved):

PR #202, opened 2022-07-18
refactor: readme/docs

DavHau:
The motivation of this change is to trim down the github front page (README.md). Code examples should not be part of that page. It should be short and only contain the most important information that's necessary for a rough overview.

- add sections `Quick Start` and `Presentation ...` to docs
- move usage examples from readme to `Quick Start` in the docs
- trim down the explanation about `Documentation` a bit
- move the youtube presentation to the end


DavHau:
I just noticed that the code snippets inside the docs do not have syntax highlighting which degrades the quality for the quick start examples. Reverting this for now.

PR #209, opened 2022-07-24
fix: make dream2nix evaluation work with --read-only flag

yusdacra:
See https://github.com/NixOS/nixos-search/pull/502#issuecomment-1193305553 and the discussion there.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -139,7 +139,7 @@
       (lib.attrNames externalPaths)
       (inputName: inp."${inputName}");
 
-    overridesDirs = ["${./overrides}"];
+    overridesDirs = [(toString ./overrides)];
 
     # system specific dream2nix api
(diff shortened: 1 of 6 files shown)


yusdacra:
I checked if in particular the removal of using `builtins.path` for `files` when creating a source tree would cause problems in derivations; for Rust crane builder it doesn't seem to change anything, the derivations still seem to be same, no dependencies to the whole source tree is added. Not sure about other subsystems, but it seems to be fine.

DavHau:
Could you add a test that ensures that whatever this fixed will not be broken by a future change?

yusdacra:
OK, I added tests. I had to make it so that the read-only check isn't ran on the haskell subsystem examples because those use IFD so they won't work with `--read-only` anyways.

PR #210, opened 2022-07-24
(WIP): init: impure debian translator

a-kenji:
Initialize an impure translator from debian binaries.


examples/debian-binary/flake.lock
```diff
@@ -0,0 +1,228 @@
+{
```
yusdacra, line 1: Can you delete the `flake-lock` file? We don't keep the flake lock for examples.

a-kenji:
This is now superseded by #212.

PR #217, opened 2022-07-30
Some fixes

wmertens:
- use runCommandLocal to speed up builds and keep caches cleaner
- remove unused node2nix

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -90,10 +90,6 @@ in let
   subsystems = callPackageDream ./subsystems {};
 
   externals = {
-    node2nix = nodejs:
-      pkgs.callPackage "${externalSources.node2nix}/nix/node-env.nix" {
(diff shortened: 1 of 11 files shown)


src/default.nix
```diff
@@ -317,9 +316,7 @@ in let
             allPackages;
 
           defaultPackage =
-            allPackages
-            ."${dreamLockInterface.defaultPackageName}"
-            ."${dreamLockInterface.defaultPackageVersion}";
+            allPackages."${dreamLockInterface.defaultPackageName}"."${dreamLockInterface.defaultPackageVersion}";
```
DavHau, line 319: This change violates column 80. Please revert.
Upstream issue: https://github.com/kamadorueda/alejandra/issues/331

wmertens: done

PR #223, opened 2022-08-01
Fix some rust issues

### LINKED ISSUE #219: rust: failing to vendor dependency of sub-directory
The rust project [nitrokey-3-firmware](https://github.com/Nitrokey/nitrokey-3-firmware) seemed to be troublesome to package with current nixpkgs tooling during summer of nix due to an issue with parsing and fetching the git revision correctly.

I just tried with dream2nix and it fails with an other error. Vendoring the dependency of a sub directory seems to fail.
flake.nix:
```nix
{
  inputs.src.url = "github:Nitrokey/nitrokey-3-firmware";
  inputs.src.flake = false;
  outputs = inp:
      systemsFromFile = ./nix_systems;
      config.projectRoot = ./.;
      source = inp.src;
    };
}
```
error:
```
runner> error: failed to get `fm11nc08` as a dependency of package `runner v1.0.4 (/build/lpc55)`
```

@sbruder created a minimal reproducibel example of the problem here: https://github.com/sbruder/nix-cargo-two-unpinned-dependencies/blob/master/Cargo.lock
This as well fails with dream2nix, but this time with another error than the flake above:
```
nix-cargo-two-unpinned-dependencies>     Updating git repository `https://github.com/rust-lang/libc`
nix-cargo-two-unpinned-dependencies> error: failed to get `libc` as a dependency of package `subproject v0.1.0 (/build/2340gklvaqqbmibp3z875282cmi8bj55-source/subproject)`
nix-cargo-two-unpinned-dependencies>     ... which satisfies path dependency `subproject` (locked to 0.1.0) of package `nix-cargo-two-unpinned-dependencies v0.1.0 (/build/2340gklvaqqbmibp3z875282cmi8bj55-source)`
```

@yusdacra

yusdacra:
Hopefully this will close #219 soon

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -251,6 +251,7 @@ in let
     builder,
     builderArgs,
     fetchedSources,
+    sourceRoot,
     dreamLock,
     inject,
     sourceOverrides,
@@ -275,6 +276,7 @@ in let
         inherit
           produceDerivation
           dreamLock
+          sourceRoot
           ;
 
         inherit
@@ -370,6 +372,7 @@ in let
         fetchedSources
         allOutputs
         sourceOverrides
+        sourceRoot
         ;
 
       builder = builder';
--- a/src/subsystems/rust/builders/build-rust-package/default.nix
+++ b/src/subsystems/rust/builders/build-rust-package/default.nix
@@ -32,9 +32,9 @@
     buildPackage = pname: version: let
       src = utils.getRootSource pname version;
       vendorDir = vendoring.vendoredDependencies;
-      replacePaths = utils.replaceRelativePathsWithAbsolute {
-        paths = subsystemAttrs.relPathReplacements.${pname}.${version};
-      };
+      replacePaths =
+        utils.replaceRelativePathsWithAbsolute
+        subsystemAttrs.relPathReplacements.${pname}.${version};
       writeGitVendorEntries = vendoring.writeGitVendorEntries "vendored-sources";
 
       cargoBuildFlags = "--package ${pname}";
--- a/src/subsystems/rust/builders/crane/default.nix
+++ b/src/subsystems/rust/builders/crane/default.nix
@@ -44,9 +44,9 @@
     };
 
     buildPackage = pname: version: let
-      replacePaths = utils.replaceRelativePathsWithAbsolute {
-        paths = subsystemAttrs.relPathReplacements.${pname}.${version};
-      };
+      replacePaths =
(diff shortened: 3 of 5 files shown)


yusdacra:
So while I was trying to implement support for duplicated versions, I came across to https://github.com/rust-lang/cargo/issues/10310. The package in question has duplicated versions of a dependency in it's lockfile which `cargo vendor` does not support, and by extension the vendor directory hierarchy does not also support it. Either upstream needs to fix this (faster) or we wait until cargo supports this (probably going to take some time considering the issue is still open...).

DavHau:
Quick question before I review this. Is this PR just working around an upstream issue or does it actually fix bugs on our side? We should probably not increase complexity on our side if this is an upstream issue.

yusdacra:
> Quick question before I review this. Is this PR just working around an upstream issue or does it actually fix bugs on our side? We should probably not increase complexity on our side if this is an upstream issue.

This PR only fixes bugs on our side. I made an issue on upstream that should handle the issue I mentioned.

src/default.nix
```diff
@@ -251,6 +251,7 @@ in let
     builder,
     builderArgs,
     fetchedSources,
+    sourceRoot,
```
DavHau, line 254: I think this can be removed. It isn't used by the builders. Also I think that we should change the way we handle relative path sources. I think we should always pass the root source as `src` and then set `sourceRoot` of `mkDerivation`. The unpackPhase will then switch to the correct subdir. That would allow packages to always refer to files in parent directories without issues. Some nodejs builds currently crash because sub-projects want to access files of parent projects in order to build.

yusdacra: It is used by the Rust builders with this PR, to be able to get the full project repository so path dependencies work properly. Before it would only capture the path dependency directory, which will break if the path dependency depends on other path dependencies that are in a parent directory. We could remove this fine if we removed the `builtins.path` usage for path depedencies, but I think that would cause regressions?

DavHau: True. This is a trade-off situation that should be carefully thought through.
The isolation of sources gives us better incremental builds, but it also breaks some builds. We should probably provide the user with the choice to turn this on or off, and have the default on the safer side which is probably not to do source isolation. At least for nodejs that would be the case. Maybe the default can be different from subsystem to subsystem.

yusdacra: Yes, I agree. I think we should have different functions for getting an isolated path vs a non-isolated path. And then the builder can decide which one to use.

PR #224, opened 2022-08-02
Add example of overriding devShells to prettier flake

billewanick:
Pulled out `makeFlakeOutputs` and checks into a `let`.
Then override the devShells to add to `buildInputs`.

Thanks to @DavHau for helping me figuring this out.


DavHau:
merged and improved via: https://github.com/nix-community/dream2nix/pull/341

PR #225, opened 2022-08-03
fix: nodejs: put direct dependencies in PATH first

### LINKED ISSUE #221: Node: transitive dependency bin taking priority over direct dependency
## Intro

I have a direct binary dev dependency `svgo@2` in `package.json`.
Another direct dependency depends on an earlier version `svgo@1`.
Dream2Nix resolves the linked binary in `$out/node_modules/.bin` to the earlier version - the transitive dependency - instead of the direct dependency.

## Details

While packaging [Castopod](https://code.castopod.org/adaures/castopod) with dream2nix:

`package.json`:
```
"devDependencies": {
    "svgo": "^2.2.2",
    "cssnano": "^4.1.10"
},
```
With `cssnano -> svgo@1`.

tinybeachthor:
Fixes #221 

Put direct dependency bins (collected by `src/subsystems/nodejs/builders/granular/install-deps.py`) to `PATH` first.

### DIFF
--- a/src/subsystems/nodejs/builders/granular/default.nix
+++ b/src/subsystems/nodejs/builders/granular/default.nix
@@ -426,6 +426,7 @@
 
           # symlink sub dependencies as well as this imitates npm better
           python $installDeps
+          ADD_BIN_PATH=$(cat $TMP/ADD_BIN_PATH)
 
           echo "Symlinking transitive executables to $nodeModules/.bin"
           for dep in ${l.toString nodeDeps}; do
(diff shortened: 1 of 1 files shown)


DavHau:
Thanks for noticing this. But this doesn't seem to fix the root cause, which is that the symlinks in .bin are done incorrectly. I think the symlinking logic should be fixed instead so that transitive dependencies don't take precedence over the ones of direct dependencies.

tinybeachthor:
@DavHau Yeah I was hoping this would be the easy way to fix it, but it does not.

I will update the symlinking logic and put up a new PR.

PR #226, opened 2022-08-04
init: debian indexer

a-kenji:
Add an indexer for debian packages based on debians popularity contest.

options:
	- outputFile
	- maxPackages
	- exclusions
	- inclusions

Co-authored-by: Ctem <c@ctem.me>


DavHau:
What is the status of this? In case it is done, please rebase once more on main.

PR #174, opened 2022-06-13
Update all-cabal-hashes as a flake input

ners:

PR #228, opened 2022-08-05
translators/stack-lock: Use nix-community/all-cabal-json

leungbk:
Some things I'm not sure of:

1. Do we additionally want to use `all-cabal-json` for the top-level
   file? This would have the drawback of yielding outdated data for a
   developer of a Haskell project who wants to use `dream2nix` to
   manage their project. I'm not sure if `dream2nix` is really
   designed for that use though.
2. Do we want to continue using `all-cabal-hashes` in the `stack-lock`
   translator? We end up copying a Cabal file from `all-cabal-hashes`
   in the builder anyway. Though I'm not sure why we don't simply use
   the Cabal file from the source tree.

### DIFF
--- a/src/subsystems/haskell/translators/stack-lock/default.nix
+++ b/src/subsystems/haskell/translators/stack-lock/default.nix
@@ -77,7 +77,7 @@ in {
     utils,
     ...
   }: let
-    haskellUtils = import ./utils.nix {inherit dlib lib pkgs;};
+    stackLockUtils = import ./utils.nix {inherit dlib lib pkgs;};
     all-cabal-hashes = let
       all-cabal-hashes' = pkgs.runCommandLocal "all-cabal-hashes" {} ''
         mkdir $out
@@ -114,14 +114,14 @@ in {
         (l.attrNames projectTree.files);
 
       cabalFile = projectTree.getNodeFromPath (l.head cabalFiles);
-      cabal = haskellUtils.fromCabal cabalFile.fullPath project.name;
+      cabal = stackLockUtils.fromCabal cabalFile.fullPath project.name;
       defaultPackageVersion =
         l.concatStringsSep
         "."
         (l.map l.toString cabal.description.package.version);
 
       stackLock =
-        haskellUtils.fromYaml
+        stackLockUtils.fromYaml
         (projectTree.getNodeFromPath "stack.yaml.lock").fullPath;
 
       snapshotEntry = l.head (stackLock.snapshots);
@@ -131,7 +131,7 @@ in {
         sha256 = snapshotEntry.completed.sha256;
       };
 
-      snapshot = haskellUtils.fromYaml snapshotYamlFile;
+      snapshot = stackLockUtils.fromYaml snapshotYamlFile;
 
       hidden =
         hiddenPackagesDefault;
@@ -153,8 +153,10 @@ in {
         (rawObj: dlib.nameVersionPair rawObj.name rawObj.version)
         serializedRawObjects;
 
+      haskellUtils = import ../utils.nix {inherit lib;};
+
       cabalData =
-        haskellUtils.batchCabalData
+        stackLockUtils.batchCabalData
         allCandidates;
 
       parseStackLockEntry = entry:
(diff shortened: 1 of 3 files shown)


src/subsystems/haskell/translators/stack-lock/default.nix
```diff
@@ -114,14 +114,14 @@ in {
         (l.attrNames projectTree.files);
 
       cabalFile = projectTree.getNodeFromPath (l.head cabalFiles);
-      cabal = haskellUtils.fromCabal cabalFile.fullPath project.name;
+      cabal = stackLockUtils.fromCabal cabalFile.fullPath project.name;
```
leungbk, line 117: See point 1 in https://github.com/nix-community/dream2nix/pull/228#issue-1330391847.

src/subsystems/haskell/translators/stack-lock/default.nix
```diff
@@ -77,7 +77,7 @@ in {
     utils,
     ...
   }: let
-    haskellUtils = import ./utils.nix {inherit dlib lib pkgs;};
+    stackLockUtils = import ./utils.nix {inherit dlib lib pkgs;};
     all-cabal-hashes = let
```
leungbk, line 81: See point 2 in https://github.com/nix-community/dream2nix/pull/228#issue-1330391847

src/subsystems/haskell/translators/stack-lock/utils.nix
```diff
@@ -36,39 +36,4 @@ in rec {
...
-      cabalFile=${all-cabal-hashes}/${name}/${version}/${name}.cabal
-      if [ -e $cabalFile ]; then
-        echo "converting cabal to json: ${name}-${version}"
-        mkdir -p $out/${name}/${version}
-        ${cabal2json}/bin/cabal2json \
-          $cabalFile \
-          > $out/${name}/${version}/cabal.json
-      else
-        echo "all-cabal-hashes" seems to be outdated
-        exit 1
```
leungbk, line 53: I forgot about including something like this in my modified versions of this function. Is there a clean way of doing this check without Bash?

leungbk: OK, I used `throw`.

DavHau:
> Some things I'm not sure of:
> 
>     1. Do we additionally want to use `all-cabal-json` for the top-level
>        file? This would have the drawback of yielding outdated data for a
>        developer of a Haskell project who wants to use `dream2nix` to
>        manage their project. I'm not sure if `dream2nix` is really
>        designed for that use though.
> 
>     2. Do we want to continue using `all-cabal-hashes` in the `stack-lock`
>        translator? We end up copying a Cabal file from `all-cabal-hashes`
>        in the builder anyway. Though I'm not sure why we don't simply use
>        the Cabal file from the source tree.

1. I think for the top-level package we should only use the actual cabal file from the source. I think using dream2nix on in-development projects should be supported and will be one of the major use-cases. Though for information about transitive deps we still need the cabal files from all-cabal-json (see answer 2.).

2. The cabal file from the source tree will only provide us with the dependency names of direct dependencies. In order to find transitive dependencies we need to consult the cabal files of our direct deps. Therefore we need either `all-cabal-hashes` or `all-cabal-json` in the translator as well.

I actually forgot about the fact that the builder uses the cabal files from all-cabal-hashes as well to replace the one coming from the package. Though it should only do that for dependencies. Not for our top-level package. If that is not the case, we should fix it.

Also I think, eventually we should only use one of the `all-cabal-...` repos. I think we should update `all-cabal-json` to also contain the cabal files. Then we can get rid of `all-cabal-hashes`.

src/subsystems/haskell/translators/utils.nix
```diff
@@ -0,0 +1,57 @@
+{
+  lib,
+  pkgs,
+}: let
+  l = lib // builtins;
+
+  all-cabal-json = let
```
DavHau, line 7: We should probably manage this as a flake input as seen in https://github.com/nix-community/dream2nix/pull/174/files

DavHau:
Merging for now as this is already a huge improvement. Feel free to do further improvements in another PR.

PR #242, opened 2022-08-10
Nodejs fixes

wmertens:


wmertens:
Note that the test failed due to memory issues while building prettier, which are unrelated to this PR

wmertens:
and the prettier build is failing because [it has a `build` script](https://github.com/prettier/prettier/blob/00ec91bfdee2e9bc592e4d1cf842507e50a86bbb/package.json#L165), which I think is non-standard but the builder executes it because it's the main package.

Would it not be better to require main packages to specify `d2nBuild` or use standard npm scripts?

src/subsystems/nodejs/builders/granular/tsconfig-to-json.js
```diff
@@ -1,19 +1,21 @@
...
+  console.error(err);
 }
 
-config = ts.parseConfigFileTextToJson(data)
-newdata = JSON.stringify(config)
-fs.writeFileSync('tsconfig.json', newdata);
+config = ts.parseConfigFileTextToJson(data);
+
+// https://www.typescriptlang.org/tsconfig#preserveSymlinks
+config.compilerOptions.preserveSymlinks = true;
```
DavHau, line 19: This logic is redundant with https://github.com/nix-community/dream2nix/blob/3223d29754c1ba1223c848e274cd4481c2bcd979/src/subsystems/nodejs/builders/granular/default.nix#L415

src/subsystems/nodejs/builders/granular/fix-package.py
```diff
@@ -14,7 +14,7 @@
 
 changed = False
 
-# fail if platform incompatible
+# fail if platform incompatible - should not happen due to filters
```
DavHau, line 17: Which filters do you mean?

DavHau:
> and the prettier build is failing because [it has a `build` script](https://github.com/prettier/prettier/blob/00ec91bfdee2e9bc592e4d1cf842507e50a86bbb/package.json#L165), which I think is non-standard but the builder executes it because it's the main package. Plus I suppose it needs to build more than in other tests because the node version is now 16.

We should either increase the `--max-old-space` value set [here](https://github.com/prettier/prettier/blob/7ced9e66549b8d5a8e3a74473af462f86fb0245c/package.json#L164) or simply update the version of prettier in the example (I guess).

> Would it not be better to require main packages to specify d2nBuild or use standard npm scripts?

What do you mean by standard npm scripts? `npm run build` is the standard for building a package, or isn't it? Because that is exactly what we are executing by default.

Issue #235, opened 2022-08-09
nodejs: crashes if a defined binary is dynamically built


tinybeachthor:
With: https://github.com/vercel/serve

The binary is compiled from a typescript file.

`package.json`:
```json
{
  ...
  "bin": {
    "serve": "./build/main.js"
  },
  "scripts": {
    "compile": "tsup ./source/main.ts",
    ...
  },
  ...
}
```

The `fix-package.py` script in `d2nPatchPhase` crashes (this is before the build step, so the binary is not built yet, but is expected by the script).

```command
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking sources
unpacking source archive /nix/store/2vx6vyhwvk9g931mmgs8g8mf1dpach39-yqrybhmmsbfzsyqzsvms5dj06plwfpva-source/
@nix { "action": "setPhase", "phase": "patchPhase" }
patching sources
@nix { "action": "setPhase", "phase": "d2nLoadFuncsPhase" }
d2nLoadFuncsPhase
@nix { "action": "setPhase", "phase": "d2nPatchPhase" }
d2nPatchPhase
Traceback (most recent call last):
  File "/nix/store/3hqxxgs2z739dyzkk0ha4hllxc32a9dk-fix-package.py", line 98, in <module>
    symlink_bin(f'{out}/lib/node_modules/.bin/', package_json)
  File "/nix/store/3hqxxgs2z739dyzkk0ha4hllxc32a9dk-fix-package.py", line 95, in symlink_bin
    link(name, relpath)
  File "/nix/store/3hqxxgs2z739dyzkk0ha4hllxc32a9dk-fix-package.py", line 82, in link
    os.chmod(relpath, 0o777)
FileNotFoundError: [Errno 2] No such file or directory: './build/main.js'
```

PR #248, opened 2022-08-11
test: add integration tests

tinybeachthor:
Add integration tests to detect regressions for #221, #235.
Run tests in github action.


DavHau:
Wonderful! Could you add a comment to the top of each flake.nix explaining in one or two sentences why the test exists / what it does?

yusdacra:
This is great, but doesn't this basically copy all of `examples` testing code? Maybe instead of copying it all we could parametrize the testing code so we can pass `examples` dir and `integration-tests` dir etc.

tinybeachthor:
@DavHau 
Added descriptions for the integration tests

@yusdacra 
The test runner is not the same, for examples we are running `nix check` and don't evaluate haskell packages, for the integration tests we want to actually build the default package.
I personally don't mind code duplication in test setups. Unifying and parameterizing the scripts would only subtract from readability.

Issue #179, opened 2022-06-16
haskell: implement discoverer.


DavHau:
The discoverer should be able to read the `name` from the cabal file without using IFD. This will unbreak `nix flake show` etc.
Also this will allow us to decide which translator should be used for a given haskell project.
As a reference, see other discoverers in `./src/subsystems/xxx/discoverers`

m1-s:
I am on it. I will also try to make the discoverer support multipackage projects.

DavHau:
Just note that projects can consist of more than one package. Not every single package needs to be detected by the discoverer as an individual project. Usually all packages which share the same lock file belong to a single project detected by the discoverer as `one` project.

PR #250, opened 2022-08-15
add Haskell discoverer

m1-s:
Work done:
- Added haskell/stack multiproject example

To undraft:
- Add haskell discoverer

Fixes https://github.com/nix-community/dream2nix/issues/179


src/subsystems/haskell/translators/stack-lock/default.nix
```diff
@@ -170,16 +170,14 @@ in {
         };
 
         # name of the default package
-        defaultPackage = cabal.description.package.name;
+        defaultPackage = l.mapAttrsToList (n: v: n) builtPackages;
```
DavHau, line 173: This should only be a string (the name of the default package; just pick one).

PR #252, opened 2022-08-15
fix(rust): don't replace path deps if they are already in source

yusdacra:


johnae:
@yusdacra Yeah this fixes it! Thanks.

Issue #177, opened 2022-06-16
haskell: detect GHC version during translation


DavHau:
see: https://github.com/nix-community/dream2nix/pull/164#discussion_r895712272
Also the hidden packages should be dynamically generated depending on the ghc version.
This can be done by parsing this file:
https://gitlab.haskell.org/bgamari/ghc-utils/-/blob/master/library-versions/pkg_versions.txt

bezmuth:
@DavHau why is generating the list of hidden packages required? If they are all contained within ghc why should it be needed to know what version these packages are?

DavHau:
> @DavHau why is generating the list of hidden packages required? If they are all contained within ghc why should it be needed to know what version these packages are?

I think we only want to specify a dependency on a builtin package if the required version is different from the one shipped with ghc. If a default version of a builtin is required, the nixpkgs haskell architecture wants this package to be `null`.
But maybe this could be solved differently as well. I'm not sure.

DavHau:
> @DavHau why is generating the list of hidden packages required? If they are all contained within ghc why should it be needed to know what version these packages are?

I think we only want to specify a dependency on a builtin package if the required version is different from the one shipped with ghc. If a default version of a builtin is required, the nixpkgs haskell architecture wants this package to be `null`.
But maybe this could be solved differently as well. I'm not sure.

DavHau:
> @DavHau why is generating the list of hidden packages required? If they are all contained within ghc why should it be needed to know what version these packages are?

I think we only want to specify a dependency on a builtin package if the required version is different from the one shipped with ghc. If a default version of a builtin is required, the nixpkgs haskell architecture wants this package to be `null`.
But maybe this could be solved differently as well. I'm not sure.

DavHau:
This has been solved by @leungbk

PR #257, opened 2022-08-18
Remove hardcoded ghc version.

bezmuth:
This removes the hardcoded ghc version (8107) and instead uses one specified in the settings for stack-lock or auto detected for cabal-plan.

In reference to https://github.com/nix-community/dream2nix/issues/177  some more work still needs to be done to flesh out the hidden package dynamic generation but I think this is a step in the right direction.

### DIFF
--- a/src/specifications/subsystems/haskell/dream.lock.example.json
+++ b/src/specifications/subsystems/haskell/dream.lock.example.json
@@ -1,8 +1,8 @@
 {
   "_subsystem": {
     "compiler": {
-    "name" = "ghc",
-    "version" = "8.10.7"
+      "name": "ghc",
+      "version": "8.10.7"
     }
   }
 }
--- a/src/subsystems/haskell/builders/default/default.nix
+++ b/src/subsystems/haskell/builders/default/default.nix
@@ -35,17 +35,22 @@ in {
     # Example:
     #   produceDerivation name (mkDerivation {...})
     produceDerivation,
-    compiler ? null,
     ...
   } @ args: let
     b = builtins;
 
     compiler =
-      if args ? compiler
-      then args.compiler
-      else
-        pkgs.haskell.packages."${subsystemAttrs.compiler}"
-        or (throw "Could not find ${subsystemAttrs.compiler} in pkgs");
+      pkgs
+      .haskell
+      .packages
+      ."${subsystemAttrs.compiler.name}${
+        l.stringAsChars (c:
+          if c == "."
+          then ""
+          else c)
+        subsystemAttrs.compiler.version
+      }"
+      or (throw "Could not find ${subsystemAttrs.compiler.name} version ${subsystemAttrs.compiler.version} in pkgs");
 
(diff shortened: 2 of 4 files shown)


bezmuth:
> 

I chose to store the compiler versions using just numbers to match how they are formatted in nixpkgs, could you give an example of how you want the version numbers to be formatted?

DavHau:
> I chose to store the compiler versions using just numbers to match how they are formatted in nixpkgs, could you give an example of how you want the version numbers to be formatted?

How about storing it like this:
```nix
{
  compiler = {
    name = "ghc";
    version = "8.10.7";
  };
}
```

src/specifications/subsystems/haskell/dream.lock.example.json
```diff
@@ -0,0 +1,8 @@
+{
+  "_subsystem": {
+    "compiler": {
+      name = "ghc",
+      version = "8.10.7"
```
leungbk, line 5:
``` suggestion
    "name" = "ghc",
    "version" = "8.10.7"
```

src/subsystems/haskell/builders/default/default.nix
```diff
@@ -35,10 +35,18 @@ in {
     # Example:
     #   produceDerivation name (mkDerivation {...})
     produceDerivation,
+    compiler ? null,
```
DavHau, line 38: Actually there is not need to have this as an argument. There is no UI to set these arguments

src/subsystems/haskell/translators/stack-lock/default.nix
```diff
@@ -237,8 +237,15 @@ in {
...
-        # TODO: put ghc version here
-        subsystemAttrs = {};
+        subsystemAttrs = {
+          compiler = "${args.compiler.name}${
+            l.stringAsChars (c:
+              if c == "."
+              then ""
+              else c)
+            args.compiler.version
+          }";
```
DavHau, line 247: Also here, the type should be an attrset, instead of a string.

src/subsystems/haskell/builders/default/default.nix
```diff
@@ -35,10 +35,18 @@ in {
...
     ...
   } @ args: let
     b = builtins;
 
+    compiler =
+      if args ? compiler
+      then args.compiler
+      else
+        pkgs.haskell.packages."${subsystemAttrs.compiler}"
```
DavHau, line 47: `subsystemAttrs.compiler` is a name version attrset now.

src/specifications/subsystems/haskell/dream.lock.example.json
```diff
@@ -0,0 +1,8 @@
+{
+  "_subsystem": {
+    "compiler": {
+    "name" = "ghc",
+    "version" = "8.10.7"
```
DavHau, line 5: json is invalid

DavHau:
Thanks for that

Issue #240, opened 2022-08-10
Language request: PHP


aszenz:
PHP has standardized on composer as it's package manager, would be great to support it here.

raboof:
It would be useful to have a composer translator and php builder, perhaps taking inspiration from https://github.com/svanderburg/composer2nix

I see the composer.lock file format does have a `shasum` field, but it seems to be left empty for the composer.lock files that I've seen so far, not sure what's going on there. Some more discussion on the hash input at https://github.com/composer/composer/issues/2540 .

wmertens:
I sometimes have to touch PHP and I see that there's often a `composer dump-autoload` step that seems to add extra php files.

Also, some PHP modules seem to use relative imports to get configuration from the main project, or is that rare? So if the `vendor` dir were a symlink to the nix store, those modules wouldn't work.

Basically, I'm wondering how the vendor dir would best be built.

aszenz:
Some frameworks also have  composer plugins which add their own php files.

Case in point Symfony flex

wmertens:
So, two modes, one just symlinks a vendor dir (when all components behave), and the other mode copies the entire thing?

aszenz:
It would be cool if we test this packaging solution against demo repos of popular frameworks like symfony demo:

https://github.com/symfony/demo

Having demo apps working out of the box would be a great DX for nix usage in php ecosystem

aszenz:
It is also a good way to document how to build php apps which depend on nodejs/webpack for their front-end using dream2nix since it supports building packages for multiple ecosystems.

tinybeachthor:
- [ ] add meta to packages 
  - like #294
- [ ] php versions 
  - right now it's just php 8.1 from nixpkgs or abort if incompatible
  - get a php version from https://github.com/fossar/nix-phps
- [ ] override package version 
  - usually packages don't have `version` field in `composer.json` resulting in package names like `name-unknown`
  - add a subsystem attribute to allow setting the version field

PR #258, opened 2022-08-18
php: add translator `composer-lock`

tinybeachthor:
Towards #240

Seems like `composer.lock` does not write hashes for dependency archives,
for a pure translator we will need to use git sources only, therefore building everything from source.


tinybeachthor:
@DavHau 

impure translator does make a lot of sense, a lot of php projects don't commit `composer.lock` (some to get auto-updating-like behaviour, some just don't).
+
Docs entry on php subsystems (translators only so far).
\=
#262

Next up builder + examples

PR #261, opened 2022-08-21
WIP: indexers: make use of source-less translators

DavHau:
Previously the output of an indexer was a list of sources.
Since #259, dream2nix supports translation without source.
This change updates the indexers to support that feature.
With this change, the output of each indexer will be a list of project specifications instead of sources.

The new model:
  - Covers more use cases
  - Allows to translate from arbitrary data. For example {translator="apt"; name="htop"} could be a valid project specification.
  - Moves the logic for fetching inside the translator instead of using dream2nix fetchers. (less calls to nix)
  - does not use dream2nix makeFlakeOutputs logic in order to execute translation scripts (This is better, as depending on our high level abstractions would make refactoring/deprecating such more difficult)
  
  
TODO:
 - [x] update app .#translate
 - [x] update app .#translate-index
 - [x] fix output for indexer libraries-io
 - [x] fix output for indexer npm (can probably be removed anyways)
 - [ ] fix output of indexer creates-io
 - [ ] fix output of indexer creates-io-simple
 - [ ] fix makeOutputsForIndexes
 
 @yusdacra As we now have the multi purpose libraries-io indexer, do we still need the `creates-io` and `creates-io-simple` indexers, or should we delete them?

### DIFF
--- a/src/subsystems/nodejs/translators/npm/default.nix
+++ b/src/subsystems/nodejs/translators/npm/default.nix
@@ -1,0 +1,101 @@
+{
+  dlib,
+  lib,
+  ...
+}: {
+  type = "impure";
+
+  # the input format is specified in /specifications/translator-call-example.json
+  # this script receives a json file including the input paths and specialArgs
+  translateBin = {
+    # dream2nix utils
+    apps,
+    subsystems,
+    utils,
+    # nixpkgs dependenies
+    bash,
+    coreutils,
+    git,
+    jq,
+    moreutils,
+    nodePackages,
+    openssh,
+    writeScriptBin,
+    ...
+  }:
+    utils.writePureShellScript
+    [
+      bash
+      coreutils
+      git
+      jq
+      moreutils
+      nodePackages.npm
+      openssh
+    ]
+    ''
+      # accroding to the spec, the translator reads the input from a json file
+      jsonInput=$1
+
+      # read the json input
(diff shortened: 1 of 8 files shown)


yusdacra:
I think we should still keep at least the `crates-io` indexer because `libraries-io` is a third-party service and I would like to not depend on it if possible and directly pull data from `crates-io`.

src/subsystems/nodejs/translators/npm/default.nix
```diff
@@ -0,0 +1,101 @@
...
+      fi
+
+
+      pushd $TMPDIR
+      newSource=$(pwd)
+
+      npm install $candidate --package-lock-only $npmArgs
+
+      jq ".source = \"$newSource\"" -c -r $jsonInput > $TMPDIR/newJsonInput
+      jq ".project.relPath = \"\"" -c -r $TMPDIR/newJsonInput | sponge $TMPDIR/newJsonInput
```
yusdacra, line 59:
```suggestion
      jq ".source = \"$newSource\" | .project.relPath = \"\"" -c -r $jsonInput > $TMPDIR/newJsonInput
```

DavHau:
@yusdacra Could you review my changes (specifically on the crates-io indexer)? This is my first time ever touching rust code.

yusdacra (approved):
crates-io changes look fine to me, everything else does too, LGTM

PR #262, opened 2022-08-23
php: impure translator `composer-json` + fix `composer-lock`

tinybeachthor:
Towards #240  

Found out a couple of issues with `composer-lock` translator (mostly semver resolution).

- [x] fix php semver resolution
- [x] handle composer dependency `replace`
- [x] handle composer dependency `provide`
- [x] impure translator `composer-json`
- [x] fix dev dependencies resolution
- [x] add php subsystems docs section

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -10,6 +10,7 @@
 - [Python](./subsystems/python.md)
 - [Node.js](./subsystems/node.md)
 - [Haskell](./subsystems/haskell.md)
+- [PHP](./subsystems/php.md)
 
 # Concepts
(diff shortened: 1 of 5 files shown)


DavHau:
> * php example (can we do this without a builder?)

Actually we need a builder first, in order for the example to work.

tinybeachthor:
PHP packages have their own name but can also `replace` other packages or `provide` implementations - both of which can define their own versions.

This PR fixes the `composer-lock` translator to take `replace` and `provide` into account + some issues with the semver parsing (composer semver is a bit relaxed compared to the standard: ~ is treated almost like ^).

Issue #254, opened 2022-08-16
PR 249 breaks rust build


johnae:
This commit: https://github.com/nix-community/dream2nix/commit/d2b389de98c54d45db0134984b7d8e209cba6907 - part of PR https://github.com/nix-community/dream2nix/pull/249, results in an error that I didn't get in the previous commit https://github.com/nix-community/dream2nix/commit/07547c2284e9f521013556f361996be9aba3a3fd

```sh
❯ nix build '.#nu'
warning: Git tree '/home/john/Development/world' is dirty
warning: Using saved setting for 'extra-experimental-features = nix-command flakes' fr
om ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-substituters = https://nix-community.cachix.or
g https://insane.cachix.org https://cachix.cachix.org' from ~/.local/share/nix/trusted
-settings.json.
warning: Using saved setting for 'extra-trusted-public-keys = nix-community.cachix.org
-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs= insane.cachix.org-1:cLCCoYQKkmEb/M88UI
ssfg2FiSDUL4PUjYj9tdo4P8o= cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSo
MckM=' from ~/.local/share/nix/trusted-settings.json.
error: attribute 'nu-pretty-hex' missing

       at /nix/store/76yd1ixrinygifpwqp7vhzjy4l9g483h-source/src/lib/simpleTranslate2.
nix:196:15:

          195|             children =
          196|               depGraphWithFakeRoot."${node.name}"."${node.version}";
             |               ^
          197|
(use '--show-trace' to show detailed location information)
```

It's a Rust build so not sure what `node` has to do with it, though I'm very unfamiliar with the internals of this project.

yusdacra:
This shouldn't be related to the Rust subsystem I think but rather a more general issue I believe. But something might have changed that might have broken the code in the Rust subsystem. I'm not exactly sure why however. If you post a flake here I can debug it.

yusdacra:
Does https://github.com/nix-community/dream2nix/commit/0131c22c369f8c4c661aea7a395e94669a7f5974 fix your issue?

johnae:
Thanks @yusdacra. Seems that fixes the issue.

PR #264, opened 2022-08-23
simpleTranslate2: Fix behavour with a single dependency

bezmuth:
This (might) fix https://github.com/nix-community/dream2nix/issues/254

### DIFF
--- a/src/lib/simpleTranslate2.nix
+++ b/src/lib/simpleTranslate2.nix
@@ -193,7 +193,9 @@
 
           findCycles = node: prevNodes: cycles: let
             children =
-              depGraphWithFakeRoot."${node.name}"."${node.version}";
+              if l.hasAttr "${node.name}" depGraphWithFakeRoot
+              then depGraphWithFakeRoot."${node.name}"."${node.version}"
+              else [];
 
             cyclicChildren =
               lib.filter


yusdacra:
I can confirm this seems to fix the issue. Using this flake:
```nix
{
  description = "Test";

  inputs = {
    ## flakes
    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
    flake-utils.url = "github:numtide/flake-utils";
    dream2nix = {
      url = "github:nix-community/dream2nix";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    nushell = {
      url = "github:nushell/nushell/0.66.2";
      flake = false;
    };
  };

  outputs = {
    self,
    nixpkgs,
    flake-utils,
    dream2nix,
    ...
  } @ inputs: let
    l = nixpkgs.lib // builtins;

    d2n = rec {
      init = pkgs:
        dream2nix.lib.init {
          inherit pkgs;
          config.projectRoot = ./.;
        };
      makeOutputs = {
        pkgs,
        source,
        packageOverrides ? {},
      }: let
        outputs = (init pkgs).makeOutputs {
          inherit source packageOverrides;
        };
      in {
        inherit (outputs) packages;
      };
    };

    d2nPackages = flake-utils.lib.eachDefaultSystem (
      system: let
        pkgs = nixpkgs.legacyPackages.${system};
      in
        d2n.makeOutputs {
          inherit pkgs;
          source = inputs.nushell;
          packageOverrides."^.*".addDeps = {
            nativeBuildInputs = old: old ++ [pkgs.pkgconfig pkgs.python3];
            buildInputs = old: old ++ [pkgs.openssl pkgs.zstd pkgs.xorg.libX11];
            doCheck = false;
            buildFeatures = ["extra"];
            cargoUpdateHook = ''
              cargo add zstd-sys --features pkg-config --offline
              cargo update --package zstd-sys --offline
            '';
          };
        }
    );
  in
    d2nPackages;
}
```
`nix build .#nu` will fail if it points to the main dream2nix branch. If point to this fix branch, it builds fine.

DavHau:
Still there must be another bug in the rust translator which triggers this.
`nu-pretty-hex` does have dependencies, which are declared here: https://github.com/nushell/nushell/blob/main/crates/nu-pretty-hex/Cargo.toml
But dream2nix doesn't detect them and the dependencyGraph ends up empty.

The problem with this PR's fix is that it works around a corrupt dependencyGraph. If an node of the graph points to a non-existing node, the program should crash.

I just pushed an alternative fix: 0131c22c369f8c4c661aea7a395e94669a7f5974 which just omits the cyclic deps detection if the dependencyGraph is empty (which was the case with `nu-pretty-hex`).

bezmuth:
If dream2nix attempts to build nu-pretty-hex without the required dependencies shouldn't the build fail?

Issue #265, opened 2022-08-24
cargo-lock translator attempting to build packages outside of requirment for rust package 


bezmuth:
Attempting to build:
```nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
    dream2nix.url = "github:nix-community/dream2nix";
    src.url = "github:nushell/nushell/0.67.0";
    src.flake = false;
  };

  outputs = {
    self,
    dream2nix,
    nixpkgs,
    src,
    ...
  } @ inp:
    (
      let
        pkgs = nixpkgs.legacyPackages."x86_64-linux";
      in
      dream2nix.lib.makeFlakeOutputs {
      systems = ["x86_64-linux"];
      config.projectRoot = ./.;
      source = src;
      packageOverrides."^.*".addDeps = {
        nativeBuildInputs = old: old ++ [pkgs.pkgconfig pkgs.python3];
        buildInputs = old: old ++ [pkgs.openssl pkgs.zstd pkgs.xorg.libX11];
        doCheck = false;
        buildFeatures = ["stable"];
        cargoUpdateHook = ''
          cargo add zstd-sys --features pkg-config --offline
          cargo update --package zstd-sys --offline
        '';
        shellPath = "/bin/nu";
      };

      settings = [
        {
          translator = "cargo-lock";
        }
      ];
    });
}
```
Results in this error:
```bash
error: attribute 'Cargo.lock' missing

       at /nix/store/3akswl509msksq037fps8mzb5m66ba7z-source/src/subsystems/rust/translators/cargo-lock/default.nix:94:18:

           93|     # Parse Cargo.lock and extract dependencies
           94|     parsedLock = projectTree.files."Cargo.lock".tomlContent;
             |                  ^
           95|     parsedDeps = parsedLock.package;
(use '--show-trace' to show detailed location information)
------------------------------------------------------------
```

Attaching a debugger to the build process and checking what package is being built by viewing "projectTree.files" shows a package without a cargo lock is being built.
Furthermore this package isn't part of the dependencies for nushell and instead looks to be used as a sample of some sort (https://github.com/nushell/nushell/tree/main/samples/wasm).
I believe this is something to do with the discoverer loading all directories that contain a cargo.toml, however as the cargo-lock translator only expects dirs with both a cargo.toml and a cargo.lock it fails.

yusdacra:
[...]

PR #269, opened 2022-08-26
Remove projects when discovered translators do not match user specified translators

bezmuth:
This fixes https://github.com/nix-community/dream2nix/issues/265 by requiring projects that are discovered to have compatible translators before being sent for translation.

I'm still relatively unfamiliar with the internals of dream2nix so this might be a sketchy way of doing things.


bezmuth:
Alternate solution: https://github.com/nix-community/dream2nix/pull/270

PR #274, opened 2022-08-29
nodejs/electron: add 20.1.0

fufexan:
Depends on https://github.com/NixOS/nixpkgs/pull/188710, already merged.


DavHau:
We want to keep at least one candidate for each minor version. Please add 20.1 without removing 20.0

fufexan:
Sure.

PR #275, opened 2022-08-29
fix: yarn-lock: remove unneeded parsing of version

tgunnoe:
Removes unnecessary parsing of versions from `rawObj`s. 

Fixes issues with the following repositories, where the dependencies contain `@git+` in the version:
https://github.com/tgunnoe/dream2nix-attributes-mismatch
https://github.com/fufexan/misskey-d2n

TODO: unit tests needed for lockfiles that reference versions with `@git+`


tgunnoe:
There's this patch, as well:
https://matrix.to/#/!PrIEtcffTLLzUWvJOS:matrix.org/$PLAeIO8KGgHjf9KVoHeR51lFUH-DCQU5nkjrkJF-8kU?via=nixos.org&via=matrix.org&via=nixos.dev

tgunnoe:
I added the minimal integration test, but not a package that has the `npm:` version scheme because that one is not currently working, apparently. This did not affect that either. I'll open a separate PR to fix that case and add the test there as well for that format.

Issue #272, opened 2022-08-28
sveltekit: no devShell


bbigras:
```sh
nix shell "nixpkgs#nodejs"

# skeleton project, "no" to everything
npm create svelte@latest svelte-dream2nix

cd svelte-dream2nix

mv .gitignore .gitignore.bak

nix flake init -t github:nix-community/dream2nix#simple

git init
git add .

nix eval --impure --raw --expr 'builtins.currentSystem' > ./nix_systems && git add ./nix_systems

❯ nix develop
warning: Git tree '/home/bbigras/dev/svelte-dream2nix' is dirty
error: flake 'git+file:///home/bbigras/dev/svelte-dream2nix' does not provide attribute 'devShells.x86_64-linux.default', 'devShell.x86_64-linux', 'packages.x86_64-linux.default' or 'defaultPackage.x86_64-linux'

❯ nix flake show
warning: Git tree '/home/bbigras/dev/svelte-dream2nix' is dirty
git+file:///home/bbigras/dev/svelte-dream2nix
├───packages
│   └───x86_64-linux
│       ├───resolveImpure: package 'resolve'
│       └───svelte-dream2nix: package 'svelte-dream2nix'
└───projectsJson: unknown
```

bezmuth:
When attempting to `nix build .#svelte-dream2nix` you should get this error:
```bash
error: The nodejs package svelte-dream2nix contains unresolved impurities.
       Resolve by running the .resolve attribute of this derivation
       or by resolving all impure projects by running the `resolveImpure` package
```
You can fix this by running:
```bash
nix build .#resolveImpure

./result/bin/resolve
```
Then you should be able to run `nix develop`

bbigras:
Thanks, it works. Now I get:

```
[bbigras@laptop:~/dev/svelte-dream2nix]$ npm run dev

> svelte-dream2nix@0.0.1 dev /home/bbigras/dev/svelte-dream2nix
> vite dev

failed to load config from /home/bbigras/dev/svelte-dream2nix/vite.config.js
error when starting dev server:
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'stream' imported from /home/bbigras/dev/svelte-dream2nix/node_modules/@sveltejs/kit/src/exports/node/polyfills.js
    at new NodeError (internal/errors.js:322:7)
    at packageResolve (internal/modules/esm/resolve.js:732:9)
    at moduleResolve (internal/modules/esm/resolve.js:773:18)
[...]

PR #281, opened 2022-08-31
Add a fake devshell for impure packages

bezmuth:
This should make issues like https://github.com/nix-community/dream2nix/issues/272 less common

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -707,7 +707,8 @@ in let
         ''
         impureFakeDerivations
         // (realizedProjects.packages or {})
-        // {resolveImpure = resolveImpureScript;};
+        // {resolveImpure = resolveImpureScript;}
+        // {default = l.elemAt (map (key: l.getAttr key impureFakeDerivations) (l.attrNames impureFakeDerivations)) 0;};
 
       devShells =
         l.warnIf
--- a/src/default.nix
+++ b/src/default.nix
@@ -708,7 +708,11 @@ in let
         impureFakeDerivations
         // (realizedProjects.packages or {})
         // {resolveImpure = resolveImpureScript;}
-        // {default = l.head (l.mapAttrsToList (k: v: v) impureFakeDerivations);};
+        // (
+          if l.length (l.mapAttrsToList (k: v: v) impureFakeDerivations) != 0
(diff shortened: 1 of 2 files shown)


bezmuth:
A default should probably also be added so when users attempt to run `nix build .` the warning also displays

bezmuth:
Adding a default package allows the user to run both `nix build .` and  `nix develop` and receive a warning.

src/default.nix
```diff
@@ -707,7 +707,8 @@ in let
         ''
         impureFakeDerivations
         // (realizedProjects.packages or {})
-        // {resolveImpure = resolveImpureScript;};
+        // {resolveImpure = resolveImpureScript;}
+        // {default = l.elemAt (map (key: l.getAttr key impureFakeDerivations) (l.attrNames impureFakeDerivations)) 0;};
```
DavHau, line 711:
```suggestion
        // {default = l.head (l.mapAttrsToList (k: v: v) impureFakeDerivations);};
```

DavHau:
The case where `impureFakeDerivations` is empty [still needs to be handled](https://github.com/nix-community/dream2nix/actions/runs/3047009661/jobs/4952951206#step:5:399).

bezmuth:
This commit also modifies how the testing of the examples works, only running a check if impurities have been resolved, without this the default check would be run which errors as designed with the new behavior.
Good chance this is the wrong way to do this, might have been no need to modify these tests.

tests/examples/default.nix
```diff
@@ -46,7 +46,9 @@
           && [ "$(nix flake show --json | jq 'select(.packages."x86_64-linux".default.name)')" != "" ]; then
         nix eval --read-only --no-allow-import-from-derivation .#default.name
       fi
-      nix flake check "$@"
+      if [ "$(nix flake show --json | jq 'select(.packages."x86_64-linux".resolveImpure)')" == "" ]; then
```
DavHau, line 49: `resolveImpure` will not vanish, even if impurities have been resolved. We want to give the user the possibility to re-resolve, as this can be used to update the dependencies.

I think this change should not be necessary. If the tests were failing, then there is probably a problem caused by `src/default.nix` which should be fixed

PR #283, opened 2022-09-01
Add schema validation for specific subsystems

leungbk:
- Create validation for existing schemas
- Validate only for subsystems with existing schemas
- Also add schemas for nodejs and rust

Co-authored-by: a-kenji <aks.kenji@protonmail.com>

### DIFF
--- a/src/specifications/subsystems/nodejs/dream-lock-schema.json
+++ b/src/specifications/subsystems/nodejs/dream-lock-schema.json
@@ -5,6 +5,7 @@
   "properties": {
     "_subsystem": {
       "type": "object",
+      "required": ["nodejsVersion"],
       "properties": {
         "nodejsVersion": {
           "type": "string"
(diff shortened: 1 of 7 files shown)


leungbk:
On the `main` branch, the `flake.nix` corresponding to https://github.com/nix-community/dream2nix/blob/e928819208cd72acaa7331a5e28de3f1e436ff09/examples/nodejs_eslint/flake.nix#L1 and the one corresponding to https://github.com/nix-community/dream2nix/blob/e928819208cd72acaa7331a5e28de3f1e436ff09/examples/nodejs_workspace/flake.nix#L1 have differently shaped `dream-locks`; the former has `nodejsVersion` as an integer, while the latter has it as a string. So I did some type-massaging accordingly. There may be a mistake though. I'm not completely certain where some assignments occur.

DavHau:
This is really useful. Thanks a lot!

PR #286, opened 2022-09-02
php: builder

tinybeachthor:
Towards #240 

- [x] Add a builder
- [x] Add php example
- [x] dev shell
- [x] Update php docs
- [x] more robust semver implementation
- [x] discoverer

Disable packagist.org repo for composer, use [artifact](https://getcomposer.org/doc/05-repositories.md#artifact) or path repositories.

### DIFF
--- a/src/subsystems/php/builders/simple/default.nix
+++ b/src/subsystems/php/builders/simple/default.nix
@@ -1,0 +1,135 @@
+{...}: {
+  type = "pure";
+
+  build = {
+    lib,
+    pkgs,
+    stdenv,
+    # dream2nix inputs
+    externals,
+    ...
+  }: {
+    ### FUNCTIONS
+    # AttrSet -> Bool) -> AttrSet -> [x]
+    getCyclicDependencies, # name: version: -> [ {name=; version=; } ]
+    getDependencies, # name: version: -> [ {name=; version=; } ]
+    getSource, # name: version: -> store-path
+    # to get information about the original source spec
+    getSourceSpec, # name: version: -> {type="git"; url=""; hash="";}
+    ### ATTRIBUTES
+    subsystemAttrs, # attrset
+    defaultPackageName, # string
+    defaultPackageVersion, # string
+    # all exported (top-level) package names and versions
+    # attrset of pname -> version,
+    packages,
+    # all existing package names and versions
+    # attrset of pname -> versions,
+    # where versions is a list of version strings
+    packageVersions,
+    # function which applies overrides to a package
+    # It must be applied by the builder to each individual derivation
+    # Example:
+    #   produceDerivation name (mkDerivation {...})
+    produceDerivation,
+    ...
+  } @ args: let
+    l = lib // builtins;
+
+    # the main package
+    defaultPackage = packages."${defaultPackageName}"."${defaultPackageVersion}";
+
+    # packages to export
+    packages =
+      lib.mapAttrs
+      (name: version: {
+        "${version}" = makePackage name version;
+      })
+      args.packages;
+
+    # Generates a derivation for a specific package name + version
+    makePackage = name: version: let
+      dependencies = getDependencies name version;
+      allDependencies = let
+        getAllDependencies = deps: let
+          getSubdependencies = dep: let
+            subdeps = getDependencies dep.name dep.version;
+          in
+            getAllDependencies subdeps;
+        in
+          deps ++ (l.flatten (map getSubdependencies deps));
+      in
+        getAllDependencies dependencies;
+
+      intoRepository = dep: {
+        type = "path";
+        url = "${getSource dep.name dep.version}";
+        options = {
+          versions = {
+            "${dep.name}" = "${dep.version}";
+          };
+          symlink = false;
+        };
+      };
+      repositories = l.flatten (map intoRepository allDependencies);
+      repositoriesString =
+        l.toJSON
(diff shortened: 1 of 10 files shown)


tinybeachthor:
Tested on https://github.com/composer/composer, build and devShell both work as expected.

src/subsystems/php/discoverers/default/default.nix
```diff
@@ -0,0 +1,31 @@
...
+  subsystem,
+  ...
+}: let
+  l = lib // builtins;
+
+  # get translators for the project
+  getTranslators = path: let
+    nodes = l.readDir path;
+  in
+    l.optionals (nodes ? "composer.lock") ["composer-lock"]
```
leungbk, line 13:
```suggestion
    l.optional (nodes ? "composer.lock") "composer-lock"
```

DavHau:
> * [ ]  ~Add php examples~ (need the builder merged first)

That should not be necessary. You can add the example flake under /examples/php/flake.nix, and then execute from the repo root: `nix run .#tests-examples php`

Alternatively navigate to /examples/php, then run:
`nix flake lock --override-input dream2nix ../../`
and
`nix run .#resolveImpre` / `nix build` ..etc.

src/subsystems/php/discoverers/default/default.nix
```diff
@@ -0,0 +1,31 @@
...
+  dlib,
+  lib,
+  subsystem,
+  ...
+}: let
+  l = lib // builtins;
+
+  # get translators for the project
+  getTranslators = path: let
+    nodes = l.readDir path;
```
DavHau, line 11: call to readDir could can be prevented by accessing `tree`.
Example: `tree.files ? "composer.lock"`

src/subsystems/php/translators/composer-lock/default.nix
```diff
@@ -246,15 +245,15 @@ in {
       };
 
       # name of the default package
-      defaultPackage = composerJson.name;
+      defaultPackage = composerJson.name or "unknown";
```
DavHau, line 248: The `name` is already determined by the discoverer. I think we should just re-use `project.name` here.

DavHau:
I just added an example via e77c129ad6643cf7deb26d71e00c551e89ab285b. Working fine.

DavHau:
It seems like the composer example adds significant overhead to the test execution. That is probably because of the slow nix git fetching. Maybe we can replace the example with a project, that doesn't have many dependencies.

src/subsystems/php/discoverers/default/default.nix
tinybeachthor: done

src/subsystems/php/translators/composer-lock/default.nix
tinybeachthor: done

src/subsystems/php/builders/simple/default.nix
```diff
@@ -0,0 +1,161 @@
...
+      allDependencies = let
+        getAllDependencies = deps: let
+          getSubdependencies = dep: let
+            subdeps = getDependencies dep.name dep.version;
+          in
+            getAllDependencies subdeps;
+        in
+          deps ++ (l.flatten (map getSubdependencies deps));
+      in
+        getAllDependencies dependencies;
```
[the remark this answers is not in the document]
tinybeachthor: nice, didn't know about `genericClosure`

tinybeachthor:
@DavHau 
I've disabled checks on the composer example, which speeds it up a little.
Running the tests locally I get:
```
Execution times:
1 sec: _d2n-indexer
37 sec: php_composer        <<--
44 sec: nodejs_workspace
52 sec: _d2n-extended-new-subsystem
56 sec: _d2n-init-pkgs
56 sec: rust_ripgrep
83 sec: nodejs_prettier
85 sec: _d2n-extended
131 sec: nodejs_eslint
143 sec: debian-binary
144 sec: rust_set-rust-toolchain
155 sec: rust_no-cargo-lock
163 sec: nodejs_mattermost-webapp
175 sec: python_aiohttp
209 sec: haskell_stack-lock
210 sec: nodejs_eslint-aggreagted
281 sec: haskell_cabal-plan
```

aszenz:
Thanks for this, I recently tried this builder on one of the projects but got this error:

```
error: Constraint "as" could not be parsed
```
The project composer.json uses aliases as described here: 
https://getcomposer.org/doc/articles/aliases.md#aliases

tinybeachthor:
@aszenz 
The builder does not handle some less-common composer features at the moment (like composer plugins, php version constraints, composer api, aliases, ...).
Could you link to the project or if it's a private, make a small example of the composer.json that produces this error?

aszenz:
> @aszenz Thank you for noticing and reporting this. The builder does not handle some less-common composer features at the moment (like composer plugins, php version constraints, composer api, aliases, update/install script hooks, ...). Could you link to the project or if it's a private, make a small example of the composer.json that produces this error?

Sure here's a sample project on which running `nix build .#packages.x86_64-linux.app/php` fails with:
`error: Constraint "as" could not be parsed`

https://github.com/aszenz/php-app-nix

If i remove the `as` then i get an error about missing extensions, seems like the php extensions mentioned in `composer.json` are not available when the build runs

tinybeachthor:
@aszenz 
#296 fixes both issues (inline version alias, extra php extenstions)

PR #294, opened 2022-09-13
feat: add metadata extraction for nodejs and rust

### LINKED ISSUE #292: Include meta in package derivations
I'm using the following to generate a basic BOM:
```
nix bundle .#my-package --bundler bundlers#toReport
```

The report will include node packages, but the they are missing license and maintainer information.  For example:

```
express-4.18.1 (no license) maintained by nobody
```

It would be nice if derivations generated for each package included meta extracted from their `package.json`s.  Typically they will include license, author, and contributor information.  According to the [package.json docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license), licenses should generally use SPDX identifiers, which should hopefully make them easily to translate to `meta.license`.

yusdacra:
Adds `meta.license`, `meta.homepage` and `meta.description` attributes to nodejs and rust packages if we can extract them. This fixes #292.

### DIFF
--- a/examples/nodejs_prettier/flake.nix
+++ b/examples/nodejs_prettier/flake.nix
@@ -1,6 +1,6 @@
 {
   inputs = {
-    dream2nix.url = "github:nix-community/dream2nix";
+    dream2nix.url = "path:../..";
     src.url = "github:prettier/prettier/2.4.1";
     src.flake = false;
   };
--- a/examples/nodejs_prettier/flake.nix
+++ b/examples/nodejs_prettier/flake.nix
@@ -1,6 +1,6 @@
 {
   inputs = {
-    dream2nix.url = "path:../..";
+    dream2nix.url = "github:nix-community/dream2nix";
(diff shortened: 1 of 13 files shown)


examples/nodejs_prettier/flake.nix
```diff
@@ -1,6 +1,6 @@
 {
   inputs = {
-    dream2nix.url = "github:nix-community/dream2nix";
+    dream2nix.url = "path:../..";
```
tinybeachthor, line 3: It would be nice to not change this, so the examples can be standalone.

examples/rust_ripgrep/flake.nix
```diff
@@ -1,6 +1,6 @@
 {
   inputs = {
-    dream2nix.url = "github:nix-community/dream2nix";
+    dream2nix.url = "path:../..";
```
tinybeachthor, line 3: It would be nice to not change this, so the examples can be standalone.

PR #296, opened 2022-09-14
php: improvements & fixes

tinybeachthor:
- use php extensions from `composer.json` and dependencies
- handle inline version aliases (`dev-master as 1.3.1`)
- refactor & cleanup translators
- refactor semver


aszenz:
@tinybeachthor 
Cool the build now works on the main branch: https://github.com/aszenz/php-app-nix/tree/main

But when i try it on a simpler composer.json here on https://github.com/aszenz/php-app-nix/tree/dev-constraint-error it fails showing:

`error: Constraint "dev" could not be parsed`

tinybeachthor:
@aszenz 
Fixed now, it was trying to do semver match on all version constraints, even on something like `dev-master`. 
I've added a guard to only semver check constraints where it makes sense.

PR #297, opened 2022-09-14
php: granular builder

tinybeachthor:
Towards #240 
After #296

Add a granular builder.

- [x] granular builder
- [x] docs
- [x] devshell

### DIFF
--- a/docs/src/subsystems/php.md
+++ b/docs/src/subsystems/php.md
@@ -28,6 +28,10 @@ generate a dream2nix lockfile.
 
 ## Builders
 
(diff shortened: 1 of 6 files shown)


DavHau:
Is this builder preferably in any case? If it doesn't have downsides, we could consider removing the non-granular builder.

tinybeachthor:
@DavHau 
The simple builder can build some more complex packages, where the granular fails for now.
The granular builder cannot handle composer plugins yet correctly, while the simple builder can.
Maybe we can remove it later once the granular builder is expanded a little more.

PR #300, opened 2022-09-17
php: indexer + fixes

tinybeachthor:
Towards #240 

Index repo: https://github.com/tinybeachthor/dream2nix-php-auto


src/subsystems/php/builders/simple/default.nix
```diff
@@ -76,6 +76,25 @@
         args.packages
       );
 
+    # Unpack dependency
+    cleanDependency = name: version:
```
DavHau, line 80: This adds an extra derivation for each dependency. Could this be implemented differently?

src/subsystems/php/builders/simple/default.nix
```diff
@@ -136,13 +163,19 @@
           # disable packagist, set path repositories
           mv composer.json composer.json.orig
 
-          cat <<EOF >> $out/repositories.json
+          cat <<EOF > $out/repositories.json
           ${repositoriesString}
           EOF
+          cat <<EOF > $out/dependencies.json
```
DavHau, line 169: Could use [passAsFile](https://nixos.org/manual/nix/stable/language/advanced-attributes.html#adv-attr-passAsFile) indead?

tinybeachthor:
Currently the indexer successfully translates: `2117/2500` most depended-upon php packages.
https://github.com/tinybeachthor/dream2nix-php-auto

src/specifications/dream-lock-schema.json
```diff
@@ -101,7 +101,8 @@
                       "rootVersion": { "type": ["string", "null"] },
                       "path": { "type": "string" },
                       "type": { "type": "string" },
-                      "dir": { "type": "string" }
+                      "dir": { "type": "string" },
+                      "hash": { "type": "string" }
```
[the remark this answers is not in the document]
tinybeachthor: We don't, I made a mistake and was kinda mashing 2 different source types together.

src/subsystems/php/builders/simple/default.nix
tinybeachthor: Thanks! That's much nicer.

src/subsystems/php/builders/simple/default.nix
tinybeachthor: Not sure. I need to create `composer.json` if it does not exist in the package sources.
Maybe adding like an `afterUnpack` hook to the `getSource`, or skipping `getSource` and using `getSourceSpec` to invoke the fetcher more directly and squeeze in the hook somewhere there.
What are your thoughts?

tinybeachthor:
Currently the indexer successfully translates: `2218/2500` most depended-upon php packages.
https://github.com/tinybeachthor/dream2nix-php-auto

aszenz:
I ran into this error on a project with no dependencies so it has no `require section` in `composer.json`, not sure it's already fixed or not

Screenshot from 2022-09-21 22-37-34

tinybeachthor:
@aszenz Thanks for reporting this! It's already fixed on this branch ([d3b8463](https://github.com/nix-community/dream2nix/pull/300/commits/d3b8463c075dedb0ab57a307269b3f9ad3460ab6))

src/subsystems/php/builders/simple/default.nix
DavHau: Adding logic to the fetcher itself won't work, as it changes the output and therefore leads to a hash mismatch.

Yes, I think you could override the derivation returned by getSource.
Something like: `(getSource name version).overrideAttrs (old:{postUnpack = "";})`
But this will break for `path` sources which are not a derivation. So that case would have to be treated differently.

Still, it all sounds a bit hacky, so maybe it's better to keep the current implementation until we find a better solution.

tinybeachthor:
Currently the indexer successfully translates: `2226/2500` most depended-upon php packages.
https://github.com/tinybeachthor/dream2nix-php-auto

src/subsystems/php/builders/simple/default.nix
DavHau: Sounds good

aszenz:
I'm getting this error when building a project with doctrine dependencies, it's related to git submodules, not sure if it's related or not with the builder/indexer.

```
evaluating derivation 'git+file:///home/asrar/Desktop/app-php#packages.x86_64-linux.app-software/app-php'fatal: unable to connect to github.com:
github.com[0: 140.82.121.4]: errno=Connection timed out

> fatal: clone of 'git://github.com/doctrine/doctrine-sphinx-theme.git' into submodule path '/run/user/1000/nix-41446-52/docs/en/_theme' failed
Failed to clone 'docs/en/_theme'. Retry scheduled
evaluating derivation 'git+file:///home/asrar/Desktop/app-php#packages.x86_64-linux.app-software/app-php'fatal: unable to connect to github.com:
github.com[0: 140.82.121.4]: errno=Connection timed out

> fatal: clone of 'git://github.com/doctrine/doctrine-sphinx-theme.git' into submodule path '/run/user/1000/nix-41446-52/docs/en/_theme' failed
Failed to clone 'docs/en/_theme' a second time, aborting
error: program 'git' failed with exit code 1
```

tinybeachthor:
Fetching git repositories with submodules is broken, `builtins.fetchGit` can fetch submodules but it needs `sha256` specified, which does not work for dream2nix (only commit hashes in lock files).

I'll see if I can fetch the source without submodules, read the submodules info, fetch the submodules one by one and reconstruct the repo manually.

Example project with submodules: https://github.com/maxmind/GeoIP2-php

This is not specific for the php subsystem, it's an issue with the git fetcher, `bultins.fetchGit` implementation specifically.

src/subsystems/php/builders/simple/default.nix
tinybeachthor: `getSource` does not have `overrideAttrs` or `override`, seems like it returns the path directly and not a derivation, not sure.

tinybeachthor:
~`93%` translate success rate on the top 2500 packages

Most of the failures are due the submodules issue, I will need to investigate that more, not sure what's the problem there. `builtins.fetchGit` seems to work fine sometimes when I invoke it directly.

tinybeachthor:
The submodule issue is about `git://` protocol being deprecated on github as of March 2022.

fails: `git://github.com/maxmind/MaxMind-DB.git`
works: `git://github.com/maxmind/MaxMind-DB.git`

Some of the older packages still have the `git` protocol in `.gitmodules`.
This is not dream2nix or nix specific, and happens with plain `git clone --recurse-submodules`.
I trying to fix this in dream2nix is really out of scope.

tinybeachthor:
As it turns out, composer does not pull submodules when installing from source, so I will just do the same here.

tinybeachthor:
The submodules issues is fixed now, the rest of the failing packages will most likely need manual overrides.

@DavHau I think this is ready.

overrides/php/default.nix
```diff
@@ -0,0 +1,147 @@
+{
+  lib,
+  pkgs,
+  ...
+}: let
+  l = lib // builtins;
+in {
+  "phpstan/phpstan-phpunit" = {
+    fix-autoload-dev = {
```
DavHau, line 9: It seems like this override is required for many packages. Could this logic just be integrated into the builder? There are probably many more packages that will require it and we would otherwise have to add an override for each of them.

src/subsystems/php/translators/packagist/default.nix
```diff
@@ -0,0 +1,101 @@
...
+        {
+          \"type\": \"git\",
+          \"url\": \"$SOURCE_URL\",
+          \"rev\": \"$SOURCE_REV\"
+        }
+      " > $TMPDIR/sourceInfo.json
+
+      ${apps.callNixWithD2N} eval --json "
+        with dream2nix.utils.dreamLock;
+        replaceRootSources {
```
DavHau, line 91: For other translators I noticed that this nix eval call is very slow and increases the duration of translating 5000 significantly. This should probably be replaced with a python script or something similar. But this can be done in a future PR.

DavHau (commented):
Nice! Could you please rebase it. The conflicts are probably introduced by files being renamed by #304

src/subsystems/php/translators/packagist/default.nix
tinybeachthor: Nice. I will do this in another PR.

overrides/php/default.nix
tinybeachthor: Yes, actually now I realize that we don't want to apply these `dev` settings for dependencies at all. 
So just adding `--no-dev` flag for dependencies to `composer install` will take care of this.

PR #304, opened 2022-09-25
feat: use modules for builders

yusdacra:


src/modules/builders/builder/interface.nix
```diff
@@ -0,0 +1,37 @@
...
+      type = t.functionTo (t.functionTo t.attrs);
+      default = _: _: {};
+    };
+    type = lib.mkOption {
+      type = t.enum [
+        "pure"
+        "ifd"
+      ];
+    };
+    version = lib.mkOption {
```
DavHau, line 32: I think builders don't need `version`. That field was specific to translators, as I needed to introduce this during a re-factoring.

DavHau (approved):

PR #308, opened 2022-09-27
feat: make discoverers use modules

yusdacra:


src/modules/discoverers/implementation.nix
```diff
@@ -0,0 +1,16 @@
+{
+  config,
+  callPackageDream,
```
DavHau, line 3: not used here. can be removed

src/modules/discoverers/discoverer/default.nix
```diff
@@ -0,0 +1,5 @@
+{
```
DavHau, line 1: My plan was to try to keep the file hierarchy as flat as possible. So how about we replace
`./modules/discoverers/discoverer/{default.nix,interface.nix}`
with just `./modules/discoverers/interface-discoverer.nix`

DavHau (approved):

src/modules/discoverers/discoverer/default.nix
yusdacra: Oops, i was supposed to leave that as a file, but at some point i also had some implementation bits so thats why i did it like that. Maybe though to make it more flat we can move out the actual interface bits to just `discoverer` in `modules`, same for builder and translator.

DavHau: I think it would be nice if the directories under ./modules have the same name as the prefix for which they declare options.

To follow this pattern, we could put it under ./modules/interfaces.discoverer/default.nix

yusdacra: I put it under `src/modules/interfaces.discoverer`. I also moved the default discoverer implementation to `src/modules/discoverers.default-discoverer`.

PR #314, opened 2022-10-01
Speed up indexers

tinybeachthor:
Replace a call to `callNixWithD2N` with a python script.

### DIFF
--- a/src/apps/replaceRootSources/replaceRootSources.py
+++ b/src/apps/replaceRootSources/replaceRootSources.py
@@ -1,0 +1,41 @@
+import sys
+import json
+
+def replace_root_sources(lock, newSource):
+
+    packages = lock['_generic']['packages']
+    sources = lock['sources']
+
+    for name, version in packages.items():
+
+        original = sources[name][version]
+
+        fixed = newSource
+
+        if 'dir' in original:
+            fixed['dir'] = original['dir']
+        elif 'relPath' in original:
+            fixed['dir'] = original['relPath']
+
+        sources[name][version] = fixed
+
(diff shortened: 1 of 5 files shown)


src/apps/replaceRootSources/replaceRootSources.py
```diff
@@ -0,0 +1,41 @@
...
+    for name, version in packages.items():
+
+        original = sources[name][version]
+
+        fixed = newSource
+
+        if 'dir' in original:
+            fixed['dir'] = original['dir']
+        elif 'relPath' in original:
+            fixed['dir'] = original['relPath']
```
DavHau, line 18: @yusdacra I saw that the `relPath` thing comes from your original implementation. But what is the purpose? According to `src/specifications/dream-lock-schema.json`, relPath cannot occur in a source.
Also, it would be nice if you could review the rest of this function. Thanks

yusdacra: relPath was used before in `path` sources iirc, but it seems to have been removed so i don't think we need to have a check for that here

yusdacra (approved):
Seems to work fine, tested with `nix run .#translate '{"name":"ripgrep","version":"13.0.0","id":"ripgrep","translator":"crates-io"}' ./test`. Only the `relPath` check needs to be removed and then it should be fine.

src/apps/replaceRootSources/replaceRootSources.py
tinybeachthor: Removed the `relPath` check

PR #317, opened 2022-10-02
Racket: add impure translator and simple builder

leungbk:


src/subsystems/racket/translators/racket-impure/generate-dream-lock.rkt
```diff
@@ -0,0 +1,150 @@
...
+                                  (list (cons (string->symbol package-name)
+                                              (make-immutable-hash
+                                               `((0.0.0 . ,(make-immutable-hash
+                                                            `((type . "path")
+                                                              (path . ,src-path))))))))
+                                  (set-map names-of-overridden-packages
+                                           (lambda (name)
+                                             (cons (string->symbol name)
+                                                   (make-immutable-hash
+                                                    `((0.0.0 . ,(make-immutable-hash
```
leungbk, line 130: The `version` property that people sometimes put in their `info.rkt` is essentially meaningless since every commit that occurs after version X is cut will replace update the repo with the new source but still keep it at version X.

src/subsystems/racket/discoverers/default/default.nix
```diff
@@ -0,0 +1,37 @@
...
+  ...
+}: let
+  l = lib // builtins;
+
+  discover = {
+    tree,
+    topLevel ? true,
+  }:
+    if (tree ? files."info.rkt")
+    then [
```
DavHau, line 14: According to this implementation, the recursion stops, whenever a `info.rkt` file is found.
If a racket project contains another racket project inside a sub-directory, the sub-project will be ignored. Should we cover that case as well?

DavHau (approved):

DavHau:
Thanks for all of this! If you find some time it would be nice if you could add a section about racket to the `./docs`

src/subsystems/racket/discoverers/default/default.nix
leungbk: This was intentional on my part. The situation I wanted to avoid can be seen here:

- https://github.com/stchang/parsack/blob/master/parsack-lib/info.rkt
- https://github.com/stchang/parsack/blob/master/parsack-lib/parsack/info.rkt

The first of these two files is effectively the `package.json`. I have heard (don't remember where, but I want to say it was from the Racket Discord channel) that people sometimes use files like the second one, contained within top-level projects, to set variables, and not to declare sub-projects. So scanning indiscriminately for `info.rkt` files would incorrectly register the second `info.rkt` as a project root.

At the root of the aforementioned `parsack` repo, you will find that there are four sibling projects, each containing their own top-level `info.rkt`. This is the most common practice for repos that contain more than one package; I'm actually not aware of any scenarios where a sub-project is wholly contained within another project.

PR #323, opened 2022-10-06
refactor: cleanup old code, refactor some module stuff

yusdacra:
- remove old module code from src/lib
- remove `subsystems` related code
- move subsystem loading to into its own file to avoid infinite recursion errors
- change interfaces to fit the module system better (ie. `translate`, `translateBin` and `build` no longer need to be "loaded" because we can directly use pkgs and others from module system)
- fixup fetchers, translators and builders

after this I will work on moving dlib to src/modules as well as utils and also integrate apps and indexers to module system


src/subsystems/nodejs/translators/npm/default.nix
```diff
@@ -1,29 +1,27 @@
...
+      bash
+      coreutils
+      git
+      jq
+      moreutils
+      nodePackages
+      openssh
+      python3
+      ;
+  in
```
DavHau, line 24:
```suggestion
  translateBin =
```

src/subsystems/nodejs/translators/npm/default.nix
```diff
@@ -1,29 +1,27 @@
...
     utils.writePureShellScript
     [
```
DavHau, line 26:
```suggestion
    (with pkgs; [
```

src/modules/translators/implementation.nix
```diff
@@ -1,25 +1,33 @@
...
   config = {
     # The user can add more translators by extending this attribute
-    translators = funcs.import_ collectedModules;
-
-    /*
-    translators wrapped with extra logic to add extra attributes,
-    like .translateBin for pure translators
-    */
-    translatorInstances = funcs.instantiate config.translators loader;
+    translators = config.lib.mapAttrs (_: wrapTranslator) importedModules;
```
DavHau, line 29: The problem I see with this, is that the user cannot add translators easily anymore via the module system by adding an attribute `translators.new-translator = ...`. Now the user manually has to take care to call wrapTranslator.

If we have it separated like before, the user could just add to `translators` and `translatorInstances` would always contain the wrapped version of the translator.

yusdacra: I forgot to take that into account before I changed it to this so I will just change it back to how it was before. It used an `implementation.nix` in `interfaces.translator` to load these stuff so it would work just with `translators` but that caused infinite recursion errors and I wasn't able to fix it.

yusdacra: I was able to find a different solution so I used that.

DavHau:
Thanks a lot for putting in the effort!

PR #328, opened 2022-10-14
Add cabal-freeze support

leungbk:
This is almost done, but I need some help to figure out how to remove IFD.

I am guessing that adding the `all-cabal-json` and `ghc-utils` repos as flake inputs would solve the issue, but I'm not sure how to do it cleanly; something like https://github.com/leungbk/dream2nix/commit/5cae258c0f80238173bb220ffb7d49446952d3c0 (not on this branch) might be too error-prone, plus I'm not sure if it's very clean to simply pass those two repos' sources as a default value of `inputs` to `src/default.nix` (which I haven't done; I think something like that would be necessary to make the tests pass). Assuming that works around IFD, is there a way of passing down those repos that is cleaner than drilling `inputs` all the way?

### DIFF
--- a/src/subsystems/haskell/builders/simple-haskell/default.nix
+++ b/src/subsystems/haskell/builders/simple-haskell/default.nix
@@ -124,6 +124,9 @@ in {
           doCheck = false;
           doBenchmark = false;
 
+          # FIXME: this skips over the default package if its name isn't set properly
+          configureFlags = subsystemAttrs.cabalFlags."${name}"."${version}" or [];
+
           libraryToolDepends = libraryHaskellDepends;
           executableHaskellDepends = libraryHaskellDepends;
           testHaskellDepends = libraryHaskellDepends;
--- a/src/subsystems/haskell/builders/simple-haskell/default.nix
+++ b/src/subsystems/haskell/builders/simple-haskell/default.nix
@@ -2,6 +2,7 @@
   pkgs,
   utils,
   externals,
+  inputs,
   lib,
   ...
 }: let
@@ -142,15 +143,16 @@ in {
         specified in the dream-lock
         */
         // (
-          l.optionalAttrs
-          (
-            (name != defaultPackageName)
-            && cabalFiles ? "${name}#${version}"
-          )
+          l.optionalAttrs (name != defaultPackageName)
           {
-            preConfigure = ''
-              cp ${cabalFiles."${name}#${version}"} ./${name}.cabal
-            '';
+            preConfigure =
+              if cabalFiles ? "${name}#${version}"
+              then ''
+                cp ${cabalFiles."${name}#${version}"} ./${name}.cabal
+              ''
+              else ''
+                cp ${inputs.all-cabal-json}/${name}/${version}/${name}.cabal ./
+              '';
           }
         )
         # enable tests only for the top-level package
--- a/src/subsystems/haskell/translators/cabal-freeze/default.nix
+++ b/src/subsystems/haskell/translators/cabal-freeze/default.nix
@@ -1,0 +1,295 @@
+{
+  dlib,
+  lib,
+  name,
+  pkgs,
+  inputs,
+  ...
+}: let
+  l = lib // builtins;
+in {
+  type = "pure";
+
+  /*
+  Automatically generate unit tests for this translator using project sources
+  from the specified list.
+
+  !!! Your first action should be adding a project here. This will simplify
+  your work because you will be able to use `nix run .#tests-unit` to
+  test your implementation for correctness.
+  */
+  generateUnitTestsForProjects = [
+    (builtins.fetchTarball {
+      url = "https://github.com/leungbk/ghcid/tarball/1a1aa2f3ee409a0044340f2759d21b64b56b0010";
+      sha256 = "sha256:1mp33xkyyb4jqqriczai80sqwlrjcwd94l7bv709ngwjqy56h2n9";
+    })
+  ];
+
+  /*
+  Allow dream2nix to detect if a given directory contains a project
+  which can be translated with this translator.
+  Usually this can be done by checking for the existence of specific
+  file names or file endings.
+
+  Alternatively a fully featured discoverer can be implemented under
+  `src/subsystems/{subsystem}/discoverers`.
+  This is recommended if more complex project structures need to be
+  discovered like, for example, workspace projects spanning over multiple
+  sub-directories
+
+  If a fully featured discoverer exists, do not define `discoverProject`.
+  */
+  discoverProject = tree:
+  # Example
+  # Returns true if given directory contains a file ending with .freeze
+    l.any
+    (filename: l.hasSuffix ".freeze" filename)
+    (l.attrNames tree.files);
+
+  # translate from a given source and a project specification to a dream-lock.
+  translate = {
+    /*
+    A project returned by `discoverProjects`
+    Example:
+
+    {
+      "dreamLockPath": "packages/optimism/dream-lock.json",
+      "name": "optimism",
+      "relPath": "",
+      "subsystem": "nodejs",
+      "subsystemInfo": {
+        "workspaces": [
+          "packages/common-ts",
+          "packages/contracts",
+          "packages/core-utils",
+        ]
+      },
+      "translator": "yarn-lock",
+      "translators": [
+        "yarn-lock",
+        "package-json"
+      ]
+    }
+
+    */
+    project,
+    /*
+    Entire source tree represented as deep attribute set.
+    (produced by `prepareSourceTree`)
+
+    This has the advantage that files will only be read once, even when
+    accessed multiple times or by multiple translators.
+
+    Example:
+      {
+        files = {
+          "package.json" = {
+            relPath = "package.json"
+            fullPath = "${source}/package.json"
+            content = ;
+            jsonContent = ;
+            tomlContent = ;
+          }
+        };
+
+        directories = {
+          "packages" = {
+            relPath = "packages";
+            fullPath = "${source}/packages";
+            files = {
+
+            };
+            directories = {
+
+            };
+          };
+        };
+
+        # returns the tree object of the given sub-path
+        getNodeFromPath = path: ...
+      }
+    */
+    tree,
+    # arguments defined in `extraArgs` (see below) specified by user
+    ghcVersion,
+    defaultPackageName,
+    defaultPackageVersion,
+    ...
+  } @ args: let
+    # get the root source and project source
+    rootSource = tree.fullPath;
+    projectSource = "${tree.fullPath}/${project.relPath}";
+    projectTree = tree.getNodeFromPath project.relPath;
+
+    parsedCabalFreeze = l.pipe projectTree.files [
+      l.attrNames
+      (
+        l.findFirst (l.hasSuffix ".freeze")
+        (throw "No cabal.project.freeze file in the tree")
+      )
+      projectTree.getNodeFromPath
+      (l.attrByPath ["fullPath"] "")
+      (import ./parser.nix {inherit dlib lib;})
+    ];
+
+    haskellUtils = import ../utils.nix {inherit inputs lib pkgs;};
+
+    hiddenPackages = haskellUtils.ghcVersionToHiddenPackages."${ghcVersion}";
+
+    serializedRawObjects = l.filter ({name, ...}: (! hiddenPackages ? ${name})) parsedCabalFreeze.packagesAndVersionsList;
+
+    cabalData =
+      haskellUtils.batchFindJsonFromCabalCandidates
+      serializedRawObjects;
+
+    cabalFreezeFlags = l.pipe serializedRawObjects [
+      (l.map ({
+        name,
+        version,
+      }:
+        l.nameValuePair name {"${version}" = parsedCabalFreeze.cabalFlags."${name}" or [];}))
+      (flagsAlist: [(l.nameValuePair defaultPackageName {"${defaultPackageVersion}" = parsedCabalFreeze.cabalFlags."${defaultPackageName}" or [];})] ++ flagsAlist)
+      l.listToAttrs
+    ];
+  in
+    dlib.simpleTranslate2.translate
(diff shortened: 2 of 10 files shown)


yusdacra:
You don't pass `inputs`, instead declare it as a flake input and then in `flake.nix` add it to `externalPaths`. Then you can access it via `externalSources`.

DavHau:
@leungbk In the example project the settings were just necessary because the `stack-lock` translator was matching on all projects with a `.cabal` file. I fixed that via https://github.com/nix-community/dream2nix/pull/328/commits/0b956108947f18cf83d8b83e7a2210db367d173c and now the example builds

leungbk:
After rebasing, one problem I notice is that I use the Cabal files from the Hackage tarballs, which may contain stale metadata. In my case, I have an overly-stringent constraint that is preventing the build from succeeding; [`ghc-bignum  >=1.0  && <1.1`](https://github.com/haskellari/integer-logarithms/blob/a54ae7adbc579d1c665a5f19d32bdd3fd797d200/integer-logarithms.cabal#L68) is in the stale Cabal file, whereas the updated metadata has [`ghc-bignum >=1.0 && <1.4`](https://hackage.haskell.org/package/integer-logarithms-1.0.3.1/dependencies).

I see that in the dream2nix Haskell builder, this is dealt with by passing a name, version, and Cabal-file hash to `fetchCabalFile`, and fetching the corresponding Cabal file with the matching hash. 

However, for the cabal-freeze translator here, we aren't given any Cabal-file hashes by the freeze file. Directly copying the Cabal files provided by `all-cabal-json` will not always be correct since those Cabal files may not necessarily be up to date.

`cabal.project.freeze` provide an `index-state` field that looks like `index-state: hackage.haskell.org 2022-10-13T02:49:29Z`. Ideally, I could use this `index-state` to directly find the right Cabal-file revision, but I'm not aware of any obvious API that enables this.

I'm not entirely sure how something like `cabal build` works with freeze files at all; from skimming https://cabal.readthedocs.io/en/3.8/cabal-project.html and https://cabal.readthedocs.io/en/3.8/cabal-config-and-commands.html, it sounds like, assuming the naming of `cabal.project` and `cabal.project.freeze` are consistent, a `cabal build` will attempt to build according to the freeze file. But I assume it also does some form of revision-fetching, since I don't encounter my aforementioned version-constraint problem when I manually run `cabal build` in my repo, outside of a Nix sandbox.

leungbk:
When running the `unit-tests` for the pure translators, the `src` directory of dream2nix is installed into [its own path in the Nix store](https://github.com/nix-community/dream2nix/actions/runs/3255088451/jobs/5344035726#step:5:1756), so the [use of a relative path](https://github.com/nix-community/dream2nix/blob/19b63ebd0a4a8d3bd3096b0d7dbe80d2b06f1cc2/src/default.nix#L20-L25) in `src/default.nix` [prevents the root directory](https://github.com/nix-community/dream2nix/actions/runs/3255088451/jobs/5344035726#step:5:456) from being found. I'm not sure how to cleanly fix this.

DavHau:
> When running the `unit-tests` for the pure translators, the `src` directory of dream2nix is installed into [its own path in the Nix store](https://github.com/nix-community/dream2nix/actions/runs/3255088451/jobs/5344035726#step:5:1756), so the [use of a relative path](https://github.com/nix-community/dream2nix/blob/19b63ebd0a4a8d3bd3096b0d7dbe80d2b06f1cc2/src/default.nix#L20-L25) in `src/default.nix` [prevents the root directory](https://github.com/nix-community/dream2nix/actions/runs/3255088451/jobs/5344035726#step:5:456) from being found. I'm not sure how to cleanly fix this.

This should work:
`fix-unit-tests.patch`
```patch
diff --git a/tests/unit/default.nix b/tests/unit/default.nix
index 4e7ac62..96cb4f2 100644
--- a/tests/unit/default.nix
+++ b/tests/unit/default.nix
@@ -22,6 +22,6 @@ in
     nix
   ]
   ''
-    export dream2nixSrc=${dream2nixWithExternals}
+    export dream2nixSrc=${../../.}/src
     ${pythonEnv}/bin/pytest ${self}/tests/unit -n $(nproc) -v "$@"
```

> Directly copying the Cabal files provided by all-cabal-json will not always be correct since those Cabal files may not necessarily be up to date.

`all-cabal-json` always contains the the latest revision of each cabal file. So the cabal file taken from there is only outdated if the `all-cabal-json` input is outdated.

> cabal.project.freeze provide an index-state field that looks like index-state: hackage.haskell.org 2022-10-13T02:49:29Z. Ideally, I could use this index-state to directly find the right Cabal-file revision, but I'm not aware of any obvious API that enables this.

AFAIK, cabal's index (http://hackage.haskell.org/01-index.tar.gz) is a tarball that is appended to on each index update. So you can go back in time by ignoring all data in the tarball after the given timestamp. Not exactly sure what tool can be used in order to view the tarball at a specific date in time. With this command, you can at least see that the tarball contains several revisions of the same file:
```
curl http://hackage.haskell.org/01-index.tar.gz | gzip -d | tar --list | grep zlib/0.6.1.1/zlib.cabal
```
So theoretically we could use `01-index.tar.gz` to build a DB of all revisions of all cabal files (marked with timestamps). Maybe that could be added to all-cabal-json.

Anyways, I think we should just go with the latest revision that we can get from all-cabal-json for now. If you like you can still add this index state handling in another PR.

> But I assume it also does some form of revision-fetching (or alternatively version-relaxing), since I don't encounter my aforementioned version-constraint problem when I manually run cabal build in my repo, outside of a Nix sandbox.

It would be really valuable to find out why it behaves differently inside the sandbox.

src/subsystems/haskell/translators/utils.nix
```diff
@@ -45,4 +45,71 @@ in {
...
+      (targetBuildDepends ++ buildToolDepends ++ condBuildDepends);
+  in
+    depNames;
+
+  # XXX: This might be better as a function.  See
+  # https://github.com/nixos/nixpkgs/blob/773c2a7afa4cc94f91d53d8cb35245cbf216082b/pkgs/development/haskell-modules/configuration-ghc-9.4.x.nix#L49-L56
+  # Packages we shouldn't always overwrite with null, but presently do anyway, are
+  # 1) terminfo, when we cross-compile
+  # 2) xhtml, when ghc.hasHaddock is set to false
+  ghcVersionToHiddenPackages = l.pipe "${inputs.ghc-utils}/library-versions/pkg_versions.txt" [
```
DavHau, line 88: This is simply amazing!

leungbk:
This PR should be ready for review. However, the CLIs I tried to build with a `cabal.project.freeze` in the Nix sandbox required some extra hand-holding beyond the scope of this PR: when I try building `ghcid`, I end up having to build the Haskell package `zlib`, which the dream2nix Haskell builder struggles to do regardless of what translator is used; when I try building `cabal2json` with a `cabal.project.freeze`, I find that some dependencies in one of its dependencies are hidden behind `condTree`-related stuff, which we don't presently handle in our translator utilities.

I decided to try building `unordered-containers` with a `cabal.project.freeze` I added myself. It builds, but I don't know how to easily test it since it's not a CLI. I've attempted to export a Haskell environment containing the library so that I can test some of the library's APIs. Stuff like

```diff
diff --git a/src/subsystems/haskell/builders/simple-haskell/default.nix b/src/subsystems/haskell/builders/simple-haskell/default.nix
index 0c9a9db..a7e6bbb 100644
--- a/src/subsystems/haskell/builders/simple-haskell/default.nix
+++ b/src/subsystems/haskell/builders/simple-haskell/default.nix
@@ -163,6 +163,8 @@ in {
       # apply packageOverrides to current derivation
       produceDerivation name pkg;
   in {
-    inherit packages;
+    # inherit packages;
+    # packages = {default = compiler.ghcWithPackages (_: [packages.default]);};
+    packages = packages // {thing = {"1.0" = compiler;};};
   };
 }
-- 
```

doesn't work. I've also tried modifying the testing flake like

```diff
diff --git a/examples/haskell_cabal-freeze/flake.nix b/examples/haskell_cabal-freeze/flake.nix
index 344ba17..7d435d1 100644
--- a/examples/haskell_cabal-freeze/flake.nix
+++ b/examples/haskell_cabal-freeze/flake.nix
@@ -9,15 +9,19 @@
     self,
     dream2nix,
     src,
-  } @ inp: (dream2nix.lib.makeFlakeOutputs {
-    systems = ["x86_64-linux"];
-    config.projectRoot = ./.;
-    source = src;
-    settings = [
-      {
-        translator = "cabal-freeze";
-      }
-      {subsystemInfo = {defaultPackageName = "unordered-containers";};}
-    ];
-  });
+  } @ inp:
+    (dream2nix.lib.makeFlakeOutputs {
+      systems = ["x86_64-linux"];
+      config.projectRoot = ./.;
+      source = src;
+      settings = [
+        {
+          translator = "cabal-freeze";
+        }
+        {subsystemInfo = {defaultPackageName = "unordered-containers";};}
+      ];
+    })
+    // {
+      packages."x86_64-linux"."myghc" = self.pkgs.ghc;
+    };
 }
-- 
```

but this doesn't work either.

Finally, I'm not sure why this happens on the CI: https://github.com/nix-community/dream2nix/actions/runs/3263735047/jobs/5363257618#step:5:448

I don't encounter that problem locally. Does the CI environment not have Git?

src/subsystems/haskell/translators/cabal-freeze/default.nix
```diff
@@ -276,5 +277,12 @@ in {
       examples = ["9.0.2" "9.4.1"];
       type = "argument";
     };
+
+    defaultPackageName = {
+      description = "Name of default package";
+      default = "main";
+      examples = ["main" "default"];
+      type = "argument";
+    };
```
leungbk, line 286: Is there a way of setting the package name without adding a new entry to `extraArgs`?

leungbk:
> all-cabal-json always contains the the latest revision of each cabal file. So the cabal file taken from there is only outdated if the all-cabal-json input is outdated.

Right, that's what I meant; the `all-cabal-json` input will generally lag behind since we won't be updating the corresponding flake input in dream2nix all the time. I'm not sure how else we can guarantee a reproducible build though, so I guess `all-cabal-json` will have to do.

src/subsystems/haskell/builders/simple-haskell/default.nix
```diff
@@ -125,6 +125,9 @@ in {
           doCheck = false;
           doBenchmark = false;
 
+          # FIXME: this seems to skip over the default package
+          configureFlags = subsystemAttrs.cabalFlags."${name}" or [];
```
leungbk, line 129: In the `cabal.project.freeze` files I generated, I generally noticed that only one version of a package ended up being prescribed, so I didn't bother to key the attrset according to both name and version. 

But I should probably reshape this anyway since I think in general Haskell closures can end up with more than one version of a package.

Is there a way of specifying the version of the `defaultPackage` that is cleaner than simply adding another entry to `extraArgs`?

ParetoOptimalDev:
I'll be happy to test this out on some of my and various projects if it helps.

Should I be safe testing with 31858?

leungbk:
> I'll be happy to test this out on some of my and various projects if it helps.
> 
> Should I be safe testing with [318586a](https://github.com/nix-community/dream2nix/commit/318586a750a8d4cc4096bed3293d38d31d179144)?

Thanks for being willing to test. Yes, it should be fine to test at that commit, with the caveat I mentioned [above](https://github.com/nix-community/dream2nix/pull/328#issuecomment-1280540438) that there is a pretty good chance that whatever you test may require improvements to dream2nix that go beyond the scope of this PR.

src/subsystems/haskell/translators/cabal-freeze/default.nix
DavHau: Maybe regex the .cabal files for the name and version?
Alternatively cabal2json could be used but this introduces a layer of IFD.

DavHau:
> I end up having to build the Haskell package zlib, which the dream2nix Haskell builder struggles to do regardless of what translator is used

zlib fails because of the missing zlib c-library input? We can fix that by making use of the `pkgconfig-depends` field found in the `call-cabal-json` data.
We'd just need to know the mapping of pkg-config names to nixpkgs names. cabal2nix already has that.
If you take a look at the `haskell` branch of dream2nix, I already [started importing the name mapping](https://github.com/nix-community/dream2nix/blob/haskell/src/haskell-pkg-config.json) from `cabal2nix/Name.hs`.
Just the [guessing logic](https://github.com/NixOS/cabal2nix/blob/0d1a5a7d227683e9d099665f645a8618ec48ef94/cabal2nix/src/Distribution/Nixpkgs/Haskell/FromCabal/Name.hs#L237) still needs to be ported to nix.
I did not have the time yet to complete this. If you are interested in working on this, it would be a great help.

> I find that some dependencies in one of its dependencies are hidden behind condTree-related stuff, which we don't presently handle in our translator utilities.

cabal2json just got an improvement to better render the condTree logic in the outputted JSON (see https://github.com/NorfairKing/cabal2json/pull/5), and `all-cabal-json` is already updated.
It should now be easier to add this feature to dream2nix.

---
Concerning the tests, here is a list of projects which we know for sure are building against the current dream2nix. Maybe just pick a project for there? https://mc.tomberek.info/#pkgs-haskell (check the green ones on [hercules-ci](https://hercules-ci.com/github/nix-community/dream2nix-auto-test/jobs/85))

> I'm also not sure if it's worth adding my flake.nix to the tests at all, since we don't actually try to build it in the tests-examples test. The tests-unit already tests translation.

You can make it build in the CI by un-commenting the `# checks = self.packages;`. I think, to have one tested example would be good. Though we should work on the scalability of our example tests very soon ;)

DavHau:
Oh I just saw that there is nothing to `un-comment`. For enabling the checks, see an example [here](https://github.com/nix-community/dream2nix/blob/1e0d20959aa64bafc7851dad509c2b9e1d10a24e/examples/nodejs_eslint/flake.nix#L25)

I think we should use a really simple package for the example test which does not have many dependencies. Maybe the [hello](https://hackage.haskell.org/package/hello) package? Otherwise `unordered-containers` seems fine for now.

PR #329, opened 2022-10-14
chore: manage all-cabal-json via flake input

DavHau:
related: https://github.com/nix-community/dream2nix/pull/328

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,7 +15,6 @@
 
     pre-commit-hooks.url = "github:cachix/pre-commit-hooks.nix";
     pre-commit-hooks.inputs.nixpkgs.follows = "nixpkgs";
-    # upstream flake-utils dep not supporting `aarch64-darwin` yet
     flake-utils-pre-commit.url = "github:numtide/flake-utils";
     pre-commit-hooks.inputs.flake-utils.follows = "flake-utils-pre-commit";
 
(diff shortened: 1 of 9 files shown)


flake.nix
```diff
@@ -150,7 +156,7 @@
     dream2nixFor = forAllSystems (system: pkgs:
       import ./src rec {
         externalDir = externalDirFor."${system}";
-        inherit externalPaths externalSources lib pkgs;
+        inherit all-cabal-json externalPaths externalSources lib pkgs;
```
yusdacra, line 159: Maybe instead of directly adding this and any potential future inputs here, we could add it to `externalSources` instead? Or maybe just a new attribute called `sources`. Ideally we should not really add random inputs to any interface IMO since we want to port these to the module system, so we should either reuse `externalSources` or decide on another attribute. Besides I think it would make it less cluttered anyways.

src/lib.nix
```diff
@@ -93,6 +94,7 @@
       inherit lib dlib externalSources;
       dream2nixConfig = config;
       dream2nixConfigFile = l.toFile "dream2nix-config.json" (l.toJSON config);
+      all-cabal-json = throw "all-cabal-json is not available before nixpkgs is imported";
```
yusdacra, line 97: I don't think we need nixpkgs for this, nor the system. Just not passing the system should work looking at `flake-compat.nix`, since all we are really doing is using it as a source, we don't access any default outputs (which seem to be the only place where system is used).

yusdacra (changes requested):

PR #339, opened 2022-10-28
feat: add a flake-parts module

yusdacra:
Implements a flake-parts module.

### DIFF
--- a/examples/_d2n-flake-parts/flake.nix
+++ b/examples/_d2n-flake-parts/flake.nix
@@ -1,0 +1,29 @@
+{
+  inputs = {
+    dream2nix.url = "path:../..";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    src.url = "github:BurntSushi/ripgrep/13.0.0";
+    src.flake = false;
+  };
+
+  outputs = {
+    self,
+    dream2nix,
+    flake-parts,
+    src,
+    ...
+  }:
+    flake-parts.lib.mkFlake {inherit self;} {
+      imports = [dream2nix.flakePartsModule];
+      systems = ["x86_64-linux"];
+      dream2nix = {
+        config.projectRoot = ./.;
+        projects = [
+          {
+            source = src;
+            settings = [{builder = "crane";}];
+          }
+        ];
+      };
+    };
+}
--- a/examples/_d2n-flake-parts/flake.nix
+++ b/examples/_d2n-flake-parts/flake.nix
@@ -16,15 +16,13 @@
   }:
     flake-parts.lib.mkFlake {inherit self;} {
       systems = ["x86_64-linux"];
-      imports = [dream2nix.flakePartsModule];
+      imports = [dream2nix.flakeModule];
       dream2nix = {
         config.projectRoot = ./.;
-        projects = [
-          {
-            source = src;
-            settings = [{builder = "crane";}];
-          }
-        ];
+        projects."ripgrep" = {
+          source = src;
+          settings = [{builder = "crane";}];
+        };
       };
       perSystem = {
         config,
--- a/examples/_d2n-flake-parts/flake.nix
+++ b/examples/_d2n-flake-parts/flake.nix
@@ -1,6 +1,7 @@
 {
   inputs = {
-    dream2nix.url = "path:../..";
+    dream2nix.url = "github:nix-community/dream2nix";
+    nixpkgs.follows = "dream2nix/nixpkgs";
     flake-parts.url = "github:hercules-ci/flake-parts";
     src.url = "github:BurntSushi/ripgrep/13.0.0";
     src.flake = false;
@@ -14,8 +15,8 @@
     ...
   }:
     flake-parts.lib.mkFlake {inherit self;} {
-      imports = [dream2nix.flakePartsModule];
       systems = ["x86_64-linux"];
+      imports = [dream2nix.flakePartsModule];
       dream2nix = {
         config.projectRoot = ./.;
         projects = [
@@ -24,6 +25,22 @@
             settings = [{builder = "crane";}];
           }
         ];
+      };
+      perSystem = {
+        config,
+        lib,
+        pkgs,
+        ...
+      }: let
+        inherit (config.dream2nix) outputs;
+      in {
+        packages.ripgrep = outputs.packages.ripgrep.overrideAttrs (old: {
+          buildInputs = (old.buildInputs or []) ++ [pkgs.hello];
+          postInstall = ''
+            ${old.postInstall or ""}
+            hello
+          '';
+        });
       };
     };
 }
--- a/examples/_d2n-flake-parts/flake.nix
+++ b/examples/_d2n-flake-parts/flake.nix
@@ -16,29 +16,14 @@
   }:
     flake-parts.lib.mkFlake {inherit self;} {
       systems = ["x86_64-linux"];
-      imports = [dream2nix.flakeModule];
+      imports = [dream2nix.flakeModuleBeta];
       dream2nix = {
         config.projectRoot = ./.;
-        projects."ripgrep" = {
+        # define an input for dream2nix to generate outputs for
+        inputs."ripgrep" = {
           source = src;
           settings = [{builder = "crane";}];
         };
-      };
-      perSystem = {
-        config,
-        lib,
-        pkgs,
-        ...
-      }: let
-        inherit (config.dream2nix) outputs;
-      in {
-        packages.ripgrep = outputs.packages.ripgrep.overrideAttrs (old: {
-          buildInputs = (old.buildInputs or []) ++ [pkgs.hello];
-          postInstall = ''
-            ${old.postInstall or ""}
-            hello
-          '';
-        });
       };
     };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -194,7 +194,7 @@
     # kept for compat
     lib2 = d2n-lib;
 
-    flakeModule = {
+    flakeModuleBeta = {
       imports = [./src/modules/flake-parts];
       dream2nix.lib = d2n-lib;
     };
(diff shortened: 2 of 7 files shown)


DavHau:
Nice, I was just playing with the thought of adding flake-parts support, then saw this PR.

Kranzes:
Good stuff Yusuf. Robert and I were showing @DavHau how cool flake-parts is when we were in Paris last week.

examples/_d2n-flake-parts/flake.nix
```diff
@@ -0,0 +1,44 @@
+{
```
DavHau, line 1: From this example alone, it is not obvious if defining `perSystem` is actually necessary to make things work or if it's just an optional thing for customization.
Maybe we could have two examples for flake parts. One minimal one, and another one that shows how to override the package/devShell.

src/modules/flake-parts/interface.nix
```diff
@@ -0,0 +1,59 @@
...
+      config = l.mkOption {
+        type = t.submoduleWith {
+          modules = [../config];
+        };
+        default = {};
+        description = ''
+          The dream2nix config.
+        '';
+      };
+      projects = l.mkOption {
```
DavHau, line 28: I wonder if this is the right time to properly specify what a project is. Especially the `settings` field and how it's merged currently is really hacky and confusing.
Though this might be a bigger task, because some existing code might need to be refactored in order to make it fit to a nicer interface.
Then there is the question if we want to have backward compatibility (and the extra complexity coming with it), or just rip off the band-aid.

Maybe we could merge this, but before we but it into the readme as the `default` way to use dream2nix, we improve/refactor the interface in an extra PR.

yusdacra: We can change the name from `projects` to something else, but I'm not sure what would be a better name. Maybe `sources` would be a better name? It would conflict with `source` but I feel like it might be better anyways. Changing the interface should be very simple, after we introduce modules for the arguments of `makeFlakeOutputs` we can just change the type to `t.submoduleWith { modules = [./src/modules/args]; }`. It's just being passed to `makeFlakeOutputs` anyways. So not much is going to change, maybe if we decide on a better name now we can avoid breaking that later though.

DavHau: How about renaming it to `inputs`.

Currently `makeFlakeOutputs` already has an argument `projects` https://github.com/nix-community/dream2nix/blob/2b7456e3d2f0053bc2474fb0c461dd468545277f/examples/debian-binary/flake.nix#L13

So maybe that could lead to confusion.

I think my final goal would be to make a module that is not based on makeFlakeOutputs. It could be designed, to have similar convenience as makeFlakeOuputs, but at the same time allows lower level access, for example specifying the projects manually instead of using the discoverers. 

Some people were mentioning that they have not been very happy with what the discovery phase delivered for them and would prefer to specify the `project` manually. Especially since sometimes, people want to use dream2nix on a single package of their monorepo. And dream2nix just processing everything it sees is not really compatible to that.

Anyways. for now Maybe we just keep it like that, maybe just rename it, and improve it later.

flake.nix
```diff
@@ -194,6 +194,11 @@
     # kept for compat
     lib2 = d2n-lib;
 
+    flakeModule = {
```
DavHau, line 197: Should be rename this to `flakeModuleBeta`, to make it really clear, that this is subject to changes?

src/modules/flake-parts/interface.nix
yusdacra: `inputs` sounds alright. Also I definitely agree with defining a project manually, but I don't think it really matters here. If we introduce that, nothing really would change in regards to this interface, since we pass everything to the correct dream2nix interface internally. We would continue to define things in the same way and just pass the args to dream2nix, only then we would change the type so that it validates the arguments.

flake.nix
yusdacra: That would be fine to me. I was thinking of adding a `lib.warn` but that might be too much nagging :shrug:

examples/_d2n-flake-parts/flake.nix
yusdacra: I removed that part of the example, for now. I think later we can make a more extended example covering everything, after things get a little bit more stable.

src/modules/flake-parts/interface.nix
Kranzes: `inputs` nor `sources` don't make sense to me compared to `projects` in my opinion. Inputs and sources are things very computing and data specific than `projects` which is more vague. It makes sense that a *project* has *inputs* and *sources*. But I don't see a how `input` == `project.

yusdacra: I agree with that, but it's hard to choose a good name because we also use `projects` in the interface currently other than here... I'd rather keep `project` but get rid of the `project` in interface later tbh.

Kranzes: Is the plan to move all the interface to flake-parts later on?

yusdacra: That would be nice IMO. Since we are already moving internals to modules and we will also move interfaces to modules, there isn't really any reason not to use `flake-parts`. If someone wants to avoid that, it would also be easy since they could just evaluate the modules themselves.

src/modules/flake-parts/implementation.nix
```diff
@@ -0,0 +1,49 @@
...
+        (
+          l.filterAttrs
+          (_: attrs: l.isAttrs attrs && l.hasAttr system attrs)
+          outputs
+        );
+    in {
+      config = {
+        dream2nix.outputs = systemizedOutputs;
+      };
+    };
```
DavHau, line 47: Why do we need the whole `perSystem` thing? As we already defined `flake`, isn't that redundant? Does it give the user any extra features?

yusdacra: It gives convenience yes. If you are in a system context, (ie. `perSystem`) you will be able to use `config.dream2nix.outputs.packages.ripgrep` instead of `config.dream2nix.outputs.packages.${system}.ripgrep`. Which removes one nesting so it is worth it IMO.

src/modules/flake-parts/interface.nix
DavHau: Agreed, going full modules is where I'd like to go. We can still have wrappers if required or alternative entry points.

DavHau:
Thanks a lot!

roberth:
The options are now rendered on [flake.parts](https://flake.parts/options/dream2nix.html) if that's ok with you.

src/modules/flake-parts/implementation.nix
roberth: For comparison, other modules like haskell-flake and pre-commit-hooks-nix exist mostly at the `perSystem` level. Users may expect to define things there instead of at the top level.
By taking configuration from the top-level instead of from the `perSystem` module, you deny users the ability to reference `system` in their dream2nix config. I'm not familiar with dream2nix yet, so I can't tell if this is a huge loss.

src/modules/flake-parts/implementation.nix
```diff
@@ -0,0 +1,49 @@
...
+        l.mapAttrs
+        (_: attrs: attrs.${system})
+        (
+          l.filterAttrs
+          (_: attrs: l.isAttrs attrs && l.hasAttr system attrs)
+          outputs
+        );
+    in {
+      config = {
+        dream2nix.outputs = systemizedOutputs;
```
roberth, line 45: Are these flake-style outputs? You could assign them to the options so that they're automatically exported.

yusdacra: Those are outputs that don't have a `system` yes. I didn't assign them here since we already export everything that was generated before in `config.flake` option. But kept this here in case people want to directly use any of the outputs that was generated by dream2nix.

yusdacra: Ah, that's true, I forgot about that, I thought `flake-parts` would pass the stuff set in `config.flake` to the `perSystem` options when possible. I will push a fix for that :+1:

src/modules/flake-parts/implementation.nix
DavHau: Going the `perSystem` route by default, would probably solve problems like: https://github.com/nix-community/dream2nix/issues/345

So as the `perSystem` logic is now provided by flake-parts, we could just remove all multi-system handling from the entire dream2nix and reduce complexity in places. What do you think @yusdacra

DavHau:
> The options are now rendered on [flake.parts](https://flake.parts/options/dream2nix.html) if that's ok with you.

Having flake-parts rendering documentation for all flake-parts supporting projects is a wonderful idea.
This just removes the whole burden of maintaining that on our side.
I think this could be a big selling point for flake-parts. `Integrate with flake-parts and get documentation for free`.

Are the docs updated automatically, and if yes, how often?

Kranzes:
Let's put the docs on their on repo and update them automatically with the new fancy `onSchedule` Hercules-CI job trigger.

roberth:
> Let's put the docs on their o[w]n repo and update them automatically

I liked the idea of having the options in one place. How it's set up, it will also detect name collisions.

> with the new fancy `onSchedule` Hercules-CI job trigger.

A feature which is to be released in a couple of days. :)
You can use part of the functionality with hercules-ci-agent master. You won't need to bring a personal access token in a couple of days from now.

> Are the docs updated automatically, and if yes, how often?

This is the plan. It could run a couple of times per day, no problem.

Kranzes:
> I liked the idea of having the options in one place. How it's set up, it will also detect name collisions.

The problem is that updating the docs will spam bunch of commits to the flake-parts repo, unless you want to put it on its own clean branch, but at that point it isn't much different than using a separate repo.

roberth:
Yeah I was thinking of doing it on a separate repo.

src/modules/flake-parts/implementation.nix
yusdacra: I just made #366 for this! Please check it out.

PR #366, opened 2022-11-11
refactor(flake-parts): use perSystem for defining inputs, drop makeFl…

yusdacra:
…akeOutputs usage

### DIFF
--- a/examples/_d2n-flake-parts/flake.nix
+++ b/examples/_d2n-flake-parts/flake.nix
@@ -1,6 +1,6 @@
 {
   inputs = {
-    dream2nix.url = "path:../..";
+    dream2nix.url = "github:nix-community/dream2nix";
     nixpkgs.follows = "dream2nix/nixpkgs";
     flake-parts.url = "github:hercules-ci/flake-parts";
     src.url = "github:BurntSushi/ripgrep/13.0.0";
--- a/examples/_d2n-flake-parts/flake.nix
+++ b/examples/_d2n-flake-parts/flake.nix
@@ -17,10 +17,12 @@
     flake-parts.lib.mkFlake {inherit self;} {
       systems = ["x86_64-linux"];
       imports = [dream2nix.flakeModuleBeta];
-      dream2nix = {
-        config.projectRoot = ./.;
+
+      dream2nix.config.projectRoot = ./.;
+
+      perSystem = {config, ...}: {
         # define an input for dream2nix to generate outputs for
-        inputs."ripgrep" = {
+        dream2nix.inputs."ripgrep" = {
           source = src;
           settings = [{builder = "crane";}];
         };
--- a/examples/_d2n-flake-parts/flake.nix
+++ b/examples/_d2n-flake-parts/flake.nix
@@ -1,6 +1,6 @@
 {
   inputs = {
-    dream2nix.url = "github:nix-community/dream2nix";
+    dream2nix.url = "path:../..";
     nixpkgs.follows = "dream2nix/nixpkgs";
     flake-parts.url = "github:hercules-ci/flake-parts";
     src.url = "github:BurntSushi/ripgrep/13.0.0";
@@ -17,8 +17,6 @@
     flake-parts.lib.mkFlake {inherit self;} {
       systems = ["x86_64-linux"];
       imports = [dream2nix.flakeModuleBeta];
-
-      dream2nix.config.projectRoot = ./.;
 
       perSystem = {config, ...}: {
         # define an input for dream2nix to generate outputs for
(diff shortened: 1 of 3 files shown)


src/modules/flake-parts/implementation.nix
```diff
@@ -5,44 +5,50 @@
...
+          inherit instance;
+          outputs =
+            # if only one input was defined, then only export outputs from
+            # that since there is nothing else
+            if l.length (l.attrNames outputsRaw) != 1
+            then outputsRaw
+            else l.head (l.attrValues outputsRaw);
+        };
+        devShells = mkDefaultRecursive (getAttrFromOutputs "devShells");
+        packages = mkDefaultRecursive (getAttrFromOutputs "packages");
```
roberth, line 51: You may consider using `lib.mkMerge` here to let the module system do the merging, unless you need to ignore collisions here.

src/modules/flake-parts/interface.nix
```diff
@@ -52,11 +31,27 @@ in {
...
+              type = t.attrsOf t.attrs;
+              default = {};
+              description = ''
+                A list of inputs to generate outputs from.
+                Each one takes the same arguments `makeOutputs` takes.
+              '';
+            };
             outputs = l.mkOption {
-              type = t.attrsOf t.raw;
+              type = t.either (t.attrsOf t.raw) (t.attrsOf (t.attrsOf t.raw));
```
roberth, line 50:
```suggestion
              type = t.lazyAttrsOf (t.lazyAttrsOf t.raw);
```

The either was only needed for the single `inputs` case, a consistent schema would be nicer; see earlier comment about 1 -> 2 -> 1.

`lazyAttrsOf` has better performance because for evaluating `foo.bar` it doesn't have to check all the other `foo.*` values just to produce a set of names that is then only queried for `bar`.

examples/_d2n-flake-parts/flake.nix
```diff
@@ -17,10 +17,12 @@
     flake-parts.lib.mkFlake {inherit self;} {
       systems = ["x86_64-linux"];
       imports = [dream2nix.flakeModuleBeta];
-      dream2nix = {
-        config.projectRoot = ./.;
+
+      dream2nix.config.projectRoot = ./.;
```
roberth, line 21: You could default the `projectRoot` to `self` in `flakeModule`. `self` is a module argument.

src/modules/flake-parts/implementation.nix
```diff
@@ -5,44 +5,50 @@
...
       config = {
-        dream2nix.outputs = systemizedOutputs;
+        dream2nix = {
+          inherit instance;
+          outputs =
+            # if only one input was defined, then only export outputs from
+            # that since there is nothing else
+            if l.length (l.attrNames outputsRaw) != 1
+            then outputsRaw
+            else l.head (l.attrValues outputsRaw);
```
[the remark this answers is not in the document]
yusdacra: Hmm, that's true, but I didn't think it would be too surprising. But on second thought it's probably just better to keep it consistent and have users do `let out = outputs."projectName"; in`.

src/modules/flake-parts/implementation.nix
```diff
@@ -5,44 +5,44 @@
...
+          l.mapAttrsToList
+          (_: output: mkDefaultRecursive output.${attrName})
           outputs
         );
     in {
       config = {
-        dream2nix.outputs = systemizedOutputs;
+        dream2nix = {inherit instance outputs;};
+        # TODO(yusdacra): we could combine all the resolveImpure here if there are multiple
+        # TODO(yusdacra): maybe we could rename outputs with the same name to avoid collisions?
```
DavHau, line 43: If we do this automatically, it might introduce confusion. When a project is removed, then another one might be re-named back to it's original name for example.
Let's better throw an error, and give the user the possibility to re-name a project manually

src/modules/flake-parts/implementation.nix
```diff
@@ -5,44 +5,44 @@
...
+        l.mkMerge (
+          l.mapAttrsToList
+          (_: output: mkDefaultRecursive output.${attrName})
           outputs
         );
     in {
       config = {
-        dream2nix.outputs = systemizedOutputs;
+        dream2nix = {inherit instance outputs;};
+        # TODO(yusdacra): we could combine all the resolveImpure here if there are multiple
```
DavHau, line 42: Sounds like a good idea. But also it get ugly as we now have multiple levels where `resolve` scripts are combined.
Maybe the solution gets much nicer once we refactor more part to use the module system. Each package could add it's resolve scripts to a list (or better attrset) of resolve scripts, for example.

PR #371, opened 2022-11-13
No discovery

DavHau:
This change allows to take control over the project definition completely and not use discovery at all.

Here an example flake that builds prettier using the current PR:
```nix
{
  inputs = {
    dream2nix.url = "github:nix-community/dream2nix/ba44952315ea9aa11f45d80b563ffeea996e06bf";
    nixpkgs.follows = "dream2nix/nixpkgs";
    flake-parts.url = "github:hercules-ci/flake-parts";
    src.url = "github:prettier/prettier/2.4.1";
    src.flake = false;
  };

  outputs = {
    self,
    dream2nix,
    flake-parts,
    src,
    ...
  }:
    flake-parts.lib.mkFlake {inherit self;} {
      systems = ["x86_64-linux"];
      imports = [dream2nix.flakeModuleBeta];

      perSystem = {config, ...}: {
        # define an input for dream2nix to generate outputs for
        dream2nix.inputs."prettier" = {
          source = src;
          projects = {
            prettier = {
              name = "prettier";
              subsystem = "nodejs";
              translator = "yarn-lock";
            };
          };
        };
      };
    };
}
```

### DIFF
--- a/examples/_d2n-define-projects-manually/flake.nix
+++ b/examples/_d2n-define-projects-manually/flake.nix
@@ -1,0 +1,35 @@
+{
+  inputs = {
+    dream2nix.url = "github:nix-community/dream2nix";
+    nixpkgs.follows = "dream2nix/nixpkgs";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    src.url = "github:prettier/prettier/2.4.1";
+    src.flake = false;
+  };
+
+  outputs = {
+    self,
+    dream2nix,
+    flake-parts,
+    src,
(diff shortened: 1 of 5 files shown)


yusdacra (changes requested):
Can we add thay example to examples? Also is setting `discoveredProjects` necessary in the example? From what I see it is not right?

DavHau:
That's true, I removed the `discoveredProjects` declaration and added an example.

yusdacra (approved):

PR #372, opened 2022-11-13
fix: Added type check for packagesDir option

### LINKED ISSUE #157: Ensure packagesDir is type checked to be a string with a relative directory
First, If you try to define `config.packagesDir` without the directory already existing (in this case, a dir name same as default-- but any name produces the same errors), you get:

`error: getting status of '/nix/store/7458lklxyckwz2gcifnv29vys0sybvmf-source/dream2nix-packages': No such file or directory`

Upon creating the directory and adding it to git, now running the command `nix run .#test.resolve` gives this error:

```
error: the string '.*/nix/store/y35q3pg84s5d855kchj1h2nj7wnrvmqa-dream2nix-packages' is not allowed to refer to a store path (such as '/nix/store/y35q3pg84s5d855kchj1h2nj7wnrvmqa-dream2nix-packages')

       at /nix/store/dq2r5b5rl9y2kvf4j37hajrmzacnz82h-source/src/lib/default.nix:199:10:

          198|       (baseNameOf path != "flake.nix")
          199|       && l.match ''.*/${sanitizedPackagesDir}'' path == null
             |          ^
          200|       && (l.any
```

Flake:
```
  {
    outputs = { self, dream2nix }@inputs:
      let
        supportedSystems = [ "x86_64-linux" ];
      in
        let
            systems = supportedSystems;
            config.projectRoot = ./.;
            config.packagesDir = ./dream2nix-packages;
          };
        in
          dream2nix.makeFlakeOutputs {
            pname = "test";
            source = ./.;
          };
  }
```

antotocar34:
Fixes #157

### DIFF
--- a/src/modules/config/interface.nix
+++ b/src/modules/config/interface.nix
@@ -1,6 +1,16 @@
 {lib, ...}: let
   l = lib // builtins;
   t = l.types;
+
+  relativePathString = l.types.mkOptionType {
+    name = "relativePathString";
+    description = "Relative path in a string.";
+    check = s:
+      (t.str.check s)
+      && (l.substring 0 2 s == "./");
+    descriptionClass = "noun";
+    inherit (t.str) merge;
+  };
 in {
   options = {
     disableIfdWarning = l.mkOption {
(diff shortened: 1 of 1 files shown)


antotocar34:
I tried to see whether it was possible to write a nice error message for UX purposes. I'm not certain but don't think this is possible. Here is the relevant code in `nixpkgs`:

https://github.com/NixOS/nixpkgs/blob/74b5398c530d09ee5e4e66e2b6383c2c50c904a6/lib/modules.nix#L746-L762

PR #379, opened 2022-11-16
feat: make utils use modules

yusdacra:


src/modules/utils/dream-lock.nix
```diff
@@ -1,18 +1,12 @@
-{
```
DavHau, line 1: We should probably rename `dream-lock.nix` to `dreamLock.nix` or instead change the attribute path to `utils.dream-lock.xyz`.

The structure of this now also violates a bit what we did with other modules.
Instead of having the whole utils interface in `utils/interface.nix` we could have the the different parts like `dreamLock` have it's own directory.
So that the interface and implementation of dreamLock reside in `./modules/utils.dreamLock/{interface.nix,implementation.nix}`

With this there is again a clear mapping between an attribute path and the location of its definition.

yusdacra: This should be done now. The dream-lock.nix and etc. were already modules but just resided in `utils` dir so I just moved them now :+1:

DavHau:
Simply wonderful. Thanks

Issue #374, opened 2022-11-14
docs: Nodejs quickstart documentation .gitignore error


antotocar34:
This is not such a terrible error, but it would be better if the user did not encounter any errors while following the quickstart example.
```
$ git clone https://github.com/piuccio/cowsay /tmp/my_project
$ cd /tmp/my_project
$ nix flake init -t github:nix-community/dream2nix#simple
warning: Using saved setting for 'extra-substituters = https://nix-community.cachix.org' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-trusted-public-keys = nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs=' from ~/.local/share/nix/trusted-settings.json.
warning: ignoring untrusted substituter 'https://nix-community.cachix.org'
refusing to overwrite existing file '/tmp/my_project/.gitignore'
 please merge it manually with '/nix/store/bz60fzw5br99jrkvljaw92zmalpavvh5-source/templates/simple/.gitignore'
wrote: /tmp/my_project/flake.nix
error: Encountered 1 conflicts - see above
```

DavHau:
Maybe we should just remove the .gitignore

PR #380, opened 2022-11-16
add comprehensive description to flake template & del `.gitignore`

hsjobeki:
removed the `.gitignore` from simple template as this always leads to conflicts with existing projects. 

Add a `description` to the flake template that describes how to use the template. (see in diff)
Add `.envrc` instead to allow for `direnv` compatibility

-> Future PR might introduce some logic for merging the `result` (and other nix files) line into an potentially existing `.gitignore`

potentially solves #374 #353


DavHau:
fixes #374 #353

DavHau:
Wonderful, thank you!

PR #382, opened 2022-11-16
Nodejs builder: modular phases

hsjobeki:
Moved all `bash` phases into seperate nix files.

pros:
- better clearance in default.nix
- better seperation
  - variables only used in one phase can be declared there
 
cons:
- in nix phases of mkDerivation are not meant to be modularized.
- overrides happen at build-time, thats why nix variables cannot be used directly in that phases if we want to allow overrides

future steps: 
- add shellcheck to the bash code
- imrove the override functionality
- improve isolation between phases

(maybe we can solve future steps directly in nixpkgs)


DavHau:
Generally I'm in favor of splitting thing up like this. This makes it simpler to get an overview of the builder without having to scroll through all the implementation details.

As we just noticed in the call, the devShell still needs to be fixed. If you can rebase and push the fix, I think this can be merged

PR #384, opened 2022-11-16
fix: nodejs devShell creates broken links

hsjobeki:
The devShell in nodeJs subsystem creates broken symlinks (unrelated to my other current PRs)

fixes the following two problems:

- `../$target` is creating a relative link, but $target already hold an nix store path, which is absolute
- line `for link in $(ls ${binDir}); do` the `ls` command should not be used in shell `for...in` as here an additional `magic` file `->` appears because it tries to split the string that is also holding an -> from the ls command. e.g. `ls $binDir` returns `tsc -> nix/store/..../.bin/tsc` but that string gets splitted and it creates an file for ->


Kazimazi:
For me this breaks symlinks.

hsjobeki:
Can you describe what doesnt work anymore ?

hsjobeki:
> For me this breaks symlinks.

Fixed in #388

PR #388, opened 2022-11-18
replace readlink by realpath

hsjobeki:
realpath resolves all links recursively where the executable for sure exists. 

Otherwise creating the devShell will fail.

readlink does only resolve one layer of links, which in some cases is not enough as if one link gets cleaned up after build it will break.

@Kazimazi this should resolve your recent problems.


Kazimazi:
works perfectly

PR #391, opened 2022-11-19
refactor: use flake-parts for the flake

yusdacra:
Adds flake-parts to use for the flake. ~~Ideally we would make this so that `src/modules/top-level.nix` is added to `imports` (with a little wrapper that adds it under `dream2nix` option), so we completely remove `src/default.nix` from the picture. I'm creating this as a draft so we can first do that first and then we can merge it.~~ I moved the interface and updaters to modules and separated the flake itself a bit more (templates moved under templates and tests apps moved under tests). `src/default.nix` is still there but it's probably better to keep it for compat anyways


DavHau (commented):
Maybe at some point we could remove `src/lib.nix` completely and instead implement makeFlakeOutputs as a wrapper for using the flake-parts module.

Also I feel the historically grown file structure could be refined.
We have:
- functions.
- dlib.
- utils.
- dream2nix-interface.

... which I feel overlap a bit and maybe we should not have this separation/or another separation.
What do you think?
Should we have just anything under `functions.{something}.{function}` in the future?

yusdacra:
> Maybe at some point we could remove `src/lib.nix` completely and instead implement makeFlakeOutputs as a wrapper for using the flake-parts module.

Yes, we should do that.

> Also I feel the historically grown file structure could be refined. We have:
> 
> * functions.
> * dlib.
> * utils.
> * dream2nix-interface.
> 
> ... which I feel overlap a bit and maybe we should not have this separation/or another separation. What do you think? Should we have just anything under `functions.{something}.{function}` in the future?

We can do that, I just didn't do it right now since I felt changing all the internal stuff in one go would be confusing. We can put all of them under `functions` and mark most of the functions as `internal` so that it's apparent it's not meant for public use.

DavHau:
Sounds good. If you can fixed the two commented things, feel free to merge it.

PR #395, opened 2022-11-21
WIP frozen python requirements

chaoflow:


DavHau:
@chaoflow regarding the name, `pinned` doesn't really give any clue on what format or what tool is used to create the lock file. It should at least reflect that it's related to pip, as there are many tool to manage dependencies besides it. `pip freeze` is usually the command that is used to create exactly that format that we parse here, so I thought that  this might be the best name to pick.

You can still handle the format more strictly, if you see it is a pip-compiled file with hashes. This can either be exposed as a new translator called `pip-compiled` or just in the already existing translator.

What do you think?

chaoflow:
@DavHau great! Let's see what we end up supporting.

PR #401, opened 2022-11-25
Support build requirements for pinned python

chaoflow:
@DavHau This makes changes to the simple-python builder. We need to check/align with other python translators.

### DIFF
--- a/examples/python_pip-freeze/flake.nix
+++ b/examples/python_pip-freeze/flake.nix
@@ -37,6 +37,10 @@
                 "requirements.txt"
                 "requirements-dev.txt"
               ];
+              # For now, just installed into same environment before the rest.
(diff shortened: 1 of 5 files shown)


DavHau:
I now merged https://github.com/nix-community/dream2nix/pull/397 which also did some changes on the builder.
It would be great if you could rebase this PR and add your changes to the builder, while checking to not break the other example flakes.
I'd propose we just focus on getting the package build work and do the devShells in another PR.

domenkozar:
Having requirements.txt support would be fantastic, what's missing here to merge?

phaer:
Work here seems to have stalled in the past 6 months? I believe the new v1 pip builder should work fine for this use-case? :)

PR #404, opened 2022-11-28
pre-commit checks: add black formatting for all python files

hsjobeki:
As we have a lot of python files it makes sense to check the formatting before allowing to commit.
Just as we currently do with `treefmt` and nix files

Black is one of the best python formatters i know. So i propose to just add it to the pre-commit checks.

This becomes especially usefull if we wwill create larger python builders / applications inside this repo.

Things changed:
- added black to flake.nix

- formatted all python files via `pre-commit run --all-files`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -193,10 +193,15 @@
           // {
             # passes through extra flags to treefmt
             format.type = "app";
-            format.program =
+            format.program = let
+              path = lib.makeBinPath [
+                alejandra.defaultPackage.${system}
+                pkgs.python3.pkgs.black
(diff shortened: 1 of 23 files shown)


DavHau:
We should not trust the user to run the pre-commit hooks. Please also add a test to our tests-format CI job (flake app).

DavHau:
There is no need to add black itself to pre-commit, as we already have treefmt and we can therefore integrate black into treefmt. I did that via d98d3efb4a315ce3db3aa513f90e51c54e544254

PR #405, opened 2022-11-29
Poetry: support build requirements

phaer:
This is a continuation of my work in #397. It adds rather naive support for a subsystem attr `buildRequirements` in the translator.

This means we can install those `buildRequirements` during `buildPhase` instead of hard-coding packages like `poetry-core` in `buildInputs`. That approach didn't work for packages that explicitly require another version of `poetry-core` than what's in nixpkgs, as `buildPhase` would try and fail to uninstall the nixpkgs version before installing the required version.

It's naive in the sense that it leaves version constraints and platform compatibility tags ("markers") to pip. Which might or might not be a good idea for `buildRequirements`. 

[X] actually install the built wheel
[X] make sure buildRequirements don't leak into the final derivation or use buildRequirements from nixpkgs

Needs to be integrated with  #401

### DIFF
--- a/src/subsystems/python/builders/simple-python/default.nix
+++ b/src/subsystems/python/builders/simple-python/default.nix
@@ -46,7 +46,6 @@
         [pkgs.autoPatchelfHook]
         ++ (with python.pkgs; [
           pip
-          poetry-core
           wheel
         ]);
       propagatedBuildInputs = [python.pkgs.setuptools];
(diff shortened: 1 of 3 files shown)


DavHau:
Sorry, I might have rebased on the wrong branch there. It now contains the experimental nodejs stuff from the `nodejs` branch, which it should not.

phaer:
@DavHau No worries, I took the liberty to force-push again, without anything from the `nodejs` branch but rebased on `main`, so with your fix for the lockfile v2 issue, not mine.

Just resumed working on it a bit yesterday in the morning, but didn't make much progress so far. Hope to make it to the weekly, but due to the re-schedule and another call right before that I can't promise anything atm.

DavHau:
> Note: test results sadly include a false positive here. `python_pip` mistakenly yields a build of its dependency `llhttp`, when it should yield `aiohttp`.

fixed via e82afbcb55433ef739d2ab1f31783d79380c645c

phaer:
@DavHau I believe this should be good to merge?

src/subsystems/python/builders/simple-python/default.nix
```diff
@@ -61,17 +68,32 @@
           fname=$(stripHash $fname)
           cp $file dist/$fname
         done
-        ${python}/bin/python -m pip install \
-          --find-links ./dist/ \
+
+        mkdir -p "$out/${python.sitePackages}"
+        export PYTHONPATH="$out/${python.sitePackages}:$PYTHONPATH"
+        ${python}/bin/python -m pip wheel --verbose --no-index --no-deps --no-clean --no-build-isolation --wheel-dir dist .
```
FRidh, line 74: If you make this `${python.interpreter}` it also becomes useful for pypy

src/subsystems/python/builders/simple-python/default.nix
```diff
@@ -61,17 +68,32 @@
...
+        pipInstallFlags="--find-links ./dist/ \
           --no-build-isolation \
           --no-index \
           --no-warn-script-location \
           --prefix="$out" \
           --no-cache \
-          . \
-          $pipInstallFlags
+          $pipInstallFlags"
+        ${python}/bin/python -m pip install $pipInstallFlags dist/*
```
FRidh, line 95: same here

Issue #387, opened 2022-11-17
Python subsystem progress thread


DavHau:
Missing translators:
- [ ] requirements.txt with hashes
- [ ] poetry
- [ ] pipenv

Missing builders:
- [ ] granular builder (each dependency built in separated derivation)

Other stuff:
- [ ] implement an indexer, in order to add stats about build failures to https://nix-community.github.io/dream2nix-auto-test
- [ ] current builder crashes on projects without a setup.py (https://github.com/nix-community/dream2nix/issues/358)

PR #406, opened 2022-12-01
Stub of a description of the python subsystem

chaoflow:
The goal of this PR is to sketch out where we want to go with the python subsystem.


phaer:
I've started noting down a few things and then asked myself if https://github.com/nix-community/dream2nix/issues/387 wasn't a better place? 

(also moved the file to notes, the docs/subsystems/python folder would be another option but we don't have any other README.md files in the src directory iteself atm)

chaoflow:
@phaer I'd prefer having things in a repo instead of issues and beyond that if it's meant for developers close to the code instead of elsewhere.

chaoflow:
@phaer Maybe we have a different view of the document. I intend it as full description of the python subsystem. Something like "current limitations" I'd see on an issue that aims to change something whereas in the document I'd focus on what is supposed to work and how.

DavHau:
`docs/src/subsystems/python.md` sounds like the right place for this.
see for example the [haskell subsystem page](https://nix-community.github.io/dream2nix/subsystems/haskell.html) where the development status is mentioned

con-f-use:
>  1. A python environment with all packages from one or more requirements.txt files with exact versions and optionally hashes.

This sounds a bit narrow.

The aim should be to support the entire [feature-set of requirements.txt](https://pip.pypa.io/en/stable/reference/requirements-file-format/) files. For instance there is a lot of people who use indices other than pypi.org, some of them self-hosted like [devpi](https://github.com/devpi/devpi). So at least `--index-url` support and some certificate stuff would be almost crucial. Then there is the url  and git syntax:

```
# requirements.txt
http://wxpython.org/Phoenix/snapshot-builds/wxPython_Phoenix-3.0.3.dev1820+49a8884-cp34-none-win_amd64.whl
some-lib @ git+https://github.com/owner/some-lib@0.1
```

phaer:
Closing this, as the v1 pip builder implements most of these features.

@con-f-use PRs welcome :)

PR #411, opened 2022-12-03
fix: python: poetry: incompatible with newer version of poetry.lock

DavHau:
https://github.com/python-poetry/poetry/commit/41706e649cf3ee9b066c59920b4e10e686f8c6d5 broke the parser.
Poetry changed it's lock file structure some time ago.
This fixes the translator crashing on the new format

@phaer

### DIFF
--- a/src/subsystems/python/translators/poetry/default.nix
+++ b/src/subsystems/python/translators/poetry/default.nix
@@ -53,12 +53,49 @@ in {
     };
 
     # use dream2nix' source tree abstraction to access json content of files
-    sources =
+    poetryLock =
       (projectTree.getNodeFromPath "poetry.lock").tomlContent;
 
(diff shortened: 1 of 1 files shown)


DavHau:
@phaer Oops I jsut noticed that you already fixed that in https://github.com/nix-community/dream2nix/pull/405.
Feel free to revert my change when you complete the PR instead of rebasing

phaer:
@DavHau No worries! I guess it became a bit confusing when the not-quite-finished poetry support has been merged, because its not totally clear whether it's something we support already or something still in development. 

It's interesting how both of us decided to support both old and new lock files. Thinking about it, supporting just the new one is probably good enough? I'd care about older versions for `pyproject.toml`, but I believe for lock files it might be good enough to just assume that those can be updated automatically? (i.e. by re-locking)

DavHau:
I think supporting both old and new format would be great as long as it's not too complex to do so.

Issue #407, opened 2022-12-01
Disable automatic discovery of projects by default


DavHau:
The automatic discovery was a nice idea but seems to lead to too many issues right now.
Projects are detected which should not be detected, people are confused, nix evaluation is crashing, etc.

For most people it's probably better to specify the projects manually.

For the `makeFlakeOutputs` interface, we could do the following:
- introduce a new field `apiVersion` which defaults to `0` but if set two `1`, disables automatic project discovery and forces the user to specify the projects manually. (this will become the default for our flake templates and examples)
- introduce a new field `discoverProjects` which if set to `true` re-enables automatic project discovery even if `apiVersion = 1`.
- change all examples under `./examples` to use `apiVersion = 1` and specify projects manually. They should basically use [this style](https://github.com/nix-community/dream2nix/blob/ce047a9ceaf1040dc14e393cc3243d22dd0dbcdc/examples/_d2n-define-projects-manually/flake.nix) instead of [this style](https://github.com/nix-community/dream2nix/blob/ce047a9ceaf1040dc14e393cc3243d22dd0dbcdc/examples/nodejs_prettier/flake.nix)
- provide a CLI utility that discovers projects and dumps them to a `projects.toml` (just a `nix eval` wrapper around `framework.functions.discoverers.discoverProjects`)
- change the basic flake template, to read the projects list from the `projects.toml` file that doesn't exist by default.
- if user initializes our basic flake template and does not specify any project, crash and instruct the user on how to use the new CLI urtility to generate the `projects.toml`.

What do you think about that?
@yusdacra @chaoflow @phaer @hsjobeki

phaer:
I agree that it's probably I good idea to disable auto discovery at by default. 

Personally I wouldn't feel like it's necessary to keep backwards-compatibility as the upgrade should be rather clear and simple, but as I am very new  to this project I might be missing context and/or use cases here.

Whats the rationale behind `projects.toml`? Is it because toml might be friendlier to users new to nix? Otherwise wouldn't a cli which outputs a nix snippet to include in your flake (or default.nix) work just as well?

[...]

PR #415, opened 2022-12-07
Disable automatic project discovery in makeFlakeOutputs

DavHau:
fixes https://github.com/nix-community/dream2nix/issues/407

The diff seems large, but most of it is just modifications to the examples, adding the project.toml etc.
Only the last commit contains changes to the examples.

Changes the following for makeFlakeOutputs:
- adds flag `autoProjects` (bool; default = false)
- crashes if autoProjects is set false and no `projects` specified

Setting `autoProjects = false` restores the old behavior.

If projects are not specified, which will be the default after initializing the basic flake, the following error will be shown:
(`/nix/store/XXX-path-to-source` is replaced with the actual path pointing to the source, so the user can copy paste the command)
```
error: Please pass `projects` to makeFlakeOutputs.
       `projects` can be:
         - an attrset
         - a path to a .toml file (not empty & added to git)
         - a path to a .json file (not empty & added to git)

       To generate a projects.toml file automatically:
         1. execute:
           nix run github:nix-community/dream2nix#detect-projects /nix/store/XXX-path-to-source > projects.toml
         2. review the ./projects.toml and edit it if necessary.
         3. pass `projects = ./projects.toml` to makeFlakeOutputs.

       Alternatively pass `autoProjects = true` to makeFlakeOutputs.
       This is not recommended as it doesn't allow you to review or filter the list
         of detected projects.
```

### DIFF
--- a/examples/_d2n-extended/flake.nix
+++ b/examples/_d2n-extended/flake.nix
@@ -50,12 +50,7 @@
         '')
       ];
       source = src;
-      projects.linemd = {
-        name = "linemd";
-        subsystem = "rust";
-        translator = "cargo-toml-new";
-        builder = "brp-new";
-      };
+      projects = ./projects.toml;
     })
     // {
       # checks.x86_64-linux.linemd = self.packages.x86_64-linux.linemd;
--- a/examples/_d2n-extended/flake.nix
+++ b/examples/_d2n-extended/flake.nix
@@ -50,12 +50,12 @@
         '')
       ];
       source = src;
-      settings = [
-        {
-          builder = "brp-new";
-          translator = "cargo-toml-new";
-        }
-      ];
+      projects.linemd = {
+        name = "linemd";
+        subsystem = "rust";
+        translator = "cargo-toml-new";
+        builder = "brp-new";
+      };
     })
     // {
       # checks.x86_64-linux.linemd = self.packages.x86_64-linux.linemd;
--- a/src/apps/detect-projects/default.nix
+++ b/src/apps/detect-projects/default.nix
@@ -13,9 +13,10 @@ utils.writePureShellScriptBin
 ''
   set -e
 
-  export source=$(realpath .)
+  # Use $1 as source if set. Otherwise default to ./.
+  export source=$(realpath ''${1:-.})
 
   ${apps.callNixWithD2N} eval --json \
-    "dream2nix.framework.functions.discoverers.discoverProjects2 {source = builtins.getEnv \"source\";}" \
+    "dream2nix.functions.discoverers.discoverProjects2 {source = builtins.getEnv \"source\";}" \
     | yq --toml-output
 ''
--- a/src/apps/detect-projects/default.nix
+++ b/src/apps/detect-projects/default.nix
@@ -18,5 +18,6 @@ utils.writePureShellScriptBin
 
   ${apps.callNixWithD2N} eval --json \
     "dream2nix.functions.discoverers.discoverProjects2 {source = builtins.getEnv \"source\";}" \
-    | yq --toml-output
+    | yq --toml-output \
+    | (cat ${./template.toml} && echo "" && cat)
 ''
--- a/src/apps/detect-projects/default.nix
+++ b/src/apps/detect-projects/default.nix
@@ -1,0 +1,21 @@
+{
+  pkgs,
+  utils,
+  apps,
+  ...
+}:
+utils.writePureShellScriptBin
+"detect-projects"
+(with pkgs; [
+  coreutils
+  yq
+])
+''
+  set -e
+
+  export source=$(realpath .)
+
+  ${apps.callNixWithD2N} eval --json \
+    "dream2nix.framework.functions.discoverers.discoverProjects2 {source = builtins.getEnv \"source\";}" \
+    | yq --toml-output
+''
--- a/src/lib.nix
+++ b/src/lib.nix
@@ -182,7 +182,12 @@
             ;
         };
       in
-        allOutputs)
+        framework.dlib.recursiveUpdateUntilDrv
+        allOutputs
+        {
+          apps.detect-projects =
+            dream2nixFor.${system}.flakeApps.detect-projects;
+        })
       allPkgs;
 
     flakifiedOutputsList =
--- a/src/lib.nix
+++ b/src/lib.nix
@@ -75,8 +75,8 @@
     Please pass `projects` to makeFlakeOutputs.
     `projects` can be:
       - an attrset
-      - a path to a .toml file
-      - a path to a .json file
+      - a path to a .toml file (not empty & added to git)
+      - a path to a .json file (not empty & added to git)
 
     To generate a projects.toml file automatically:
       1. execute:
--- a/src/lib.nix
+++ b/src/lib.nix
@@ -195,12 +195,8 @@
       (allOutputs: output: lib.recursiveUpdate allOutputs output)
       {}
       flakifiedOutputsList;
-
-    flakeOutputs =
-      {projectsJson = l.toJSON finalProjects;}
-      // flakeOutputsBuilders;
   in
-    flakeOutputs;
+    flakeOutputsBuilders;
 
   makeFlakeOutputsForIndexes = {
     systems ? [],
--- a/src/lib.nix
+++ b/src/lib.nix
@@ -71,6 +71,24 @@
   }:
     initDream2nix (loadConfig config) pkgs;
 
+  missingProjectsError = ''
+    Please pass `projects` to makeFlakeOutputs.
+    `projects` can be:
+      - an attrset
+      - a path to a .toml file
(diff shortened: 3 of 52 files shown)


examples/_d2n-extended/flake.nix
```diff
@@ -50,12 +50,12 @@
...
-          builder = "brp-new";
-          translator = "cargo-toml-new";
-        }
-      ];
+      projects.linemd = {
+        name = "linemd";
+        subsystem = "rust";
+        translator = "cargo-toml-new";
+        builder = "brp-new";
+      };
```
yusdacra, line 58: This should point to `projects.toml` instead I think, with the translator and builder set correctly.

src/apps/detect-projects/default.nix
```diff
@@ -0,0 +1,23 @@
...
+  set -e
+
+  # Use $1 as source if set. Otherwise default to ./.
+  export source=$(realpath ''${1:-.})
+
+  ${apps.callNixWithD2N} eval --json \
+    "dream2nix.functions.discoverers.discoverProjects2 {source = builtins.getEnv \"source\";}" \
+    | yq --toml-output \
+    | (cat ${./template.toml} && echo "" && cat)
+''
```
yusdacra, line 23: This doesn't need to be addressed right now but I think it could be cool if perhaps this would let users choose translator and builder in an easy fashion so that they wouldn't need to edit `projects.toml` directly. It could present the builders and translators as options and show a small description and link to more extensive documentation in case (it could also be useful to include documentation links in the generated `projects.toml` just so that it is more accessible).

DavHau:
Actually with this, we can now deprecate the weird `settings` field with custom options merging.
I already removed `settings` from the examples. Instead, the settings for individual projects can be put in projects.toml.

src/lib.nix
```diff
@@ -71,6 +71,24 @@
...
+    `projects` can be:
+      - an attrset
+      - a path to a .toml file (not empty & added to git)
+      - a path to a .json file (not empty & added to git)
+
+    To generate a projects.toml file automatically:
+      1. execute:
+        nix run github:nix-community/dream2nix#detect-projects ${source} > projects.toml
+      2. review the ./projects.toml and edit it if necessary.
+      3. pass `projects = ./projects.toml` to makeFlakeOutputs.
```
yusdacra, line 85: I think instead of running `detect-projects` from an external dream2nix we should expose an app called `detect-projects` under outputs. The main dream2nix repo here may not have the discovers / translators / builders etc. a user may add which would cause projects to not get detected etc.

yusdacra (changes requested):

yusdacra:
It looks like all of the examples not use auto discovery now, so we should probably add a test / examples that will test auto discovery? The indexers technically test that already but we only index haskell / rust / nodejs and not the other subsystems so it might still be useful.

src/apps/detect-projects/default.nix
DavHau: Where would you suggest to have these options, if not in the project.toml?
The `project.toml` suits well for editing, why not have it there?

In general I like the idea. We could have commented out options, similar to the generated `configuration.nix` in nixos.
But one significant difference is, that we might output several projects withing one `project.toml`, so I'm not sure if inline explanations of fields will bloat the whole thing too much.
Maybe have a big comment at the beginning of the file, explaining all the options and referencing the documentation?

DavHau:
> It looks like all of the examples not use auto discovery now [...]

Some of them do:
 - examples/_d2n-extended-new-subsystem/flake.nix
 - examples/nodejs_workspace/flake.nix
 - ... also some integration tests do

But I will add one explicit example, to provide one obvious example for the user

src/lib.nix
yusdacra: That doesn't address the issue, because modules are added in the flake itself. So the modules would still be unloaded for that particular instance of dream2nix that would be used. I think adding an app would be fine, since even if there is an issue somewhere else in the flake itself, running `nix run .#detect-projects` will still work regardless because of lazyness.

src/apps/detect-projects/default.nix
yusdacra: I meant options as in a CLI app, when you run it it would let you choose the builder / translator etc. For documentation we can just add it at the start of `projects.toml` as links like `builder: https://nix-community.github.io/dream2nix/subsystems/foo/bar`.

chaoflow:
@DavHau Could we support passing the projects directly as an attribute set, without a separate projects.toml file?

phaer:
> @DavHau Could we support passing the projects directly as an attribute set, without a separate projects.toml file?

I believe [that's already supported](https://github.com/nix-community/dream2nix/blob/d11618fa8fa0a592128e51cc18917272e801fe5c/src/lib.nix#L123) :)

chaoflow:
@DavHau @phaer Great! I think this could be made clearer in the examples, e.g. one of the examples in README.md could use the embedded projects.

src/lib.nix
phaer: > I thought about that, but decided against it, because the flake itself might crash because of some mistake in the toml or flake and then the app won't be available.

What kind of mistakes are we talking about here? 

Malformed TOML files seem to produce relatively specific errors (albeit without the files path ofc because fromTOML doesn't know that), e.g. `error: while parsing a TOML string: [...]`.

Configuration errors are a much more open problem, but not specific to `detect-projects`, unless I am missing something? 

But on the other hand, @yusdacra, wouldn't your use case be covered by using `nix run path-to-local-dream2nix-checkout#detect-projects` or is this a documentation issue?

yusdacra: > wouldn't your use case be covered by using nix run path-to-local-dream2nix-checkout#detect-projects or is this a documentation issue?

No, the issue here is that a user may initialize a dream2nix instance with their own modules, in their `flake.nix` file. This will be specific to *only* that instance, and not any other instance, `nix run dream2nix-checkout#detect-projects` will not be aware of this. And so if new discoverers / translators etc. are added in the user's instance, the `nix run checkout#detect-projects` will not detect the projects that the new discoverers would, because that instance will not have the modules added.

phaer:
Running `nix run .#detect-projects` on dream2nix itself fails for me on `examples/nodejs_workspace`. I'd expect that to yield all example projects? But don't understand the discovery machinery well enough to debug it quickly myself.

<details>
<summary>detailed error message</summary>

```
[dream2nix-devshell]$ nix run .#detect-projects
# To re-generate this file, run:
#   nix run github:nix-community/dream2nix#detect-projects $source
# ... where `$source` points to the source of your project.

error: value is a string while a set was expected

       at /nix/store/8dqk2l8803plqjxkgx32r1g3svsc9r37-dream2nix-full-src/src/subsystems/nodejs/discoverers/discoverer-nodejs/default.nix:38:19:

           37|   getTranslatorNames = tree: let
           38|     packageJson = tree.files."package.json".jsonContent;
             |                   ^
           39|     translators =
[cli]
name = "cli"
relPath = "examples/nodejs_workspace/packages/cli"
subsystem = "nodejs"

[cli.subsystemInfo]
workspaceParent = "examples/nodejs_workspace"
```

</details>

hsjobeki:
>         37|   getTranslatorNames = tree: let
           38|     packageJson = tree.files."package.json".jsonContent;
           
somehow `tree` is a string it seems

src/lib.nix
DavHau: resolved by f83a752d38e95e4748b5b672f19b4ab4bddb8cc8

DavHau:
> @DavHau @phaer Great! I think this could be made clearer in the examples, e.g. one of the examples in README.md could use the embedded projects.

fixed via 011b8ba2382663e315be6ee74e6a8f9811bf73f7



> Running `nix run .#detect-projects` on dream2nix itself fails for me on `examples/nodejs_workspace`. I'd expect that to yield all example projects? But don't understand the discovery machinery well enough to debug it quickly myself.

fixed via c924a9edbccd417c052732bbc5b7d25a17d6872d

src/apps/detect-projects/default.nix
DavHau: > For documentation we can just add it at the start of projects.toml as links like builder: https://nix-community.github.io/dream2nix/subsystems/foo/bar.

This would result in dead links for several subsystems as of now. Maybe we should first ensure that we have a complete option reference generated via our nixos modules?
@antotocar34 Already got us a step further via https://github.com/nix-community/dream2nix/pull/399, but there are still some parts left to complete, like:
https://github.com/nix-community/dream2nix/blob/3d6f13ef9d4cb8c41bc83383bbec3e74865ef90d/src/modules/flake-parts/makeOutputsArgs.nix#L18-L23

DavHau:
Merged. If there is further feedback, I will resolve it in an extra PR. Thanks for your reviews.

DavHau:
Some more fixes here: https://github.com/nix-community/dream2nix/pull/421

PR #421, opened 2022-12-13
fix: makeFlakeOutputs: throw lazily when projects unspecified

DavHau:
Still allow the user to execute the `detect-projects` app

### DIFF
--- a/src/lib.nix
+++ b/src/lib.nix
@@ -201,24 +201,20 @@
       {}
       flakifiedOutputsList;
 
-    errorFlakeOutputs = {
+    errorFlakeOutputs = l.warn (missingProjectsError source) {
       apps =
         l.mapAttrs
         (system: pkgs: {
           detect-projects =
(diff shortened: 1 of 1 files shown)


src/lib.nix
```diff
@@ -200,8 +200,29 @@
...
+        allPkgs;
+    };
+
+    finalOutputs = let
+      givenProjects = args.projects or {};
+    in
+      if
+        (givenProjects == {} && autoProjects == false)
+        || (l.isPath givenProjects && ! l.pathExists givenProjects)
+      then errorFlakeOutputs
```
yusdacra, line 222: I think we should replace the `error` app with `builtins.warn (missingProjectsError source) errorFlakeOutputs` here. It would be more prevalent since it would be showed with every command.

yusdacra (changes requested):

yusdacra (approved):

PR #422, opened 2022-12-13
fix: python: remove spurious --ignore-installed

impl:
This arg was inadvertently added to `pip install` in #397; it breaks transitive dependencies on the packages we custom-install, like setuptools and wheel.

Per @phaer, it wasn't supposed to be there and has already been removed in the next set of larger changes, but this should fix those transitive dependencies in the meantime.


phaer:
Can confirm, shouldn't be there IMO (but can't merge it myself).

Issue #424, opened 2022-12-14
refine flake-parts interface (translators)


DavHau:
Currently the [translator field](https://flake.parts/options/dream2nix.html#opt-perSystem.dream2nix.inputs._name_.projects._name_.translator) accepts any string, and also there is no documentation for the translator options which differ between the individual translators.

I think in order to have nice documentation for the translators, they each need to have their own options like:  
```nix
{
  perSystem.dream2nix.inputs.<name>.projects.<name>.translator = "package-lock";
  perSystem.dream2nix.inputs.<name>.projects.<name>.translatorArgs.package-lock.nodejs = 18;
}
```

While touching the translators, we should probably also:
  - move them to `/src/module` to follow the same schema like our other modules. Currently translators still reside in `/src/subsystems/{subsystem}/translators`
  - after doing the same with builders and discoverers, this will allow us to get rid of our legacy module loading mechanism defined in [src/modules/subsystem-loading.nix](https://github.com/nix-community/dream2nix/blob/59ec60324627ee70108ee882b3ae03ddce44b8a8/src/modules/subsystem-loading.nix)

DavHau:
I started implementing this via https://github.com/nix-community/dream2nix/pull/425. So far I only moved the `package-lock` translator as an example

PR #425, opened 2022-12-14
Integrate translators into /src/modules

DavHau:
solves https://github.com/nix-community/dream2nix/issues/424


DavHau:
@yusdacra In case you have any thoughts on this, please let me know

PR #431, opened 2022-12-20
new nodejs strict builder

hsjobeki:
# Draft: Complete new node builder inspired by the granular builder.

## Features

- __Fully npm compatible (creates node_modules folder ~100% identical to npm)__ 
- __No Patches / Overrides required__ (if "installMethod = copy")
- (Most) Complex building of node_modules is fully implemented as python application, because it requires a lot of control flow.
- Multiple outputs ["out" "lib" "deps"] (explained below) 
- Dedicated flattening of node_modules:
  - Conflicts are resolved during flattening the node_modules folder in favor of the highest semver)
  - Creates node_modules tree directly from package-lock.json informations. (Through the translator)
  - Creates node_modules tree with other lock files (such as yarn-lock) most optimal.
  - TODO: optimize the tree in non-deterministic lock-file formats, where we don't know the structure of the node_modules folder (e.g. yarn-lock)

## Open points:

- [x] Finish off multiple outputs feature
- [x] add devShell
- [ ] node-gyp support - make sure node-gyp works as expected
- [ ] electron - make sure electron apps can be built as expected
- [ ] show how the system can inject dependencies from dream2nix packaged node projects. (Composition without any registry)
- [ ] pull dependencies up into the parent's outer scope (hoisting) if possible to reach the 100% parity with npm's node_modules
 
## Multiple Outputs

 - `.node_modules`: __contains: 'node_modules'__ which can directly be reused e.g. create a devShell.
 - $lib: __package_content, extracted, scripts runned__ TODO: apply the 'files' attribute to filter files like npm would.
 - $out: __common package including /bin folder for usage in other nix expression or nix-shell__
   - $out/bin - includes links to binaries ( -> $lib/src/example-script.js)
   - $out/lib/pkg-content - link to -> $lib
   - $out/lib/node_modules - link to $node_modules
   
## How node_modules are built

- Render the full node_modules structure during eval-time
- Realize the tree structure with a python-builder
- system-incomplatible packages are omitted (e.g. package.json states: os="linux", cpu="x64")
- All dependency binaries are linked to node_modules/.bin
- result can be inspected with 'tree', to find that the output looks almost 100% identical to npm's node_modules

## Under the hood

- Good code quality by formatting, linting and type-checking the python builders.
- Modular phases: Each phase is declared in its own module (nix or python) 
-


DavHau (commented):
Let's add some examples to `/examples` that use the new builder.
Generally more comments for all the variables and functions declared in mkNodeModule and resolveChildren would be nice. 
The diff contains `pyc` files. Once added to the `.gitignore` we should make sure to remove them from the git history. Maybe just squash all commits?

It would be great if you could factor out the re-formatting changes on existing python files and the one on package-lock/default.nix and open a separate PR. We can already merge these right away.

src/subsystems/nodejs/builders/strict-builder/default.nix
```diff
@@ -0,0 +1,341 @@
...
+    produceDerivation,
+    ...
+  }: let
+    l = lib // builtins;
+    b = builtins;
+    inherit (pkgs) stdenv python3 python310Packages makeWrapper jq;
+
+    nodejsVersion = subsystemAttrs.nodejsVersion;
+
+    defaultNodejsVersion = "14";
```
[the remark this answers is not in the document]
hsjobeki: changed to: `lib.versions.major pkgs.nodejs.version`

hsjobeki:
@DavHau I think i am done so far with this PR.
Maybe we could show how to use drv-parts but i think it makes sense to have a seperate PR for it.

docs/src/subsystems/node.md
```diff
@@ -32,3 +40,98 @@ Builds all the dependencies in isolation, moving upwards to the top
...
+> ./projects.toml
+
+```toml
+{{#include ../../../examples/nodejs_alternative_builder/projects.toml}}
+```
+
+#### Override on the `strict-builder`
+
+As currently the builder has three outputs: `$out`, `$lib`, `$deps`.
+When overriding the `installPhase` the user is required to create all three out-paths:
```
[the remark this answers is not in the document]
hsjobeki: Sometimes it may be necessary to override the installPhase to build e.g. static page with `webpack` or others like `react`, `vue`, `anuglar` etc. The cleanest way is then to override the installPhase. To create the output he/she wants.
The user is generally not required to have deeper understandings.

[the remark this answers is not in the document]
hsjobeki: The example:
```
      packageOverrides = {
        "next-app" = {
          builder = {
            installPhase = ''
              runHook preInstall

              next export
              cp -r out $out

              touch $deps
              touch $lib
              
              runHook postInstall
            '';
          };
        };
      };
    };
}
```

builds static page / works for `create next app`

```
johannes@johannes-nixos ~/g/d/next-app (main)> tree result/ -a -L 2
result/
├── 404.html
├── favicon.ico
├── index.html
├── _next
│   ├── 61CZE-ePDegW-aHRoz1tY
│   └── static
├── next.svg
├── nix-support
│   └── propagated-build-inputs
├── thirteen.svg
└── vercel.svg
```


Without that override the installPhase would create other folders and stuff, which is unwanted. (If you have custom builds)

docs/src/subsystems/node.md
```diff
@@ -32,3 +40,98 @@ Builds all the dependencies in isolation, moving upwards to the top
...
+    /nix/store/...-pname-1.0.0-deps
+    ├── .bin
+    ├── @babel
+    ├── ...
+    └── typescript
+```
+
+- package
+- consumable as bare package
+- containing all files from the `source`
```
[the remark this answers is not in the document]
hsjobeki: Currently yes, they are already executed. 
This happens in isolation and could potentially cause problems. I'd like to keep it like it is know.
Later we can allow to break that isolation for specific packages with dedicated override mechanisms. 
-> Switch to run pre- post- /-install scripts in the `top-level` instead of in `isolation`

src/subsystems/nodejs/builders/strict-builder/devShell.nix
```diff
@@ -0,0 +1,27 @@
...
+    in ''
+      # rsync the node_modules folder
+      # is way faster than copying everything again, because it only replaces updated files
+
+      # current Options:
+      # -a -> all files recursive, preserve symlinks, etc.
+      # -c -> calculate hashsums
+      # -E -> preserve executables
+      # --delete -> removes deleted files
+      ${rsync}/bin/rsync -acE --delete ${nodeModulesDir}/* ./node_modules/
```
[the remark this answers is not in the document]
hsjobeki: good idea. But i would say those files do not belong in the root of a project instead all cache / etc informations should live in hidden .dream2nix folder. 

`echo $ID > .dream2nix/node_modules_id`

so we create only one additional entry in the root folder.

Potential other files, specific for dream2nix should also live inside here. (e.g. other cache files)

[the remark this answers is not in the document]
hsjobeki: also the node_modules needs to be re-created if the user deleted them, e.g.  because they where broken.

docs/src/subsystems/node.md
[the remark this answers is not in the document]
DavHau: The user should not need to add build logic.
How about:
```
postBuild = "next export";
```
?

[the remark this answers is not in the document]
DavHau: *the user should not need to copy paste code in order to add build logic

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/config.py
```diff
@@ -0,0 +1,6 @@
+from pathlib import Path
+from .module import env
+
+root = Path("/build")
+node_modules = root / Path("node_modules")
+bin_dir = node_modules / Path(".bin")
```
DavHau, line 6: If there is nothing of interest in `root` other than `node_modules`, then `root` should be removed and instead `node_modules` should default to `/build/node_modules`

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/dependencies.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@dataclass
+class Dependency:
+    name: str
+    version: str
+    derivation: str
+    parent: Union["Dependency", None] = None
+    dependencies: Union[dict[str, Any], None] = None
+
```
DavHau, line 14:
```suggestion
    parent: Optional["Dependency"] = None
    dependencies: Optional[dict[str, Any]] = None

```

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/dependencies.py
```diff
@@ -0,0 +1,135 @@
...
+    def repr(self: "Dependency") -> str:
+        return f"{self.name}@{self.version}"
```
DavHau, line 16:
```suggestion
    def __str__(self: "Dependency") -> str:
        return f"{self.name}@{self.version}"
```

see https://docs.python.org/3/reference/datamodel.html?highlight=__str__#object.__str__

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/dependencies.py
```diff
@@ -0,0 +1,135 @@
...
+    def is_found(acc: Any, dep: Dependency, dep_tree: Optional[DepsTree]) -> bool:
+        return not bool(acc)
+
+    def find_exact_dependency(
+        acc: Any, dep: Dependency, dep_tree: Optional[DepsTree]
+    ) -> Any:
+        if acc:
+            return acc
+
+        if dep.repr() == f"{name}@{version}":
```
DavHau, line 33:
```suggestion
        if str(dep) == f"{name}@{version}":
```

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/dependencies.py
```diff
@@ -0,0 +1,135 @@
...
+    Generic function that traverses the dependency tree and calls
+    'cb' on every node in the tree
+
+    Parameters
+    ----------
+    deps : DepsTree
+        The nested tree of dependencies, that will be iterated through.
+    cb : Callable[[Any, Dependency, Optional[DepsTree]], Any]
+        takes an accumulator (like 'fold' )
+    acc : Any
```
DavHau, line 78: Meaning of very short variable names is hard to grasp when reading the code.

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/dependencies.py
```diff
@@ -0,0 +1,135 @@
...
+    pred: Optional[Callable[[Any, Dependency, Optional[DepsTree]], bool]] = None,
+):
+    """
+    Generic function that traverses the dependency tree and calls
+    'cb' on every node in the tree
+
+    Parameters
+    ----------
+    deps : DepsTree
+        The nested tree of dependencies, that will be iterated through.
```
DavHau, line 75:
```suggestion
        The tree of dependencies, that will be iterated through.
```

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/dependencies.py
```diff
@@ -0,0 +1,135 @@
...
+def get_all_deps(all_deps: dict[str, Any], name: str, version: str) -> list[str]:
+    """
+    Returns all dependencies. as flattened list
+    """
+
+    def is_found(acc: Any, dep: Dependency, dep_tree: Optional[DepsTree]) -> bool:
+        return not bool(acc)
+
+    def find_exact_dependency(
+        acc: Any, dep: Dependency, dep_tree: Optional[DepsTree]
```
DavHau, line 28: whats the meaning of `acc` here?

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/module.py
```diff
@@ -0,0 +1,71 @@
+import os
```
DavHau, line 1: Why is this file named `module.py`? What kind of module is meant here? A description at the top of the file could help or a more meaningful filename.

DavHau (commented):
The functions in the python files don't have any comments. I'd like to see at least a short sentence above each of them explaining what they do / why they are needed.
People who are not deep into the nodejs build system should be able to understand that code.

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/package.py
```diff
@@ -0,0 +1,85 @@
...
+    lifecycle_scripts: list[str] = [
+        "preinstall",
+        "install",
+        "postinstall",
+    ],
+):
+    return (
+        package_json
+        and package_json.get("scripts")
+        and (set(package_json.get("scripts", {}).keys()) & set(lifecycle_scripts))
```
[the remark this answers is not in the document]
hsjobeki: Manually changed it.

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/dependencies.py
hsjobeki: renamed it to `callback`

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/dependencies.py
hsjobeki: renamed to `accumulator`

src/subsystems/nodejs/builders/strict-builder/nodejs_builder/nodejs_builder/lib/config.py
hsjobeki: Refactored: node_modules is not its own derivation.
in that case `root=$out` saves copy time within the derivation.

PR #433, opened 2022-12-24
wip: std skelleton (missing d2n function)

blaggacao:


examples/_d2n-std/nix/app/packages.nix
```diff
@@ -0,0 +1,15 @@
...
+  app = {
+    source = inputs.src;
+    projects = {
+      prettier = {
+        name = "prettier";
+        subsystem = "nodejs";
+        translator = "yarn-lock";
+      };
+    };
+  };
```
blaggacao, line 14: Help, please! :pray: &mdash; This needs to be a `derivation`. How would I do that?

examples/_d2n-std/nix/app/packages.nix
```diff
@@ -0,0 +1,22 @@
+{
+  inputs,
+  cell,
+}: {
+  default = cell.packages.app;
+  app =
+    # a terrible idea
```
blaggacao, line 7: This could work but completely destroys the harmony :smiley: 

But I don't think this is the right way. There must be a better way.

examples/_d2n-std/flake.nix
```diff
@@ -0,0 +1,27 @@
...
+  outputs = {
+    std,
+    self,
+    ...
+  } @ inputs:
+    std.growOn {
+      inherit inputs;
+      cellsFrom = ./nix;
+      cellBlocks = with std.blockTypes; [
+        (installables "packages" {ci.build = true;})
```
blaggacao, line 20:
```suggestion
        # once integration is smooth, we could do
        (dream2nix "packages" {ci.build = true;})
```

examples/_d2n-std/nix/app/packages.nix
DavHau: Currently, we do not have a less terrible interface ;)
I plan to re-design this completely
I recently created a collaborative document to discuss how the new drem2nix API should look like.
Feel free to collaborate and dump your ideas there: https://pad.lassul.us/UswFgE39SDuvRcT3R7irNQ?view

To summarize it. I plan to change dream2nix to not provide any `functions` itself, and instead only export modules for drv-parts. After this, everything will evolve around individual derivations instead of flakes. Composing multiple packages to something like a flake will then be up to the user.

examples/_d2n-std/nix/app/packages.nix
```diff
@@ -0,0 +1,22 @@
...
+        prettier = {
+          name = "prettier";
+          subsystem = "nodejs";
+          translator = "yarn-lock";
+        };
+      };
+    })
+    .packages
+    .${inputs.nixpkgs.system}
+    .default;
```
DavHau, line 21: You might want to pick `prettier` here instead of default. Sometimes the `default` might not point to what you want. The heuristic is a bit poor as of now.

examples/_d2n-std/nix/app/packages.nix
blaggacao: Ok, that is frankly extremely promising.

I engaged in a round of (extensive) comments trying to fully cover my stakeholder perspective as a:

- DevOps persona operating on the intersection of Ops, Build & DevX
- Horizontal Integration Framework maintainer (i.e. `std`)

PR #434, opened 2022-12-25
Add PLTADDONDIR environment variable to racket binaries.

zainab-ali:
The PLTADDONDIR contains [user-specific Racket configuration](https://docs.racket-lang.org/reference/Filesystem.html#%28idx._%28gentag._356._%28lib._scribblings%2Freference%2Freference..scrbl%29%29%29).

For example, is used by `raco` to install and locate a user’s `raco-commands`.

For the current builder, it defaults to the user’s home directory `$TMP_RACO_HOME`. The built `raco` binary, however, defaults to the homeless shelter. This means that the built `raco` can't locate any `raco-commands` in packages installed by the builder.

This commit creates a `PLTADDONDIR` directory structure in `etc` and sets it as an environment variable to all binaries. The paths in `config.rkdt` also point to this directory structure.

There might be a better way to address this - if so, please feel free to point me in the right direction.

### DIFF
--- a/src/subsystems/racket/builders/simple-racket/default.nix
+++ b/src/subsystems/racket/builders/simple-racket/default.nix
@@ -82,24 +82,24 @@
 
           $racket/bin/racket -t ${./make-new-config.rkt}
 
-          export TMP_RACO_HOME=$out/tmp-raco-home
-          mkdir -p $TMP_RACO_HOME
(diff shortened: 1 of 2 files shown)


DavHau:
@leungbk Could you have a look at this?

leungbk (approved):

PR #436, opened 2022-12-27
Add `version` option to the `project` submodule for the flake module

oati:
This seems like it might have been overlooked. I'm currently blocked on trying this API out with the `hackage` haskell translator because this option is missing.


oati:
I renamed the branch and github closed the PR, despite github docs promising otherwise.

Issue #437, opened 2022-12-27
Haskell builder does not seem to resolve the correct dependencies


oati:
I'm trying to build purescript for `aarch64-linux` on apple M1. Here's what I have so far:

```nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
    flake-parts.url = "github:hercules-ci/flake-parts";

    dream2nix = {
      # url = "github:nix-community/dream2nix";
      # using a patch for the project.version option
      url = "github:oati/dream2nix/patch-1";
      inputs.nixpkgs.follows = "nixpkgs";
    };

    purescript = {
      # url = "github:purescript/purescript/v0.15.7";
      url = "https://hackage.haskell.org/package/purescript-0.15.7.tar.gz";
      flake = false;
    };
  };

  outputs = {
    self,
    nixpkgs,
    flake-parts,
    dream2nix,
    purescript,
  }@inputs:
    flake-parts.lib.mkFlake { inherit inputs; } {
      systems = [ "x86_64-linux" "aarch64-linux" ];
      imports = [ dream2nix.flakeModuleBeta ];

      perSystem = { pkgs, ... }: {
        dream2nix.inputs.purescript = {

          source = purescript;

          projects = {
            purescript = {
              name = "purescript";
              version = "0.15.7";
              subsystem = "haskell";
              translator = "hackage";
            };
          };

          packageOverrides = {
            zlib = {
              add-dep = {
                buildInputs = old: old ++ [ pkgs.zlib ];
              };
            };
          };
        };
      };
    };
}
```

Trying to build the purescript derivation results in an error:

```
Warning:
    This package indirectly depends on multiple versions of the same package. This is very likely to cause a compile failure.
      package purescript (purescript-0.15.7) requires process-1.6.13.1-1hIWXPWmnzXIcFtLftwLRN
      package purescript (purescript-0.15.7) requires process-1.6.13.1-1hIWXPWmnzXIcFtLftwLRN
      package purescript (purescript-0.15.7) requires process-1.6.13.1-1hIWXPWmnzXIcFtLftwLRN
      package typed-process (typed-process-0.2.10.1-CkicmSu9xcwAQXLxFp11GR) requires process-1.6.13.2
      package streaming-commons (streaming-commons-0.2.2.5-2PxHkvTZbWKB0N81TiQ1wx) requires process-1.6.13.2
[...]

PR #439, opened 2022-12-29
haskell builder: add a `--constraint` to `configureFlags` for each dependency

oati:
This is one solution for #437.

We add a constraint to `configureFlags` for each dependency with the version specified in our dream-lock.

This fixes the issue, but it's a bit hacky. I don't know enough about the haskell toolchain to know whether there's a better solution.

This will probably at least be a good stop-gap until there's a better solution.


DavHau:
Thanks a lot. Just the formatting needs to be fixed.

oati:
I didn't realize there was a check for formatting. I've updated everything using alejandra.

PR #443, opened 2022-12-29
haskell: add impure cabal-plan translator based on hackage translator

oati:
The translator proposed here is basically an "impure wrapper" for the cabal-plan translator. It's a more general version of the hackage translator--all that's changed is that the source is pulled from the `source` option, instead of being downloaded from hackage using `name` and `version`.

The translator copies the source into a temporary directory, runs `cabal freeze`, runs the cabal-plan translator, and then patches in the original source (just like the hackage translator).

### DIFF
--- a/examples/haskell_cabal/flake.nix
+++ b/examples/haskell_cabal/flake.nix
@@ -1,0 +1,24 @@
+{
+  inputs = {
+    dream2nix.url = "github:nix-community/dream2nix";
+    src.url = "github:davhau/cabal2json/plan-json";
+    src.flake = false;
+  };
+
+  outputs = {
+    self,
+    dream2nix,
+    src,
+  } @ inp:
+    (dream2nix.lib.makeFlakeOutputs {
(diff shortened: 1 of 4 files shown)


oati:
I'm not familiar with how the tests work, but I've added an example identical to the cabal-plan example, with only the translator changed.

oati:
I'd edited the wrong `projects.toml`.

oati:
(I'm sorry for the small mistakes.. I'm not able to run the tests locally, since the examples assume an `x86_64-linux` system)

oati:
Ideally I think the example should be a project that doesn't keep lockfiles in its repo.

DavHau:
I opened an issue for the tests not being cross platform. Should be an easy fix: https://github.com/nix-community/dream2nix/issues/448

DavHau:
I poked around a bit and the problem was that the name declared in the `project.toml` was different from the detected package name. Therefore the generate dream-lock.json was ignored. fixed via https://github.com/nix-community/dream2nix/pull/443/commits/073357486cc374df60cfb8123634a92b93c70a21

PR #444, opened 2022-12-29
haskell: add zlib override

oati:
Add `pkgs.zlib` to haskell's `zlib` `buildInputs`.

### DIFF
--- a/overrides/haskell/default.nix
+++ b/overrides/haskell/default.nix
@@ -1,0 +1,7 @@
+{pkgs, ...}: {
+  zlib = {
+    add-deps = {
+      buildInputs = old: old ++ [pkgs.zlib];
+    };
+  };
+}


DavHau:
Thanks a lot.
On another branch I already extracted the `pkg-config -> nixpkgs` mapping from cabal2nix and [dumped it as a json](https://github.com/nix-community/dream2nix/commit/afc1eed8dbaef4a1b64ccd5aa6c4a21759b2cdd3).

The `all-cabal-json` data set which is already used in dream2nix contains the pkg-config fields. We just need to read those in the translator and run it through the mapping.
If you would have some time finalizing this, I'd highly appreciate it.

Thanks for you work so far.

DavHau:
No thinking about it, the approach I suggested is unnecessarily complex. We should probably just generate overrides from the json. No need to add complexity in the translators.

PR #452, opened 2023-01-11
flake-parts: set default project name based on attribute name

max-privatevoid:
TL;DR it makes [this](https://github.com/nix-community/dream2nix/blob/9f6911c78dcb0832f7fcc955e847db1a5a9ce29a/examples/_d2n-flake-parts/flake.nix#L27) line unnecessary.


DavHau:
thanks

PR #454, opened 2023-01-13
Support git url dependencies in Racket dream2nix generation

zainab-ali:
The racket `dream-lock.json` generation currently only supports dependencies found in the Racket package catalog. `raco` has [several other mechanisms for resolving dependencies](https://docs.racket-lang.org/pkg/Package_Concepts.html#%28tech._package._name%29), one of which is through a git url.

As an example, the following `info.rkt`, when installed manually using `raco pkg install`, pulls down the `shellac` repository:

```
#lang info

(define name "foobar")
(define deps '("base"
              "git+https://git.sr.ht/~zainab/shellac?ref=master"))
```

This PR extends the impure dream2nix.json generation to support the `git-url` package type.

Given the size of the changeset, I have split the changes into refactoring commits and logic changes. The core logic change is in the last commit. I know that this PR is a large one. If you're daunted by the changeset, please feel free to discuss it and point me in the right direction.

# How it works
We first test whether any dependencies have the `git-url` package type. If they do, we use the `pkg-stage` function to checkout the repository in a temporary directory and read its `info.rkt`. Any dependencies found in that `info.rkt` are recursively and added to the graph.

# Logging
I also added a [logger](https://docs.racket-lang.org/reference/logging.html#%28idx._%28gentag._426._%28lib._scribblings%2Freference%2Freference..scrbl%29%29%29) for troubleshooting purposes. You can activate it via the `PLTSTDERR` environment variable (e.g. `export PLTSTDERR=info`). Let me know what you think about it.

### DIFF
--- a/src/subsystems/racket/translators/racket-impure/generate-dream-lock.rkt
+++ b/src/subsystems/racket/translators/racket-impure/generate-dream-lock.rkt
@@ -34,6 +34,29 @@
             (with-handlers ([exn:fail? ignore-error])
               (info-procedure 'build-deps)))))
 
+(define (remote-pkg->source name url rev)
+
+  (define (url->source url)
+    (let* ([source-with-removed-http-or-git-double-slash (regexp-replace #rx"^(?:git|http)://" url "https://")]
+           [left-trimmed-source (string-trim source-with-removed-http-or-git-double-slash "git+" #:right? #f)]
+           [maybe-match-path (regexp-match #rx"\\?path=([^#]+)" left-trimmed-source)]
+           [trimmed-source (regexp-replace #rx"(?:/tree/.+)?(?:\\?path=.+)?$" left-trimmed-source "")])
+      (cons `(url . ,trimmed-source)
+            (match maybe-match-path
+              [(list _match dir)
+               `((dir . ,(regexp-replace* #rx"%2F" dir "/")))]
+              [_ '()]))))
+
+  (cons (string->symbol name)
+        (make-immutable-hash
+         `((0.0.0 . ,(make-immutable-hash
+                      (append
+                       (url->source url)
+                       `((rev . ,rev)
+                         (type . "git")
+                         ;; TODO: sha256?
+                         ))))))))
+
 (define (generate-dream-lock pkgs-all-path)
   (let* ([src-path (getenv "RACKET_SOURCE")]
          [rel-path (getenv "RACKET_RELPATH")]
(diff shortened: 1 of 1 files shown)


DavHau:
Thanks for that!
@leungbk Could you review this?

leungbk:
LGTM.

In the repo used to curate the catalog, I initially decided to [omit this kind of dependency](https://github.com/nix-community/pruned-racket-catalog/blob/973fd725fbf3c98474fd2d8ebae45e6c8c2f36e0/write-catalog.rkt#L129-L138) because:

0. In general, there can exist packages that are distributed without publishing VCS info to the Racket package catalog, such as https://neilvandyke.org/racket/. I noticed that he typically distributed his stuff as `.zip`s. This makes it hard for us to produce a dream-lock that guarantees that we can use a particular revision of a package. For example, if Neil van Dyke were to update his Levenshtein package (whose source URL in the Racket package catalog is http://www.neilvandyke.org/racket/levenshtein.zip) to version 2.0, when previously I had a flake trying to use some version 1.0, then the Racket package catalog cannot provide enough information to acquire version 1.0. This led me to require that the catalog used by dream2nix [would not admit any package missing VCS info](https://github.com/nix-community/pruned-racket-catalog/blob/973fd725fbf3c98474fd2d8ebae45e6c8c2f36e0/write-catalog.rkt#L115), or any package containing such a dependency in its transitive closure.
1. Because Git URL dependencies are not going to be in the Racket package catalog, we cannot know without parsing their `info.rkt`s for ourselves (something I did not want to do in the catalog repo) whether or not they contain any `.zip` dependencies. That led me to also reject those (+ reverse dependencies thereof) in our package catalog.

I think the current PR can be merged as it is. If you think it's also a good idea to permit packages with Git-URL dependencies in the catalog, you may want to submit a PR supporting that in the catalog repo. I'm not knowledgeable about Racket in general, plus I don't use the language very much, so I don't have a strong opinion about that.

src/subsystems/racket/translators/racket-impure/generate-dream-lock.rkt
```diff
@@ -16,134 +19,204 @@
...
+  (match-lambda [(or (cons pkg-name _) pkg-name) (package-source->name+type pkg-name #f)]))
+
+(define (dependency->name dep)
+  (let-values ([(name _) (dependency->name+type dep)]) name))
+
+(define (fetch-git-repositories dep graph)
+
+  (define (package-source->git-url dep)
+  (let-values ([(_ type) (dependency->name+type dep)])
+    (and (eq? type 'git-url) dep)))
```
leungbk, line 51: Indentation is off here.

zainab-ali: Thanks! I've corrected it.

zainab-ali:
Thanks for the detailed explanation!  I'm a reasonably heavy racket user for personal projects, and am in the process of nixifying more and more builds.

I haven’t yet had a problem with the catalog nix expression being noncomprehensive (all the packages I use have had VCS info), but I can see how it would be problematic. A user ho wants to depend on `levenstein` wouldn’t have a clear idea as to why they couldn’t, and would have no alternative but to patch the generated `dream-lock.json` (if the generation even succeeded).

I don’t think it’s necessary to change the logic in  `pruned-racket-catalog`, but it might be worth adding more visibility as to what was pruned. I’ll have a look into it when I next need to touch it.

PR #455, opened 2023-01-14
Waalge/bump docs

waalge:
Pointers welcome. 

mdbook build failed cos a link had fallen out of sync. 
This is fixed. I also tried to add a pre-commit hook for mdbook to prevent this, but i didn't find a good way of doing it, so I did a bad way. 

The way the guides are factored out didn't make it obvious where to add the line on `projects.toml`. 
I dumped it in nix_system which seems wrong but at least appears on both guides.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -305,7 +305,7 @@
                 entry = l.toString (pkgs.writeScript "build-docs" ''
                   #!${pkgs.bash}/bin/bash
                   errors=$(mdbook build docs/ 2>&1 | grep ERROR) # no better way?!
-                  if [ $errors ]; then
+                  if [ "$errors" ]; then
                     exit 1
                   fi
                 '');
--- a/flake.nix
+++ b/flake.nix
@@ -299,6 +299,17 @@
                   fi
                 '');
               };
+              build-docs = {
+                enable = true;
+                name = "build-docs";
+                entry = l.toString (pkgs.writeScript "build-docs" ''
+                  #!${pkgs.bash}/bin/bash
+                  errors=$(mdbook build docs/ 2>&1 | grep ERROR) # no better way?!
+                  if [ $errors ]; then
+                    exit 1
+                  fi
+                '');
+              };
             };
           };
         };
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -299,6 +299,17 @@
...
+                enable = true;
+                name = "build-docs";
+                entry = l.toString (pkgs.writeScript "build-docs" ''
+                  #!${pkgs.bash}/bin/bash
+                  errors=$(mdbook build docs/ 2>&1 | grep ERROR) # no better way?!
+                  if [ "$errors" ]; then
+                    exit 1
+                  fi
+                '');
+              };
```
DavHau, line 312: I think this should go to the .#docs derivation itself. It is already tested in CI.
Here: https://github.com/nix-community/dream2nix/blob/f11fa0fcedb1a966e86904d772c1574f36f3dfc9/flake.nix#L311

Maybe you could use `tee` to capture the output in a file while still having it sent to stdout.

DavHau (commented):
Putting it into the define-targets.md seems fine.

flake.nix
waalge: Might need you to unpack this more. 

I think you want: 
- `nix build .#docs` to produce  a log file. (Where to put this file?) When I run this I don't see any output, it seems to be suppressed.
- The log file to be read for errors and ... throw an error if it finds one?
- the hook to test the derivation ... ie just execute it. 

How far off the mark is that?

DavHau: What I meant is, don't add a pre commit hook. Instead put the check inside the 'docs' derivation that already exists.

waalge: I'm lacking context here to understand what your asking for. I don't think I've done what you want. 
By "check", do you mean something that is run by "nix flake check". 
I don't know when someone would find out that their edit has broken the docs without a hook

DavHau: Exactly like you did it in https://github.com/nix-community/dream2nix/pull/455/commits/e4dedcde75e383c56e4940d307edaa0a6e56c1d9. Thanks!

PR #456, opened 2023-01-18
Docs: Better highlighting of nix code in docs

hsjobeki:
# Things done

- Add better highlight.js `nix` stylesheet
- Add dark-mode highlighting
- Turn on dark-mode in `book.toml`

## Details 

Currently we use highlight.js@10.1.2 where mdBook applies its custom theme.
All in all looks not so great. Meanwhile the community released highlight.js@11.7.0 which doesnt work properly with mdBook anymore.

However the new (11.7.0) .css stylesheets do work backwards with our old highlight.js version.  

So i propose the `github` and `github-dark` styling as new theme for nix code highlighting. 

As it has gotten way better syntax support in the meanwhile.

- [x] Light mode
- [x] Dark mode

### DIFF
--- a/docs/book.toml
+++ b/docs/book.toml
@@ -4,3 +4,8 @@ language = "en"
 multilingual = false
 src = "src"
 title = "dream2nix documentation"
+
+[output.html]
+default-theme = "light"
+preferred-dark-theme = "navy"
+additional-css = ["theme/highlight-dark.css"]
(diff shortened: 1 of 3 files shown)


DavHau:
Thanks for that but as of now I find the old highlighting better to understand, as it uses multiple different colors, vs the new one seems to use only different shades of blue. The contrast between different highlights is low which makes it harder to grasp for me.

PR #465, opened 2023-01-31
flake-parts: add builder option + don't expose packages automatically

max-privatevoid:
This PR does 2 things:
- Adds an optional `builder` option to the flake-parts module to match the standard dream2nix interface
- Removes the module's behavior of automatically exposing all packages in the `perSystem.packages` set (same for devShells)
  - The previous logic was too strict as it resulted in running the entire translator logic just to evaluate `attrNames packages.${system}`. This is particularly problematic for larger flakes, where dream2nix and other packages are exposed together.

### DIFF
--- a/examples/_d2n-flake-parts/flake.nix
+++ b/examples/_d2n-flake-parts/flake.nix
@@ -29,6 +29,9 @@
             };
           };
         };
+        packages = {
+          inherit (config.dream2nix.outputs.prettier.packages) prettier;
+        };
       };
     };
 }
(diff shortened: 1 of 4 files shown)


phaer (approved):
Looks good to me, thanks! 

Note that there's a second PR open regarding the packages output in the flake.parts interface, which seems to be in conflict with this one? I'd personally prefer your approach of removing the default output over just merging them, due to the evaluation issue you've mentioned.


https://github.com/nix-community/dream2nix/pull/465

Issue #464, opened 2023-01-29
/examples/python_pip


InLaw:
setup:
```
 - system: `"x86_64-linux"` 
 - host os: `Linux 5.15.88, NixOS, 22.11 (Raccoon), 22.11.1705.b83e7f5a04a` 
 - multi-user?: `yes` 
 - sandbox: `yes` 
 - version: `nix-env (Nix) 2.11.1` 
 - channels(root): `"nixos-22.11, nixos-hardware"` 
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

--

https://github.com/nix-community/dream2nix/tree/main/examples/python_pip

following 
https://nix-community.github.io/dream2nix/guides/getting-started-python.html

examples/python_pip$ nix build .#default
error: flake 'path:/home//python_pip' does not provide attribute 'packages.x86_64-linux.default', 'legacyPackages.x86_64-linux.default' or 'default'

results in 
`examples/python_pip$ nix build .#default`
`error: flake 'path:/home//python_pip' does not provide attribute 'packages.x86_64-linux.default', 'legacyPackages.x86_64-linux.default' or 'default'`

phaer:
`'path:/home//python_pip'` this doesn't look like the right path. The `.` in `.#` represents the current directory, so what does `pwd` output for you? Is your copy of the example part of a git repo? What does `nix flake show` say, do you see the packages there?

InLaw:
yes, it is the example from the repo 
analogous like in https://github.com/nix-community/dream2nix/issues/429

currently 
```
cd python_pip/
direnv: loading ~/../examples/python_pip/.envrc
direnv: using flake
[1/0/1 copied (32.0/123.4 MiB), 5.8/20.9 MiB DL] fetching source from https://cache.nixos.orgdirenv: ([/nix/store/nrj92sd5xv0lym6x6z1ih0ndnlb7wc37-direnv-2.32.1/bin/direnv export bash]) is taking a while to execute. Use CTRL-C to give up.
```
takes > 3 min to load ?!

nix flake show
```
path:/home/../examples/python_pip?lastModified=1675173658&narHash=sha256-ndBXMuP%2fjFC68GhXtiMSMrJHXnSmZCpXFADKNW8wqzU=
├───apps
│   └───x86_64-linux
│       └───detect-projects: app
├───checks
│   └───x86_64-linux
│       ├───aiohttp: derivation 'aiohttp'
[...]

PR #466, opened 2023-02-01
Update getting-started-python.md

phaer:
.#default isn't defined in new flakes, see https://github.com/nix-community/dream2nix/issues/464


phaer:
CI formatter change in dream2nix/src/subsystems/haskell/translators/cabal/fixup-dream-lock.py seems unrelated to this PR?

nhooey:
Both the Python _and_ NodeJS instructions should mention running `nix run .#resolveImpure` before `nix build`, otherwise both will error and mention a lack of targets for the current system.

phaer:
@nhooey resolveImpure is mentioned later in the tutorial

PR #468, opened 2023-02-06
feat(v1 API): document v1 API examples

DavHau:
v1 API does not exist yet.
I want to make sure that all relevant use cases are sketched out before starting with the implementation.

### DIFF
--- a/docs/src/v1-api/consuming/override.md
+++ b/docs/src/v1-api/consuming/override.md
@@ -1,0 +1,198 @@
+# Consume and modify dream2nix packages
+
+## Given the following package
+`upstream/my-package.nix`
+```nix
+{config, lib, dream2nix, ...}: {
+
+  imports = [
+    dream2nix.modules.nodejs.mkDerivation
+    dream2nix.modules.nodejs.package-lock
+  ];
+
+  pname = "my-package";
+  version = "2.0.0";
+
+  src = {
+    type = github;
+    owner = "my-user";
+    repo = "my-repo";
+    ref = config.version;
+    hash = "sha256-mia90VYv/YTdWNhKpvwvFW9RfbXZJSWhJ+yva6EnLE8=";
+  };
+
+  # declare dependency on python3
+  deps = {nixpkgs, ...}: {
+    python3 = nixpkgs.python39;
+  };
+
+  nativeBuildInputs = [
+    config.deps.python3
+  ];
+
+  configurePhase = ''
+    python3 --version
+  '';
+
+  buildPhase = ''
+    python3 -c 'print("Hello World!")' > $out
+  '';
+}
+```
+
+`upstream/default.nix`
+```nix
+{
+  nixpkgs ? import <nixpkgs> {},
+  dream2nix ?
+    import
+    (builtins.fetchTarball "https://dream2nix.dev/tarball/1.0")
+    {inherit nixpkgs;},
+}: {
+  packages.my-package = dream2nix.eval ./my-package.nix;
+}
+```
+
+## 1. Override using modules
+
+### 1.1 Define a module for the override
+`my-package-override.nix`
+```nix
+{config, lib, ... }: {
+
+  version = "2.1.0";
+
+  # No need to re-define other fetcher attributes.
+  # The module system updates them for us.
+  src.hash = "sha256-LM5GDNjLcmgZVQEeANWAOO09KppwGaYEzJBjYmuSwys=";
+
+  deps = {nixpkgs, ...}: {
+
+    # change the python version
+    python3 = lib.mkForce nixpkgs.python310;
+
+    # add a dependency on hello  
+    hello = nixpkgs.hello;
+  };
+
+  # add hello to nativeBuildInputs
+  # (`oldAttrs.nativeBuildInputs + ...` not needed here)
+  nativeBuildInputs = [
+    config.deps.hello
+  ];
+
+  # add lines to configurePhase
+  postConfigure = ''
+    hello --version
+  '';
+
+  # replace the build phase via mkForce
+  buildPhase = lib.mkForce "
+    hello > $out
+  ";
+}
+```
+
+### 1.2 Apply `my-package-override.nix` via extendModules
+Using `extendModules` is simple.
+It allows to extend an existing package with another module.
+This doesn't require knowledge about the original modules that went into the package.
+
+`./default.nix`
+```nix
+let
+  nixpkgs = import <nixpkgs> {};
+  upstream = import ./upstream {inherit nixpkgs;};
+  my-package = upstream.packages.my-package;
+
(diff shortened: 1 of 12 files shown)


docs/src/v1-api/consuming/override.md
```diff
@@ -0,0 +1,198 @@
...
+  imports = [
+    dream2nix.modules.nodejs.mkDerivation
+    dream2nix.modules.nodejs.package-lock
+  ];
+
+  pname = "my-package";
+  version = "2.0.0";
+
+  src = {
+    type = github;
```
zimbatm, line 17: The value here would a string isn't it?

docs/src/v1-api/consuming/override.md
```diff
@@ -0,0 +1,198 @@
...
+`upstream/my-package.nix`
+```nix
+{config, lib, dream2nix, ...}: {
+
+  imports = [
+    dream2nix.modules.nodejs.mkDerivation
+    dream2nix.modules.nodejs.package-lock
+  ];
+
+  pname = "my-package";
```
zimbatm, line 13: `pname` is one of those historical accidents. Maybe it's a good opportunity to rename it to `name`.

docs/src/v1-api/consuming/override.md
```diff
@@ -0,0 +1,198 @@
...
+  version = "2.0.0";
+
+  src = {
+    type = github;
+    owner = "my-user";
+    repo = "my-repo";
+    ref = config.version;
+    hash = "sha256-mia90VYv/YTdWNhKpvwvFW9RfbXZJSWhJ+yva6EnLE8=";
+  };
+
```
zimbatm, line 23: Same fetch API with patches I guess, except that it would take an array.

docs/src/v1-api/consuming/override.md
```diff
@@ -0,0 +1,198 @@
...
+```nix
+{config, lib, ... }: {
+
+  version = "2.1.0";
+
+  # No need to re-define other fetcher attributes.
+  # The module system updates them for us.
+  src.hash = "sha256-LM5GDNjLcmgZVQEeANWAOO09KppwGaYEzJBjYmuSwys=";
+
+  deps = {nixpkgs, ...}: {
```
zimbatm, line 69: it's not clear why deps take another function. couldn't nixpkgs be passed to the parent function?

zimbatm (commented):
I didn't finish reading the whole doc yet so it will require another pass, but I'm out of time for today

docs/src/v1-api/consuming/override.md
DavHau: I considered that, but am not sure if this would be a good idea as of now. We still use `nixpkgs.mkDerivation` as a builder for most derivations. If we put an adapter in front of it that changes the semantics we also have to ensure that this change is reflected by all downward compatibility mechanisms. For example we need to patch `overrideAttrs` to have that behavior as well.

For keeping things simple we might want to keep drv-parts based modules as close as possible to being a drop in replacement for the underlying builder API.

docs/src/v1-api/consuming/override.md
DavHau: > it's not clear why deps take another function

`deps` is our equivalent of package function args in nixpkgs. A single place to override dependencies.

> couldn't nixpkgs be passed to the parent function?

We could pass nixpkgs via `specialArgs` but that would incurage usage like:
```
{config, nixpkgs, ...}: {
  buildInputs = [nixpkgs.python3];
  nativeBuildInputs = [nixpkgs.python3];
  preConfigure = ''
    ${nixpkgs.python3}/bin/python --version
  '';
}
```
This would make it hard to override `python3`, as there isn't a single source of truth.

Instead we want to limit the scope of `packageSets` like `nixpkgs` to a single attribute `deps` which makes overriding simple.

I should add some comments to the code to explain more about it.

Another cool thing is that`deps` is a submodule, so it allows us to specify options:
```nix
{config, lib, ...}: {
  options.deps.python3 = lib.mkOption {
    description = "Select a python3 version for building this package";
    # ...
  };
}
```

... which allows us to generate a manual for each individual package.

docs/src/v1-api/problems.md
```diff
@@ -0,0 +1,22 @@
...
+## Integration of existing lang2nix tools
+Until now, integrating existing 2nix solutions into dream2nix was hard because dream2nix imposed standards which are not met by most existing tools.
+With v1 we want to lift most of these restrictions to make integration a no-brainer
+
+(see sections [integrate lang2nix tool (pure)](../v1-api/integrate-lang2nix-pure.md) and [integrate lang2ix tool (impure)](../v1-api/integrate-lang2nix-impure.md).
+
+## Tied to flakes
+The current api is tied to flakes. The v1 API should not depend on flakes anymore.
+
+## Composability
```
tomberek, line 12: This is the rationale in encouraging the use of blueprints/package-functions/proto-derivations (naming in-progress). The goal would be to make flakes easier to compose, or to compose in different ways.

docs/src/v1-api/packaging/nodejs-init-project.md
```diff
@@ -0,0 +1,123 @@
...
+## create `my-app-shell.nix` for your dev shell
+`my-app-shell.nix`
+```nix
+{config, lib, dream2nix, ...}: {
+
+  imports = [
+    # the default dev shell for nodejs
+    dream2nix.modules.nodejs.mkShell
+    # adds dependencies of my-app to the dev shell
+    dream2nix.modules.nodejs.package-lock
```
bb010g, line 44: Why are these magic imports you have to replicate in each submodule instead of enable options provided by modules that you can just import once at a higher level (probably `default.nix` here)?

DavHau: My intention was to keep evaluation overhead minimal. I wanted to avoid the problem of always having to import all modules, like seen in nixos as it appears to be limiting scalability and slow down evaluation.

I'm not too sure about my judgement here. Maybe it's not worth while and we should switch back to just using `.enable = ...`.

phaer: I personally like importing the modules, because besides possible evaluation overhead IMO it can make it clearer which options are in scope.

> instead of enable options provided by modules that you can just import once at a higher level (probably default.nix here)

Just to avoid misunderstandings: Those imports also affect all modules which end up the same config (or derivation in the context of drv-parts). So importing them at a higher level would also work. An argument for .enable could be made for the inverse, i think: .enable would allow us to then disable again in specific cases via `enable = False` which isn't as easy for imports.

docs/src/v1-api/consuming/override.md
DavHau: This was resolved in the mean time. There is no pname anymore.

PR #481, opened 2023-03-02
feat(rust): prepare for duplicate dependency vendoring

yusdacra:
fixes issues when duplicate dependencies are encountered
errors out in vendoring phase if duplicate versions are encountered (cargo vendor still does not support this behaviour)

### DIFF
--- a/src/subsystems/rust/translators/cargo-lock/default.nix
+++ b/src/subsystems/rust/translators/cargo-lock/default.nix
@@ -97,17 +97,58 @@ in {
     parsedLock = projectTree.files."Cargo.lock".tomlContent;
     parsedDeps = parsedLock.package;
 
+    makeDepId = dep: "${dep.name} ${dep.version} (${dep.source or ""})";
+
     # Gets a checksum from the [metadata] table of the lockfile
-    getChecksum = dep: let
-      key = "checksum ${dep.name} ${dep.version} (${dep.source})";
-    in
-      parsedLock.metadata."${key}";
+    getChecksum = dep: parsedLock.metadata."checksum ${makeDepId dep}";
 
     # map of dependency names to versions
     depNamesToVersions =
-      l.listToAttrs
-      (l.map (dep: l.nameValuePair dep.name dep.version) parsedDeps);
-    # This parses a "package-name version" entry in the "dependencies"
+      l.foldl'
+      (
+        all: el:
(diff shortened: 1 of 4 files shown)


src/subsystems/rust/translators/cargo-lock/default.nix
```diff
@@ -97,17 +97,58 @@ in {
...
+        all: el:
+          if l.hasAttr el.name all
+          then all // {${el.name} = all.${el.name} ++ [el];}
+          else all // {${el.name} = [el];}
+      )
+      {}
+      parsedDeps;
+    findUniqDep = let
+      # tests whether if el contains all elements from dep
+      compareAttrsets = dep: el:
```
DavHau, line 118: I think this function could be easier to understand if it was split into several ones. The nesting and inline function definitions make it a bit hard to review.

src/subsystems/rust/translators/cargo-lock/default.nix
```diff
@@ -97,17 +97,58 @@ in {
...
+      (
+        all: el:
+          if l.hasAttr el.name all
+          then all // {${el.name} = all.${el.name} ++ [el];}
+          else all // {${el.name} = [el];}
+      )
+      {}
+      parsedDeps;
+    findUniqDep = let
+      # tests whether if el contains all elements from dep
```
DavHau, line 117: What is `el` and what kind of `elements` are we talking about here? Maybe give more meaningful names?

src/subsystems/rust/translators/cargo-lock/default.nix
yusdacra: I don't think it makes sense to split this function into more pieces, everything there is pretty much only needed for what that function does, i did add more comments though

src/subsystems/rust/translators/cargo-lock/default.nix
yusdacra: Done, added more comments

src/subsystems/rust/translators/cargo-lock/default.nix
DavHau: Thanks, the comments improve the readability a lot. Let's keep up the commenting like that. This will make the project more appealing to outside contributors.
There are some failing unit tests now. If you fix these, I think we are good to go.

PR #482, opened 2023-03-03
Support old PHP versions

phfroidmont:
As suggested in #240, adds support for all stable php versions from https://github.com/fossar/nix-phps

### DIFF
--- a/src/subsystems/php/builders/granular-php/default.nix
+++ b/src/subsystems/php/builders/granular-php/default.nix
@@ -2,6 +2,7 @@
   lib,
   pkgs,
   externals,
+  inputs,
   ...
 }: {
   type = "pure";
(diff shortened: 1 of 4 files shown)


DavHau:
Thanks for that.
According to the readme of `nix-phps`: There is also a php package which is the alias of the default PHP version in Nixpkgs.
Can we ensure that the nixpkgs php remains the preferred choice over the other php packages?

@tinybeachthor pinging you in case you'd like tor review this.

src/subsystems/php/builders/granular-php/default.nix
```diff
@@ -36,26 +37,30 @@
 
     inherit (pkgs.callPackage ../../semver.nix {}) satisfies;
 
-    # php with required extensions
-    php =
-      if satisfies pkgs.php81.version subsystemAttrs.phpSemver
+    phpPackages = l.attrValues inputs.phps.packages.x86_64-linux;
```
DavHau, line 40:
```suggestion
    phpPackages = l.attrValues inputs.phps.packages.x86_64-linux;
    phpPackages = l.attrValues (import inputs.phps).packages.x86_64-linux;
```

src/subsystems/php/builders/simple-php/default.nix
```diff
@@ -36,26 +37,30 @@
 
     inherit (pkgs.callPackage ../../semver.nix {}) satisfies;
 
-    # php with required extensions
-    php =
-      if satisfies pkgs.php81.version subsystemAttrs.phpSemver
+    phpPackages = l.attrValues inputs.phps.packages.x86_64-linux;
```
DavHau, line 40:
```suggestion
    phpPackages = l.attrValues inputs.phps.packages.x86_64-linux;
    phpPackages = l.attrValues (import inputs.phps).packages.x86_64-linux;
```

src/subsystems/php/builders/granular-php/default.nix
tinybeachthor, line 40: Probably does not make much difference, maybe passing `system` attribute here instead of hard-coding `x86_64-linux`.

phfroidmont: For my own understanding, what does import do here?

phfroidmont:
`php` being [first on the list](https://github.com/fossar/nix-phps/blob/master/flake.nix#L32) it will be the first considered if I'm not mistaken.
Still, I can add a bit of logic to ensure it is always considered first, no matter the ordering of the list.

src/subsystems/php/builders/granular-php/default.nix
DavHau: Instead of treating `inputs.php` like a flake, it treats it as a path and imports it. Currently we have a failing CI tests, because under some circumstances inputs.php is a path and not a flake. This is buggy behavior in our backward compat mechanism which I don't intend to fix anymore in favor of the upcoming v1 API changes.

aszenz:
Will php still work with the V1 api?

PR #485, opened 2023-03-13
Init module system based dream2nix (towards V1 API)

DavHau:
This adds en experimental dream2nix rewrite like described in [v1 API](https://nix-community.github.io/dream2nix/v1-api/summary.html) using the nixos module system through [drv-parts](https://github.com/DavHau/drv-parts).
This PR also adds modules for creating python packages, as an example for the new framework.

Basically the idea is that dream2nix only ever offers `drv-parts` modules.
Users can create packages by importing some of these modules.
Users can configure packages by setting options on their imported modules, or import their own modules.

Included in this PR:
- **drv-parts modules**: defined in  `v1/nix/modules/drv-parts`
  - **attrs-from-nixpkgs**: lib to extract attributes from nixpkgs packages
  - **buildPythonPackage**: module wrapping buildPythonPackage
  - **eval-cache**: module implementing a simple json based eval cache, to cache expensive nix operations like IFDs
  - **mach-nix-xs**: module implementing several aspects of python packaging (let's split that up)
  - **nixpkgs-override**: import this to inherit attributes from equally named nixpkgs packages
- **example packages**: defined in  `v1/nix/modules/drvs` (ansible, apache-airflow, odoo, pillow, tensorflow)
- **flake-parts modules** (not part of the official API):
  - **writers.nix**: contains improved `writePureShellScript[Bin]`
  - **app.update-caches**: custom app to update all eval caches for our example packages
- **fetchPythonRequirements**: A vastly improved version of what was once https://github.com/NixOS/nixpkgs/pull/121425

Not included in this PR:
- compatibility module to integrate all existing `translators` and `builders`
- unified user interfaces: How do we deal with package sets, dependency trees, community overrides etc. etc.
- deploy generated docs for all contained modules (drv-parts offers a docs module, we just need to use it)

Currently the new v1 code is located under `./v1` and integrated via flake parts into the top-level flake outputs.
Once `./v1` is ready, I'd like to move its contents to the top-level and instead move the previous top-level to `./legacy`.

Big Thanks to @phaer for putting a lot effort into the python modules.

I might merge this quickly to unblock further development. Reviews are still appreciated and can be incorporated in upcoming PRs.


hsjobeki:
@DavHau different ecosystems need to have different and interchangeable modules. Those modules need to have a clear defined interface. 
I am planning to refactor / implement my nodejs strict builder based on this idea. But i might need some initial help for bootstrapping.

PR #493, opened 2023-03-18
PHP: don't override patchPhase

raboof:
Perform the patching in `postPatch` instead.

The advantage is that this means the original `patchPhase` becomes available again, so people can use `patches` in overrides. They can also still override `postPatch` if they want, and invoke the 'original' `postPatch` if they also want to keep the php builders' behavior.

### DIFF
--- a/src/subsystems/php/builders/granular-php/default.nix
+++ b/src/subsystems/php/builders/granular-php/default.nix
@@ -181,20 +181,11 @@
           # set name & version
           jq \
             "(.name = \"${name}\") | \
-             (.version = \"${versionString}\")" \
+             (.version = \"${versionString}\") | \
+             (.extra.patches = {})" \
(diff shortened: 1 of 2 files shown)


DavHau:
How about using [preConfigurePhases](https://nixos.org/manual/nixpkgs/stable/#ssec-controlling-phases) ?
That leaves all the other variables free for the user

raboof:
> How about using [preConfigurePhases](https://nixos.org/manual/nixpkgs/stable/#ssec-controlling-phases) ? That leaves all the other variables free for the user

Sure! Or perhaps, looking at this again, we should just add it to `unpackPhase` as that's manipulating the composer.json anyway?

raboof:
(updated to do all composer.json updates in unpackPhase)

PR #497, opened 2023-03-21
small fixes + python: read egg-info metadata, not requirements.txt

phaer:
* first commit prevents infinite recursion in eval-cache if documentation is generated via drv-parts
* second commit prevents a failure if nixpkgs-overrides is called for a python package not in nixpkgs
* 3 commit replaces the ugly workaround of parsing requirements.txt with a way to parse egg-info metadata from source dists. Together with pkginfo, that seems to cover most packages.


v1/nix/modules/drv-parts/nixpkgs-overrides/default.nix
```diff
@@ -19,7 +19,10 @@
   in
     l.filterAttrs (name: _: ! excludedNixpkgsAttrs ? ${name}) pythonAttrs;
 
-  extracted = extractPythonAttrs config.deps.python.pkgs.${config.name};
+  extracted =
+    if config.deps.python.pkgs ? config.name
```
DavHau, line 23: `? config.name` -> `? ${config.name}`

phaer: :see_no_evil: thanks!

PR #501, opened 2023-03-30
fetchPip: use new `pip install --report` to acquire reliable metadata from pip

phaer:
This fixes a bug where the wrong python was used during fetching (3a242e39b2a6c3fca3209adefbe992bc20005b03) and otherwise adds a `/metadata.json` to the FOD, replacing `dependencies.json`. See snippet below.

Further work is planned in an upcoming PR to separate metadata fetching for editable Installs from the FOD and therefore refactor parts of fetchPip.

```json
{
  "ansible": {
    "version": "2.7.1",
    "dependencies": [
      "cryptography",
      "jinja2",
      "paramiko",
      "pyyaml",
      "setuptools"
    ],
    "file": "ansible-2.7.1.tar.gz",
    "hash": "e7953472347fcc6dca10839111b576a9f790e00056344f2dcf448e6c452fe939"
  },
  "cryptography": {
    "version": "38.0.4",
    "dependencies": [
      "cffi"
    ],
    "file": "cryptography-38.0.4-cp36-abi3-manylinux_2_28_x86_64.whl",
    "hash": "ce127dd0a6a0811c251a6cddd014d292728484e530d80e872ad9806cfb1c5b3c"
  },
[...]
```


v1/nix/pkgs/fetchPip/fetchPip.nix
```diff
@@ -96,7 +96,11 @@
   # we use mitmproxy to filter the pypi responses
   pythonWithMitmproxy =
     python3.withPackages
-    (ps: [ps.mitmproxy ps.python-dateutil ps.pkginfo ps.packaging]);
+    (ps: [ps.mitmproxy ps.dateutil]);
+
+  pythonWithPackaging =
```
DavHau, line 101: Maybe add a comment here on why both python environments are needed.
I tend to forget and already cycled through removing one and reverting it again before.

v1/nix/pkgs/fetchPip/fetchPip.py
```diff
@@ -113,35 +121,55 @@ def pip(venv_path, *args):
...
+        cache_path,
     ]
-    if REQUIREMENTS_FILES:
-        optional_flags += ["-r " + " -r ".join(REQUIREMENTS_FILES.split())]
     if NO_BINARY:
         optional_flags += ["--no-binary " + " --no-binary ".join(NO_BINARY.split())]
+    if WRITE_METADATA:
+        metadata_flags = ["--report", "/build/report.json"]
+
+    for req in REQUIREMENTS_LIST.split(" "):
```
DavHau, line 151: Something I completely overlooked so far. Won't we run into issues with spaces in requirements which contain markers etc. which leads to wrong splitting. Should we pass the requirements list as a JSON string?

DavHau: ... or alternatively just switch to [__structuredAttrs](https://nixos.mayflower.consulting/blog/2020/01/20/structured-attrs/) all the way.

v1/nix/modules/drvs/ansible/cache-x86_64-linux.json
DavHau, line 1: While this is a step forward for many scenarios it is also a step back for others.
When using in nixpkgs, for example, it will likely be preferred to minimize the file size and use the files coming from the large FOD directly.
Also in large CI systems eval time is a concern which usually goes up the more derivations we haven.

Nothing needs to be done for now, but in the future I would like the user to be able to configure the amount of data, which is put in the cache.

To allow this use case, it would be great if we do not start hard coding procedures that require the `file` and `hash` entry.
Do you think that this is possible?

DavHau (commented):
Cool! I really like how this simplifies the FOD logic and most likely will make it a hell lot more reliable!

v1/nix/modules/drvs/ansible/cache-x86_64-linux.json
phaer: > When using in nixpkgs, for example, it will likely be preferred to minimize the file size and use the files coming from the large FOD directly.

As recently discussed, I still hope that we can get rid of the large FOD for many if not most cases - i.e. by implementing pure translators for things like poetry and relying on the `--report` metadata to parameterize our fetchers for pip. 
If we follow that route, I believe it's a good idea to treat the FOD as an implementation detail and not to rely on it w.r.t. the API. 

I am open for ideas here, but the only alternative to passing the filename in metadata (or soon: the lockfile) would be to look them up via IFD at build-time. As i understand it, this causes one IFD operation per input file (vs one IFD operation for the whole metadata file which we need anyway) and has the drawback that other nix configuration can't refer to `mkDerviation.src` during eval time. This can i.e. be used to check if src is as wheel in pure nix.

> Nothing needs to be done for now, but in the future I would like the user to be able to configure the amount of data, which is put in the cache.

I agree about the cache, but this data will be moved to the lock-file when i change fetchPip to do metadata only.

v1/nix/pkgs/fetchPip/fetchPip.py
phaer: It should work as expected by accident for those, as they will be joined by ` ` in L158 without any separators. But the current implementation would break if a requirement file name would include a ` `, because then we end up with an extra `-r`. 

__structuredAttrs is surely a good idea, but as thats not really related to `--report` and i believe it isn't much more broken than on main, I'd prefer to do this in a follow-up pr?

v1/nix/pkgs/fetchPip/fetchPip.nix
phaer: Is this good enough? There are additional comments at the top of the file :)

v1/nix/modules/drvs/ansible/cache-x86_64-linux.json
DavHau: > I am open for ideas here, but the only alternative to passing the filename in metadata (or soon: the lockfile) would be to look them up via IFD at build-time.

That's only true if we always do granular fetching. But I'm not sure if granular fetching is preferred in any scenario given the performance drawbacks etc. If we don't fetch granular, we don't even need the file name at eval time.

We are moving in a space of dimensions `correctness`, `performance`, `simplicity`. Currently we are adding complexity to achieve better correctness but moving away from performance and simplicity.

To provide a simple example. If we want to build rust projects, we can choose between buildRustCrate which just wrapps cargo, does not have great introspection capabilities, but is very simple. On the other hand we can use `crate2nix` which doesn't use cargo at all, neither for fetching nor for building. It achieves granular fetching and building, but it is quite complex, as it re-implements much of cargo in nix. Due to this, packages maintained with `crate2nix` are harder to maintain and encounter more bugs.

I just want us to be aware that there is no simple right or wrong. Different users have different priorities. For some production use cases it is often most important that the solution is very simple and almost never breaks. Much more than anything else.

Anyways, I'm happy to move forward with this as you propose. We can still re-introduce a more stripped down version of this module later. Let's just try to keep things modular wherever possible, so that this change will be an easy one.

> when i change fetchPip to do metadata only

I think we should at least keep the old behavior (outputting files) as an option and not remove it entirely.

PR #510, opened 2023-04-26
fix: nodejs: strict-builder: patch shebangs

tgunnoe:
The shebangs were not patched for build time in the strict-builder.  That lead to `no such file or directory` in the sandbox. 
reference: https://matrix.to/#/!PrIEtcffTLLzUWvJOS:matrix.org/$zslZWYi4HF0xramEQum6xf808-VQbbJIkN4kbgGFqt0?via=nixos.org&via=matrix.org&via=nixos.dev

Also, the bins weren't runnable permissions.

### DIFF
--- a/src/subsystems/nodejs/builders/strict-builder/default.nix
+++ b/src/subsystems/nodejs/builders/strict-builder/default.nix
@@ -192,7 +192,8 @@
             runHook preConfigure
 
             cp -r ${nodeModules} ./node_modules
-            chmod -R +w node_modules
+            chmod -R +xw node_modules
+            patchShebangs ./node_modules
 
             export NODE_PATH="$NODE_PATH:./node_modules"
             export PATH="$PATH:node_modules/.bin"


DavHau:
@hsjobeki I wonder why the we even produce node_modules that contain broken shebangs. Usually patchShebang is executed automatically in the fixup phase of each individual derivation, so there shouldn't be any corrupt shebangs.

hsjobeki:
Strange. I‘d need to investigate this. 
Two questions:
- Should „patchShebangs“ be run only on node_modules/.bin 
- Should patch shebangs be in node-modules-tree.nix  (where the node_modules are created.

tgunnoe:
> * (if needed manually) Should „patchShebangs“ be run only on node_modules/.bin
It wasn't straightforward to only patch what's in `.bin`with patchShebangs. I suppose it needs to follow symlinks or something similar. However, I assumed finally that probably everything in node_modules needs to be patched for their scope, but I dont know how much of a performance loss that would entail.

> * Do we really need „+x“ applied to every file in the whole node_modules?

No, there can be a cleaner operation for the runnables. Any suggestions?

> * Should patch shebangs be in node-modules-tree.nix (where the node_modules are created.

I can do it there instead. Sorry, still not completely familiar with this builder

tgunnoe:
> I wonder why the we even produce node_modules that contain broken shebangs


Oh, I suppose since the tree is built fully with python, there's not a fixup phase that the derivation goes into at that time. As far as I understand, anyways.  I don't know what that would entail to run the nix path substitution from within the python builds

> If you could move this fix to node-modules-tree.nix and check if your use case still works.

Sounds good! no problem

DavHau:
> Oh, I suppose since the tree is built fully with python

Oh yeah, I forgot about the node-modules-tree.nix

> If you could move this fix to node-modules-tree.nix and check if your use case still works.

Note that patchShebangs can only substitute paths in shebangs for known buildInputs. The [runCommandLocal](https://github.com/nix-community/dream2nix/blob/bef06f0ad8feec43997ced0489e760b4c15b4012/src/subsystems/nodejs/builders/strict-builder/lib/node-modules-tree.nix#L130) which is used to execute the python builder probably needs nodejs as a build input and any other things that are expected to appear in shebangs.

PR #519, opened 2023-05-15
Bump version of nixpkgs to include the fix for poetry on m1 macbooks

belevy:
The node strict-builder uses poetry to build itself. On the version of nixpkgs that is used poetry fails its check phase for m1 macbooks. This PR bumps the version up ~10 days to the point where that fix was merged in. 

Running on m1 with this commit the nodejs alternative builder example should succeed now.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -192,16 +192,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1665580254,
-        "narHash": "sha256-hO61XPkp1Hphl4HGNzj1VvDH5URt7LI6LaY/385Eul4=",
+        "lastModified": 1666441811,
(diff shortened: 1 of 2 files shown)


phaer:
@belevy The commit you pinned to might have been added to nixos-unstable in the 2 weeks since this PR?

Generally, I don't think pinning to a specific commit of nixpkgs is a good solution, so I'll close this. You can use `--override-input` to use a different version of nixpkgs locally (or use `.follows` in your flake inputs)

srid:
@phaer Agreed on commit pinning; would you be open to updating just the `flake.lock` file?

phaer:
@srid Yes, absolutely! PR welcome, otherwise I'll try to get around to it soon

PR #528, opened 2023-06-14
feat: nodejs + python mixed language derivation

DavHau:


flake.nix
```diff
@@ -8,7 +8,7 @@
 
   inputs = {
     nixpkgs.url = "nixpkgs/nixos-unstable";
-    nixpkgsV1.url = "nixpkgs/nixos-22.11";
+    nixpkgsV1.url = "nixpkgs/nixos-unstable";
```
phaer, line 11: Input change might be worth cherry-picking into a separate PR for visibility in history.

v1/nix/modules/drv-parts/nixpkgs-overrides/interface.nix
```diff
@@ -14,6 +14,12 @@ in {
       description = "Attributes we do not want to copy from nixpkgs";
     };
 
+    from = l.mkOption {
+      type = t.nullOr t.package;
+      description = "package from which to extract the attributes";
+      default = config.deps.python.pkgs.${config.name} or null;
```
phaer, line 20: Is a python-specific default a good idea here?

This could be a good example of the need for `depsModules`or so in drv-parts?

v1/nix/modules/drvs/apache-airflow/default.nix
```diff
@@ -6,7 +6,7 @@
   l = lib // builtins;
   python = config.deps.python;
   extractPythonAttrs = config.nixpkgs-overrides.lib.extractPythonAttrs;
-  nixpkgsAttrs = extractPythonAttrs python.pkgs.apache-airflow;
+  nixpkgsAttrs = extractPythonAttrs config.deps.apache-airflow;
```
phaer, line 9: This conflicts with my recent changes to this package in https://github.com/nix-community/dream2nix/pull/529

v1/nix/modules/drv-parts/pip/default.nix
```diff
@@ -3,9 +3,8 @@
   lib,
   dream2nix,
   ...
-}: let
+} @ topLevel: let
```
phaer, line 6: As discussed, I don't think those changes are a good idea in the long run. I think a solution similar to the usage of `setuptools_scm` in `apache-airflow` in https://github.com/nix-community/dream2nix/pull/529 could help with your setuptools issue in a cleaner way?

phaer (changes requested):
Great work, really nice to see a mixed derivation working here :)

See comments for issues/discussion points. Might also be worthwhile to look further into floco here, especially for its npm override capabilities.

v1/nix/modules/drv-parts/nixpkgs-overrides/interface.nix
DavHau: currently this module contains python specific logic in several places. We should split it into `nixpkgs-overrides` and `nixpkgs-overrides-python` at some point I guess. I'd say we better do this in another PR.

v1/nix/modules/drvs/apache-airflow/default.nix
DavHau: I'm not quite sure why. The package seems to build fine

v1/nix/modules/drv-parts/pip/default.nix
DavHau: Reverted the changes

v1/nix/modules/drvs/apache-airflow/default.nix
phaer: You don't need it anymore, when building apache-airflow with the new python mod

v1/nix/modules/drv-parts/nodejs-package-json/default.nix
```diff
@@ -38,7 +47,7 @@ in {
           writeScript
           writeScriptBin
           ;
-        inherit (nixpkgs.nodePackages) npm;
+        npm = npm_8 nixpkgs.nodejs;
```
phaer, line 50: If i read this correctly, this means we are vendoring an older npm - which would trigger insecure errors if used from (older) nixpkgs -  and use it with newer nodejs, where it's effectively unsupported?

I understand we need to work around the missing support for v2 and v3 lock files in our builder, but defaulting to the latest supporting nodejs and requiring users to allow the insecure package would be a better solution IMO.

v1/nix/modules/drv-parts/nodejs-granular/interface.nix
```diff
@@ -33,7 +33,7 @@ in {
         Whether to run a package's build script (aka. `npm run build`)
       '';
     };
-    nodejsDeps = {
+    deps = {
```
phaer, line 36: While it's also including an extra level for package versions, conceptually it's the same as `pip.drvs`, right? Maybe we should call it like that?

v1/nix/modules/drv-parts/nixpkgs-overrides/interface.nix
phaer: Another PR is fine with me, but we should not forget it. I think the default isn't even used atm? But on second reading it's also no hard dependency on config.deps.python, so I don't care too much.

phaer (changes requested):

v1/nix/modules/drv-parts/nodejs-granular/interface.nix
DavHau: Regarding the naming, I agree let's consolidate these. Though I think just `deps` is a much better name than `drvs`. I'd prefer to do this in another PR.

Regarding the structure. We cannot have the same as in python, because in nodejs we can have several versions of the same dependency under `deps`.

flake.nix
DavHau: done, see https://github.com/nix-community/dream2nix/pull/532

v1/nix/modules/drvs/apache-airflow/default.nix
DavHau: removed it

v1/nix/modules/drv-parts/nodejs-package-json/default.nix
DavHau: solved as discussed

v1/nix/modules/drv-parts/nodejs-granular/interface.nix
phaer: Yes, structure can't obviously be the same. Not sure about `deps` though, as that could cause confusion between e.g. `config.deps` and `config.pip.deps`?

phaer (approved):

PR #531, opened 2023-06-20
Bump nixpkgs to commit that fixed poetry on m1. Update nodjs_alternat…

belevy:
Based on #519. Instead of changing the flake.nix instead just modified the flake.lock to refer to the newer commit per @phaer

### DIFF
--- a/examples/nodejs_alternative_builder/flake.nix
+++ b/examples/nodejs_alternative_builder/flake.nix
@@ -16,7 +16,7 @@ Building 'Prettier@2.4.1'.
     src,
   }:
     (dream2nix.lib.makeFlakeOutputs {
-      systems = ["x86_64-linux"];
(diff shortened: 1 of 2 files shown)


examples/nodejs_alternative_builder/flake.nix
```diff
@@ -16,7 +16,7 @@ Building 'Prettier@2.4.1'.
     src,
   }:
     (dream2nix.lib.makeFlakeOutputs {
-      systems = ["x86_64-linux"];
+      systems = ["x86_64-linux" "aarch64-darwin"];
```
phaer, line 19: As you see in the test log, we can't currently test this. So we either need to exclude it from tests or get https://github.com/nix-community/infra/issues/493 finished before merging if we include this in the example.

We could ofc just bump the flake lock in a separate PR if tests pass.

phaer (changes requested):

PR #533, opened 2023-06-21
Make hash a required field for `http` type in dream-lock-schema.json.

zainab-ali:
I noticed that `hash` is a required field in the documentation for fetchers (which makes complete sense). However, it's not required according to the json schema.

I have edited the schema manually as part of this PR. Let me know if there's another source of truth that I should edit.

There might be other places in which the `hash` is missing. If this change is appropriate, I'll happily add those in.

### DIFF
--- a/src/specifications/dream-lock-schema.json
+++ b/src/specifications/dream-lock-schema.json
@@ -91,7 +91,7 @@
                       "type": { "type": "string" },
                       "dir": { "type": "string" }
                     },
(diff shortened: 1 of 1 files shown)


DavHau:
Hey, thanks for your contribution. I think we intentionally do not require the hash, because one feature of the dream-lock is to optionally have only a combined hash over all inputs instead of individual ones.

Issue #474, opened 2023-02-20
Problem with python packaging on macOS (aarch64-darwin)


pelme:
Hey, wanted to play with dream2nix with Python on my Mac.

I tried following the "Python: getting started in 10 minutes" tutorial, this is what happens:

```
~ $ git clone https://github.com/httpie/httpie /tmp/my_project
Cloning into '/tmp/my_project'...
remote: Enumerating objects: 9749, done.
remote: Counting objects: 100% (7/7), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 9749 (delta 2), reused 4 (delta 1), pack-reused 9742
Receiving objects: 100% (9749/9749), 6.69 MiB | 1.47 MiB/s, done.
Resolving deltas: 100% (6646/6646), done.
~ $ cd /tmp/my_project/
master /tmp/my_project $ nix flake init -t github:nix-community/dream2nix#simple
warning: ignoring untrusted substituter 'https://nix-community.cachix.org'
wrote: /private/tmp/my_project/.envrc
wrote: /private/tmp/my_project/flake.nix


    You just created a simple dream2nix package!

    start with typing nix flake show to discover the projects attributes.

    commands:

    · nix develop <-- enters the devShell
    · nix build .# <-- builds the default package (.#default)

    Start hacking and -_- have some fun!

    | dont forget to add nix result folder to your .gitignore

direnv: error /private/tmp/my_project/.envrc is blocked. Run `direnv allow` to approve its content
master /tmp/my_project $ nix eval --impure --raw --expr 'builtins.currentSystem' > ./nix_systems
master /tmp/my_project $ cat nix_systems
aarch64-darwin⏎
master /tmp/my_project $ git add .
master /tmp/my_project $ nix run github:nix-community/dream2nix#detect-projects . > projects.toml
warning: ignoring untrusted substituter 'https://nix-community.cachix.org'
master /tmp/my_project $ git add .
master /tmp/my_project $ nix flake show
warning: Git tree '/private/tmp/my_project' is dirty
warning: creating lock file '/private/tmp/my_project/flake.lock'
warning: Git tree '/private/tmp/my_project' is dirty
git+file:///private/tmp/my_project
├───apps
│   └───aarch64-darwin
│       └───detect-projects: app
└───packages
    └───aarch64-darwin
        ├───main: package 'main'
        └───resolveImpure: package 'resolve'
master /tmp/my_project $ nix run .#resolveImpure
[...]

Issue #507, opened 2023-04-19
Release v1 API


phaer:
We believe it's about time to move from the old, experimental API (`legacy`) to the new one, sketched out in https://github.com/nix-community/dream2nix/pull/468 and (partially) implemented in the `v1/` directory in the main branch .

The new API is based on [drv-parts](https://github.com/DavHau/drv-parts) and does not require flake-enabled nix, but ships examples based on [flake parts](https://flake.parts/) for those of us who like flakes.

This issue aims to provide an overview of the work we still need to do before deprecating the existing API and moving `v1` to the top level of this repo. We've discussed this list in the last weekly call on 2023-04-18, you are welcome to provide feedback either here or in next weeks call (2024-04-25 14h Utc).

# Documentation

* [ ] update readme
* [ ] update manual
 * [ ] Introduction
 * [ ] Guides
 * [ ] Subsystems (those which we'll keep, see below)
 * [ ] Concepts / API: Explain drv-parts API, lock & eval-cache, nixpkgs-overrides modules
 * [ ] Contributing
 * [ ] v1 -> Move to Concepts

# Support Policies

* [ ] Define a policy for release
 We want git tags per release and a changelog announcing breaking changes with categories for the general API and specific subsystems 
* [ ] Write a paragraph or table per subsystem module to describe features, limitations and expected stability of the subsystem.

# Update Ecosystems

While there's a proof-of-concept wrapper which wraps the old API in terms of the new one (`dream2nix-legacy` module), this would still mean that we need to keep the old API inside our tree. We've decided to instead do a proper port of some subsystems to the new API. 
Due to the different completeness, community interest, and maintenance burden of each subsystem, we are probably going to drop some of them:

* nodejs: will be ported, @DavHau 
* rust: will be ported, @DavHau 
* python: examples will be ported, old API dropped. @phaer 

The following subsystems would be dropped. Please contact us here, on matrix or in a weekly call if you are interested into maintaining one of them - we'll happily provide mentoring and support.

* debian
* go
* haskell
* php
* racket

# Testing

* We'd like to have a more flexible test runner, allowing selection of tests per name or tag, like pytest cli,
in order to split tests in "slow" and "fast" sets. The idea is to make it easier and faster to run unit tests which primarily checks
api assertions and evaluation, while still being able to run heavier integration tests in CI and locally.
* We also would like to automatically test code examples included in our documentation.

DavHau:
Some more tasks that should be completed:
[...]

PR #542, opened 2023-06-30
Prototype MacOS CI

DanLipsitt:
This PR sets up Github Actions CI for MacOS. It uses a build matrix to run on both Mac and Linux builders. I've made some changes that affect the Linux build in the interest of getting this prototype working. If people are interested in this addition, I can go back add the conditional settings that will make the Linux build behave as it did before.

### Changes

- Run all tests except `tests-format` on both Mac and Linux. That one doesn't appear to be platform-specific.
- Upgrade `cachix/install-nix-action` to v22. The [release notes](https://github.com/cachix/install-nix-action/releases/tag/v22) mention a MacOS issue with previous versions.
- Un-pin `nixpkgs`. The pinned version broke the Mac build.

### Discussion

I'm new to Nix, so I don't fully understand how requesting a specific `nixpkgs` version in the install phase interacts with the `flake.lock`. The Linux tests pass after unpinning. It seems like it would be good to test against unstable as well as the pinned version anyway. If we want, we can put both in the build matrix.
This PR was motivated by #474. The CI jobs exhibit failures of several different kinds, so the PR uncovers at least one issue unrelated to #474 (and might not overlap with it at all).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,11 +12,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2.4.0
-    - uses: cachix/install-nix-action@v17
+    - uses: cachix/install-nix-action@v22
       with:
-        # version 2.11.0
-        install_url: https://nixos-nix-install-tests.cachix.org/serve/8i0c4nnvddybdq6a684ffybc5p7ziif0/install
-        install_options: '--tarball-url-prefix https://nixos-nix-install-tests.cachix.org/serve'
         extra_nix_config: |
           access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
           max-jobs = 10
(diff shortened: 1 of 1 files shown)


DavHau:
Thanks for this!

What is pinned through `install_url` is not the nixpkgs version but the nix version. Since you upgraded the action anyways, the pin is probably not required anymore.

DavHau:
@DanLipsitt I just noticed that most of these tests will be deprecated as soon as [we switch over to v1](https://github.com/nix-community/dream2nix/issues/507). Having functional CI pipelines for macos is still valuable but maybe it makes sense to wait until we have switched over, to prevent unnecessary work.

DanLipsitt:
Thanks, @DavHau. Is there a v1 branch where it would make more sense to do this? I see @phaer's [v1-test-runner](https://github.com/nix-community/dream2nix/tree/v1-test-runner) branch and your new [actions-macos-dave](https://github.com/nix-community/dream2nix/tree/actions-macos-dave). I don't mind doing extra work if it will be useful to build on later, but I'm also happy to leave it to others if it's already being taken care of.

DavHau:
The v1 code is living on main under the subdir `v1/`. The only tests for v1 are example packages defined under `v1/nix/modules/drvs` and are exposed via the main flake.

You might want to play with it a little and track our progress via the issue linked above.

During the next few days we plan to create more detailed issues and probably a project board for moving v1 forward. After that it should be easier to see how to get involved.

phaer:
Hello & thanks for your contribution @DanLipsitt !

It's a bit unlucky timing, due to the v1 switch as @DavHau wrote. We also started to use nix-community's hercules-ci.com account for a nix-native CI infra.
Hercules also has the advantage that we got an aarch64-darwin builder there, which github actions doesn't support afaik.

I've started a branch to use those builders and it seems to work well in principle, even though some jobs are still broken atm:  https://github.com/nix-community/dream2nix/pull/576

So all in all I'd propose we continue with hercules ci and close this?

phaer:
We since have switched to a nix-community-hosted buildbot instance which features support for both aarch64-darwin and x86_64-darwin.

PR #547, opened 2023-07-03
packages: allow all systems...

phaer:
..and don't filter them. I think the original idea was to prevent evaluation errors on untested platforms, but we didn't achieve that goal on current main anyway and i think its more useful to tackle that issue on the CI front.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,25 +5,6 @@ on:
 
 jobs:
 
-  tests-pure:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v3.5.3
-    - uses: cachix/install-nix-action@v22
-      with:
-        install_url: https://nixos-nix-install-tests.cachix.org/serve/8i0c4nnvddybdq6a684ffybc5p7ziif0/install
-        install_options: '--tarball-url-prefix https://nixos-nix-install-tests.cachix.org/serve'
-        extra_nix_config: |
-          access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
-          max-jobs = 10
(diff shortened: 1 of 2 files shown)


phaer:
Hm... that fails while running .update-locks on apache-airflow for aarch64-linux. Do the builders have a remote builder for aarch64 configured or so? 

Because I can't reproduce that behavior locally, neither on x86_64-linux nor on aarch64-darwin.
update-locks uses everything in `self'.packages` as input, but if i understand flake.parts correctly, that means `nix run .#update-locks` should only update for the current platform, correct?

DavHau:
The problem is that the action only updates the lock files for linux, but then runs `nix flake check` which evaluates the packages for all platforms.

phaer:
> The problem is that the action only updates the lock files for linux, but then runs `nix flake check` which evaluates the packages for all platforms.

Ah right! That's an annoying behavior of flake checks I always tend to forget about it.

Anyway, I think restricting dream2nix to x86_64-linux is not what we want. So could we e.g. move the tests from checks to just an app to run or something as a workaround @DavHau ?

DavHau:
Not sure why moving to an app would help. Still we need to lock files in order for the packages to evaluate.
Also we might want to get rid of the `nix flake check` job, and instead build all the things via hercules.

phaer:
> Not sure why moving to an app would help. Still we need the lock files in order for the packages to evaluate. Also we might want to get rid of the `nix flake check` job, and instead build all the things via hercules.

Yes, the latter is what meant. I mistakenly assumed that we were adding something to .checks, but the root cause is the automatic inclusion of packages and so you are correct that we would need to forgoe nix flake check entirely.

PR #548, opened 2023-07-03
Expose writers

phaer:
You'd need them for most of our examples yet, but they aren't available to external flake except for importing them via the path directly.

Builds on https://github.com/nix-community/dream2nix/pull/547 atm, but i am happy to change that if needed.

### DIFF
--- a/v1/nix/modules/flake-parts/writers.nix
+++ b/v1/nix/modules/flake-parts/writers.nix
@@ -1,12 +1,11 @@
 {
   perSystem = {
+    self,
     config,
     lib,
     pkgs,
     ...
-  }: let
-    writers = pkgs.callPackage ../../pkgs/writers {};
-  in {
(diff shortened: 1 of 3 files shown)


v1/nix/modules/flake-parts/writers.nix
```diff
@@ -1,12 +1,11 @@
 {
   perSystem = {
+    self,
     config,
     lib,
     pkgs,
     ...
```
DavHau, line 7:
```suggestion
{
  self,
  ...
}: {
  perSystem = {
    config,
    lib,
    pkgs,
    ...
```

DavHau: This should fix the breakage in tests-pure.

phaer:
closed in favor of #605

Issue #553, opened 2023-07-06
Could not find nodejs version '20' in pkgs


xyven1:
I am receiving this error when trying to build:
![image](https://github.com/nix-community/dream2nix/assets/35360746/4cbd50ad-0019-48fd-8717-50459b97aa5d)
Repo in question is [here](https://github.com/xyven1/home-management).
I can post trace if needed, but the error originates from [here](https://github.com/nix-community/dream2nix/blob/5a3e87301fedebbb116225d53d2ef2bdaf8f3318/src/subsystems/nodejs/builders/strict-builder/default.nix#L52C5-L57C79).

Where is pkg originating from within dream2nix? Is there any way it could depend on my system config (running nixos)?

xyven1:
Seems to be an issue with how packages are named. Node versions used to be named `nodejs-##_x` but now they are named `nodejs_##`

xyven1:
Created a fix.

PR #554, opened 2023-07-06
Fix nodejs naming

xyven1:
This is a fix for #553


DavHau:
This only seems to work for newer nodejs versions and fails for older ones.

xyven1:
Any clue why that is? If I go to nixpkg search I can see node versions node_(14|16|18|20), so I assumed all of those versions would work

DavHau:
Did you check the exact nixpkgs version pinned by dream2nix?

xyven1:
Nope, guess I shoulda done that

PR #576, opened 2023-07-12
CI: build all systems in hercules

phaer:
Since we [switched to hercules ci](https://github.com/nix-community/dream2nix/commit/27468aef5a466f625db3a3edcac003b5f8bd35ca) on monday, we can build checks and other packages in CI on the following architectures:

* aarch64-linux
* aarch64-darwin
* x86_64-darwin

This is a draft as I'd like to take a day or two trying to fix the still-failing builds or disabling them and linking a specific github issue for each.

## Questions

* [ ] Could we refresh lock files in CI?
  Need to check whether that's better implemented via hercules effects or impure derivations?
  It would already be very useful to at least verify that lock-files are reproducible - and the  
  copies in the repo are up to date.
  Maybe there could even be an optional job to refresh lock files on other platform, so that contributors could i.e. commit their x86_86-linux lock file and CI builds aarch64-linux et al?
* Should we run this on each and every commit?
  I think with the current state of things and the fact that most contributors won't have access to each supported systems.
  But you need that to keep lock files up to date, we can't expect all platforms to work on each commit to `main`.
  Additionally, we are effectively doing ~4 times the work every time, while platform-specific breakage will be much more likely for some commits (i.e. drvs changes, compared to refactoring of functions).
  -> We could only run x86_64-linux on commits to main and only test the rest for releases or upon request (not sure how to implement that in hercules yet)

### DIFF
--- a/ci.nix
+++ b/ci.nix
@@ -3,9 +3,28 @@ let
   flakeCompatSrc = b.fetchurl "https://raw.githubusercontent.com/edolstra/flake-compat/12c64ca55c1014cdc1b16ed5a804aa8576601ff2/default.nix";
   flake = (import flakeCompatSrc {src = ./.;}).defaultNix;
   pkgs = import flake.inputs.nixpkgs {};
-  recurseIntoAll = b.mapAttrs (name: val: pkgs.recurseIntoAttrs val);
+  mapRecurse = attrs: f:
+    b.mapAttrs f attrs
+    // {
+      recurseForDerivations = true;
+    };
 in
-  recurseIntoAll {
-    checks = flake.checks.x86_64-linux;
-    packages = flake.packages.x86_64-linux;
+  mapRecurse
+  {
+    x86_64-linux = {};
+    x86_64-darwin = {};
+    aarch64-darwin = {
+      knownIssues = {
+        "odoo" = "TODO github issue link here";
+      };
+    };
   }
+  (
+    system: {knownIssues ? {}}:
+      mapRecurse
+      {
+        checks = flake.checks.${system};
+        packages = pkgs.lib.filterAttrs (n: v: ! b.hasAttr n knownIssues) flake.packages.${system};
+      }
+      (_name: val: pkgs.recurseIntoAttrs val)
+  )
--- a/ci.nix
+++ b/ci.nix
@@ -11,10 +11,13 @@ let
 in
   mapRecurse
   {
+    # FIXME: We can't lock via remote-builders yet, and I don't have a aarch64-linux
+    # machine here, so we lack lock files for the evaluation atm.
+    # aarch64-linux = {};
     x86_64-linux = {};
-    aarch64-linux = {};
     x86_64-darwin = {};
     aarch64-darwin = {
+      # FIXME: hercules requireFailure might be a better solution?
       knownIssues = {
         "odoo" = "TODO github issue link here";
       };
--- a/ci.nix
+++ b/ci.nix
@@ -12,6 +12,7 @@ in
   mapRecurse
   {
     x86_64-linux = {};
+    aarch64-linux = {};
     x86_64-darwin = {};
     aarch64-darwin = {
       knownIssues = {
--- a/ci.nix
+++ b/ci.nix
@@ -16,12 +16,7 @@ in
     # aarch64-linux = {};
     x86_64-linux = {};
     x86_64-darwin = {};
-    aarch64-darwin = {
-      # FIXME: hercules requireFailure might be a better solution?
-      knownIssues = {
-        "odoo" = "TODO github issue link here";
-      };
-    };
+    aarch64-darwin = {};
   }
   (
     system: {knownIssues ? {}}:
--- a/v1/nix/modules/drvs/ansible/lock-aarch64-darwin.json
+++ b/v1/nix/modules/drvs/ansible/lock-aarch64-darwin.json
@@ -1,0 +1,101 @@
+{
+  "fetchPipMetadata": {
+    "sources": {
+      "ansible": {
+        "sha256": "e7953472347fcc6dca10839111b576a9f790e00056344f2dcf448e6c452fe939",
+        "url": "https://files.pythonhosted.org/packages/ec/ee/1494474b59c6e9cccdfde32da1364b94cdb280ff96b1493deaf4f3ae55f8/ansible-2.7.1.tar.gz",
+        "version": "2.7.1"
+      },
+      "bcrypt": {
+        "sha256": "b1023030aec778185a6c16cf70f359cbb6e0c289fd564a7cfa29e727a1c38f8f",
+        "url": "https://files.pythonhosted.org/packages/78/d4/3b2657bd58ef02b23a07729b0df26f21af97169dbd0b5797afa9e97ebb49/bcrypt-4.0.1-cp36-abi3-macosx_10_10_universal2.whl",
+        "version": "4.0.1"
+      },
+      "cffi": {
+        "sha256": "fcd131dd944808b5bdb38e6f5b53013c5aa4f334c5cad0c72742f6eba4b73db0",
+        "url": "https://files.pythonhosted.org/packages/3a/75/a162315adeaf47e94a3b7f886a8e31d77b9e525a387eef2d6f0efc96a7c8/cffi-1.15.1-cp39-cp39-macosx_11_0_arm64.whl",
+        "version": "1.15.1"
+      },
+      "cryptography": {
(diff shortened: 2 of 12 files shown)


DavHau:
Having different lock files for different systems seems to be a python specific problem. And probably just even a pip specific one, since there are other locking tools for python, like poetry that do not have that problem.

I'm not sure if this is the right time putting effort into building infra for cross system lock file management, as that is is problem of only the pip module for now, and it's unclear if or when we will have another use case for this.

If I'm not mistaken by this observation, I'd prefer to minimize our efforts in cross system locking infra. 

Us having the `${system}` in the file name for lock files is currently just done through configuration via `lock.lockFIleRel`. We should probably move this configuration to the pip module and have a system-less default for all other modules.

I'm wondering if it would generally make more sense to redirect this effort towards improving the pip module to be able to lock multiple systems at once. This improvement would free us from this problem domain entirely as of now. AFAICT, all that needs to be done is some monkeypatching of the pip vendored `packaging` library, to be able to inject a fake `environment`, specifically a fake `sys_platform` etc. Afterwards we could simply run the fetchPipMetadata script multiple times for multiple systems in one go.
This seems like a reasonable simple fix that would spare us the overhead in Ci pipeline logic and workload.

> It would already be very useful to at least verify that lock-files are reproducible 

Lock file creation being reproducible is also a property specific to the pip module, and not a property of the lock file module in general. I'm happy to have a test, but it should be a pip specific integration test and doesn't need to be a generic feature of our infra. Still personally, this test doesn't have a high prio for me, because I think non-reproducible lock file updates are not crucial for the user. Important is, that after the locking, things are reproducible, which is still the case.

phaer:
> Having different lock files for different systems seems to be a python specific problem. And probably just even a pip specific one, since there are other locking tools for python, like poetry that do not have that problem.

This could very well be the case, because the problem as such mostly exists due to pythons sdists and their missing metadata. We wouldn't have that problem with pep-518 wheels only. 
If there's really no other subsystem with a change to be ported to v1 that requires this, I agree that it might be a good idea to move the locking stuff to a separate issue and lock into what poetry does, more on that below ;)

# cross system ci

> I'm not sure if this is the right time putting effort into building infra for cross system lock file management, as that is is problem of only the pip module for now, and it's unclear if or when we will have another use case for this.

Do we in agree that it's the right time to put effort into cross-system testing, so is this just about the lock-file management or about multi-system CI in general? Just want to clarify because I think most of the runtime overhead it would cause is due to the latter, i.e. evaluating, building, testing packages on multiple systems, not locking them. Not sure if that's controversial? I think those tests are useful as platform-specific issues and requirements seem pretty common, at least for -darwin targets.

# cross-system locking

> If I'm not mistaken by this observation, I'd prefer to minimize our efforts in cross system locking infra.

The former, cross-system locking, can be seen as less urgent I suppose, at least for CI.

The thing is that I'd like to implement the ability to refresh lock files via remote-builders, if they stay platform-specific for at least one subsystem (i believe it's necessary for python, see below.

That should ease the maintenance work during development IMO, as you wouldn't need to switch machines between committing. Think that could be doable on builders supporting the ´impure-derivations´ experimental flag. And if it is, the effort to plug that into hercules with such a builder - at least for python - does not seem too high? 
If not, we can probably live with it, not a hill I am willing to die on ;)

# making lock files cross-system

> Us having the ${system} in the file name for lock files is currently just done through configuration via lock.lockFIleRel. We should probably move this configuration to the pip module and have a system-less default for all other modules.

I agree that it shouldn't generally be the case, but just to be clear about the second sentence: We would i.e. move a `lock-x86_64-linux.json` which includes paths such as `.fetchPipMetadata.sources` to `lock-x86_64-linux.json` and `.fetchPipMetadata.x86_64-liniux.sources`?

# making pips locking cross-system

> I'm wondering if it would generally make more sense to redirect this effort towards improving the pip module to be able to lock multiple systems at once. This improvement would free us from this problem domain entirely as of now. AFAICT, all that needs to be done is some monkeypatching of the pip vendored packaging library, to be able to inject a fake environment, specifically a fake sys_platform etc. Afterwards we could simply run the fetchPipMetadata script multiple times for multiple systems in one go.

I guess we should look a bit more into what poetry does here if we want to go down that route. I think it could be tricky with regards to sdists, their build requirements and their setup.py hacks. They are becoming less and less common in bigger libraries, but leaving out support for them entirely sadly isn't really feasible yet, but I might be wrong or missing a clever hack here.

Monkeypatching pips dependencies should be a proof-of-concept before suggesting an upstream patch if necessary, but that's another discussion :)

>  This seems like a reasonable simple fix that would spare us the overhead in Ci pipeline logic and workload.

Only for locking, not for building and tests, unless I am misreading? (Not arguing about the definition of "simple fix" here ;-P)

# Should lock files be reproducible and what would that mean?

> Lock file creation being reproducible is also a property specific to the pip module, and not a property of the lock file module in general. I'm happy to have a test, but it should be a pip specific integration test and doesn't need to be a generic feature of our infra. Still personally, this test doesn't have a high prio for me, because I think non-reproducible lock file updates are not crucial for the user. Important is, that after the locking, things are reproducible, which is still the case.

tl;dr: Python-specific check works for me :)

But I think we wouldn't need `pypiSnapshotDate` in fetchPipMetadata, if we don't consider that as useful feature at all, or do we? We inherited it from the old FOD builder, but as we now have hashes and urls in the lock file directly, builds would stay reproducible without it.

My thinking was mostly that it's a useful feature to review changes of the lock files in git. I'd expect the diffs to be slightly more readable due to less random changes with the reproducible ones, but might be mistaken. For python and dream2nix dev purposes alone adding some snapshot tests to fetch_pip_metadata might be the better solution for that though, now that i thought more about it.

# Invalidation hashes in lock files

> This invalidation hash can be stored inside the lock file and re-computed in the next evaluation, which would allow us to tell early if a lock file is out of date or not. If configured correctly, there wouldn't be a need to build a package or re-lock to determine if a lock file is out of date.

I think that's probably a good idea, and at least one worth exploring. Well commented inputs to the validation hash and good test coverage are probably worth the effort here but i think the invalidation hashing was and is a useful mechanism :)

abathur:
I'm not sure what/if anything should be done about it, but I'd like to gently register that there's still a sharp corner here that may deserve to be sanded down a bit.

I'm porting a web app hosted on Heroku from a mach-nix workflow to dream2nix. One of my general goals with its dev and production workflows has been to retain compatibility with its old non-nix based git-push/requirements.txt workflow   in case I get run over by a bus and grokking the nix-based setup becomes a bottleneck.

I didn't internalize, until I got to the point of validating dev environments and production builds in CI, that the lockiles I was generating while working on macOS weren't going to be portable. That was, of course, after I'd already invested quite a bit of effort in porting everything. I thought I was almost done, but now I'm pondering what I can do to keep it from being a PITA to develop against until I have cycles to figure out if it's worth abandoning the pip/requirements format  and how straightforward that is.

For now, I think I'm settling with forcing `-${system}-` into our lockfile paths (we already have separate lockfiles for dev and prod) so that we don't end up with tedious merge conflicts or stash/restore workflows around these files.

To avoid complaining without imagining improvements:
- At the risk of being mocked if I just skimmed over an existing warning (😅), prominently mentioning this sharp corner in the documentation would be a start. 
- Examples that acknowledge this pip-idiomatic limitation and model best practices for dealing with it in the least-painful way might also be nice? (I haven't lived with my current approach yet. I _imagine_ we'll have less friction if we just fold the system into the lockfile paths for intel macs, aarch64 macs, and x86/linux-based CI systems, but I don't know if I'm just retreading something that's already been discarded by people who've done this before.
- If there are already known/obvious approaches to porting projects that depend on requirements*.txt formats to another python-based module and perhaps generating the requirements.txt files, dropping warning/caution admonitions that call out the workflow problems here and leaving breadcrumbs about those might also help. (I'm not trying to go down with the pip-ship. I feel like it's responsible to ensure I remain compatible with Heroku's normal post-push `pip install -r requirements.txt` python flow, but I'm happy to port our canonical definitions to another format as long as it's a net QoL improvement and I can generate back-compatible requirements files from it. This is somewhat complicated by us having complex requirements with private indexes and such.)

Happy to open a new issue if you think it's merited.

purepani:
Hey @abathur, using pdm might work better since there's a WIP module for that here, and they've already done the hard work of making cross-platform lock files, and have the tools to convert to and from a requirements.txt as needed.

abathur:
I suspected as much myself (I already have TODOs for this, but I'm out of time to poke at it for now) but only after I'd done all of the porting work and ran into the sharp corner. 

I'm commenting less for myself (I've already sunk in the time to learn this the hard way) than to point out that there's still a sharp corner here--others can benefit from efforts to sand it down or clearly document it.

(thanks for confirming this suspicion, though)

Issue #527, opened 2023-06-14
Racket: programs installed into /bin don't work


winny-:
1. `nix shell github:winny-/ssh-hack/runAsLibrary`
2. Try to run `ssh-hack`
3. Notice `racket -l ssh-hack` works (the library has a main entrypoint)

```
winston@stargate ~ $ nix shell github:winny-/ssh-hack/runAsLibrary
winston@stargate ~ $ ssh-hack
standard-module-name-resolver: collection not found
  for module path: (lib "ssh-hack/main.rkt")
  collection: "ssh-hack"
  in collection directories:
   /home/winston/.local/share/racket/8.6/collects
   /nix/store/bk1xqmzk9aq23zqq2yw0xazvjjk84jyg-racket-8.6/share/racket/collects/
   ... [165 additional linked and package directories]
1 winston@stargate ~ $ which ssh-hack
/nix/store/bdn3cjqs7a2sy0ffx3b4ygn2cf94x20x-racket-with-ssh-hack-env/etc/8.6/bin/ssh-hack
winston@stargate ~ $
```

leungbk:
I can reproduce this bug starting with https://github.com/nix-community/dream2nix/commit/427258be by @zainab-ali.

On https://github.com/nix-community/dream2nix/commit/850d8b1c and before, everything works fine.

@zainab-ali do you have an idea of how to fix this? From looking through an `strace` of the failing case, `standard-module-name-resolver` is presently looking through some of the paths in Racket's `config.rktd` while neglecting some of the added paths in the `PLTCONFIGDIR`'s `config.rktd`.

zainab-ali:
Sorry for the delay! I think I have a solution in https://github.com/zainab-ali/dream2nix/commit/a9e9d1ebe82af2b1733383632d46445eb6569eaf. @winny- , let me know if it works for you and I'll PR the fix.

# Problem

The `ssh-hack` binary isn't present on the `PATH`.  @winny- , do you know how `ssh-hack` is added to yours?

Here's what I get when I enter the shell:
```
λ> pwd
/home/zainab/racket/ssh-hack
λ> nix shell github:winny-/ssh-hack/runAsLibrary
λ> which ssh-hack
[...]

Issue #550, opened 2023-07-04
port php to v1


DavHau:
We probably want to create the following modules under `v1/nix/modules/drv-parts`:
- `php-composer-lock` (the translator populating `options.php-composer-lock.dreamLock`)
- `php-granular` (the builder reading the `dreamLock` option and generating a package)

And maybe also these modules:
- `php-composer-json` (impure translator using the new `lock` module similar to how `nodejs-package-json` does it)
- `php-simple` (the simple builder)

To get started I propose:
- create a example under `v1/nix/modules/drvs` that imports `php-composer-lock` and `php-granualar` which fails
- implement the `v1/nix/moduels/drv-parts/php-composer-lock` modules:
   - start by copying everything from `v1/nix/moduels/drv-parts/nodejs-package-lock`
   - adapt the `interface.nix` file
   - adapt the implementation of the `default.nix` file and ensure the internal dreamLock option is populated
   - for debugging use nix repl on the example created earlier to inspect the `.config` attr of the example package
 - implement the  `v1/nix/modules/drv-parts/php-grnular` module.
   - see other builder modules like `nodejs-granular` for comparison

aszenz:
@DavHau Made a first try, ran into some difficulties

> and ensure the internal dreamLock option is populated
How do i verify this is populated

> for debugging use nix repl on the example created earlier to inspect the .config attr of the example 

I did this 
```
nix repl '<nixpkgs>'
 
phpPkg = import ./v1/nix/modules/drvs/doctrine-dbal

phpPkg {phpPkg {lib=lib;config={};} # What config to pass here
```
Now i'm not sure which config attr to inspect

DavHau:
You can do it like this:
```
nix-repl> :lf ./.
warning: Git tree '/home/grmpf/synced/projects/github/dream2nix' is dirty
Added 24 variables.

nix-repl> packages.x86_64-linux.doctrine-dbal.config.php-composer-lock.dreamLock
```

I made your example evaluate with this patch:
```
diff --git a/v1/nix/modules/drvs/doctrine-dbal/default.nix b/v1/nix/modules/drvs/doctrine-dbal/default.nix
index 41f03dae..5f384b15 100644
--- a/v1/nix/modules/drvs/doctrine-dbal/default.nix
+++ b/v1/nix/modules/drvs/doctrine-dbal/default.nix
@@ -1,13 +1,16 @@
 {
   lib,
   config,
+  dream2nix,
   ...
 }: let
[...]

PR #594, opened 2023-07-20
Racket: Add binaries to $out/bin.

zainab-ali:
Fixes https://github.com/nix-community/dream2nix/issues/527. 

I'm still figuring out how to move this change over to the new API.

### DIFF
--- a/src/subsystems/racket/builders/simple-racket/default.nix
+++ b/src/subsystems/racket/builders/simple-racket/default.nix
@@ -96,7 +96,7 @@
           export PLTADDONDIR=$PLTCONFIGDIR;
 
           mkdir -p $out/bin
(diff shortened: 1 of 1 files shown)


DavHau:
Hey @zainab-ali we'd have to port the racket subsystem to the module system. Would you be interested in helping out with that.
As an example @aszenz did start working on doing that for php: https://github.com/nix-community/dream2nix/issues/550
Just let me know if you have any questions about it.

DavHau:
If you want to bring the fix into the legacy branch, you can also just re-target to legacy for now.

zainab-ali:
Thanks! I've retargeted for now, but will look into porting racket over in the longer term.

Issue #596, opened 2023-07-21
flake-parts module for consumers


roberth:
Hey, congrats on getting the modules into main!

I was trying to render docs for the flake-parts modules, but I couldn't figure out the separation between modules that you use internally for constructing the `dream2nix` flake, and which modules are supposed to be imported by users.
I've played around a bit, removing the `flake.modules` declaration+definition, because that didn't *seem* like something users would need, but now I think I was wrong.
Not sure what's the status of this use case; using the flake-parts module outside this flake. Maybe I can be of help?

PR #597, opened 2023-07-21
Provide flake parts module

roberth:
Experimental changes I made while trying to render docs and maybe figure out how to use it in a "consumer" flake-parts based flake.
- #596

### DIFF
--- a/modules/flake-parts/all-modules.nix
+++ b/modules/flake-parts/all-modules.nix
@@ -31,7 +31,13 @@ in {
   };
 
   # generates future flake outputs: `modules.<kind>.<module-name>`
-  config.flake.modules = lib.mapAttrs (kind: _: mapModules kind) moduleKinds;
+  config.flake.modules =
+    let modules = lib.mapAttrs (kind: _: mapModules kind) moduleKinds;
+    in modules // {
+      flake-parts = modules.flake-parts // {
+        all-modules = { imports = flakePartsModules; _class = "flake-parts"; };
+      };
+    };
 
   # comapt to current schema: `nixosModules` / `darwinModules`
   config.flake.nixosModules = config.flake.modules.nixos or {};
--- a/modules/flake-parts/writers.nix
+++ b/modules/flake-parts/writers.nix
@@ -1,5 +1,5 @@
-{
-  perSystem = {
+{ flake-parts-lib, ... }: {
+  options.perSystem = flake-parts-lib.mkPerSystemOption ({
     config,
     lib,
     pkgs,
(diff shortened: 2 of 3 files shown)


modules/flake-parts/all-modules.nix
```diff
@@ -31,7 +31,13 @@ in {
   };
 
   # generates future flake outputs: `modules.<kind>.<module-name>`
-  config.flake.modules = lib.mapAttrs (kind: _: mapModules kind) moduleKinds;
+  config.flake.modules =
```
roberth, line 34: I've assumed that this was for internal purposes, and I think it would make sense to have something like this in flake-parts for publishing modules.

It would also be nice to have an option for local modules that aren't published.

modules/flake-parts/all-modules.nix
```diff
@@ -31,7 +31,13 @@ in {
...
+    let modules = lib.mapAttrs (kind: _: mapModules kind) moduleKinds;
+    in modules // {
+      flake-parts = modules.flake-parts // {
+        all-modules = { imports = flakePartsModules; _class = "flake-parts"; };
```
roberth, line 38: Kinds are actually called `class` in `evalModules` or `_class` in the module syntax.
It's a fairly new thing, not too long before 23.05. https://nixos.org/manual/nixpkgs/stable/#module-system-lib-evalModules-param-class

Haven't used `class` much yet as it's incompatible with 22.11, but that's increasingly no excuse anymore ;)

modules/flake-parts/writers.nix
```diff
@@ -1,5 +1,5 @@
-{
-  perSystem = {
+{ flake-parts-lib, ... }: {
+  options.perSystem = flake-parts-lib.mkPerSystemOption ({
```
roberth, line 2: This makes the options statically available for rendering.

modules/flake-parts/writers.nix
```diff
@@ -9,10 +9,25 @@
   in {
     options.writers = {
       writePureShellScript = lib.mkOption {
-        type = lib.types.functionTo lib.types.anything;
+        type = lib.types.functionTo (lib.types.functionTo lib.types.package);
```
roberth, line 12: I've moved the documentation here, but probably the option declarations should be moved closer to the writers.

Sort-of like this:

```nix
options.writers = lib.mkOption { type = submodule ../writers.nix; }
```

DavHau:
Merged the first commit via https://github.com/nix-community/dream2nix/pull/597
Thanks

PR #605, opened 2023-07-27
expose writers as drv-parts module...

phaer:
...and don't manually include it per module where it's used (pip and nodejs-package-json).

I think that's useful as we add more modules, to keep `deps` in consuming modules clean.

Might be a good idea to also de-duplicate the options with the flake-parts module? the callPackage invocation is the only difference between the flake-parts and drv-parts module

### DIFF
--- a/modules/drv-parts/writers/default.nix
+++ b/modules/drv-parts/writers/default.nix
@@ -1,0 +1,34 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  writers = config.deps.callPackage ../../../pkgs/writers {};
+in {
+  imports = [
+    ./interface.nix
+  ];
+
+  config.deps = {nixpkgs, ...}:
+    lib.mapAttrs (_: lib.mkOverride 1001) {
+      inherit
+        (nixpkgs)
+        callPackage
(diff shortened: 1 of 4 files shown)


modules/drv-parts/writers/default.nix
```diff
@@ -0,0 +1,34 @@
+{
+  config,
+  lib,
+  ...
+}: let
+  writers = config.deps.callPackage ../../../pkgs/writers {};
+in {
+  imports = [
+    ./interface.nix
+  ];
```
DavHau, line 10: I think it would be nice if this would import `dream2nix.modules.drv-parts.core` and thereby become usable as a drv-parts submodule like all our other drv-parts.

So basically instead of defining `config.writers`, the module could define options for the actual interface of the writer and `config.public` for the result.

phaer: A right, that's a good idea. Just did this quickly while working on a third module including writers. 

Later remembered that I still got an un-rebased https://github.com/nix-community/dream2nix/pull/548 lying around. Will merge those to and use the core interface.

PR #525, opened 2023-06-06
rust: add support for workspace-inherited crate versions

71:
Adds support for [`workspace.package.version`](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) in the root `Cargo.toml` and `package.version.workspace = true` in child `Cargo.toml` files.

Not sure how much we're expected to test things. A lot of things are breaking on my Nix setup (I'm running in codespaces which don't really support Nix), so I wasn't able to test everything. Nonetheless, the following worked:

```sh
$ nix flake show $(realpath .)/examples/rust_workspace-version-inheritance --override-input dream2nix $(realpath .)
...
├───apps
│   └───x86_64-linux
│       └───detect-projects: app
├───devShells
│   └───x86_64-linux
│       ├───app: development environment 'devshell'
│       └───default: development environment 'devshell'
└───packages
    └───x86_64-linux
        ├───app: package 'app-0.1.0'
        ├───default: package 'app-0.1.0'
        └───resolveImpure: package 'resolve'

$ nix run $(realpath .)/examples/rust_workspace-version-inheritance --override-input dream2nix $(realpath .)
...
hello from v0.1.0

$ nix run '.#detect-projects' $(realpath .)/examples/rust_workspace-version-inheritance
# projects.toml file describing inputs for dream2nix
#
# To re-generate this file, run:
#   nix run .#detect-projects $source
# ... where `$source` points to the source of your project.
#
# If the local flake is unavailable, alternatively execute the app from the
# upstream dream2nix flake:
#   nix run github:nix-community/dream2nix#detect-projects $source

[""]
name = ""
relPath = ""
subsystem = "rust"
translator = "cargo-lock"
translators = [ "cargo-lock", "cargo-toml",]

["".subsystemInfo]
workspaceMembers = [ "app",]
[["".subsystemInfo.crates]]
name = "app"
relPath = "app"
version = "0.1.0"
``` 

AFAIK the name of the workspace being empty happens regardless of my PR.

PR #606, opened 2023-07-29
WIP: Port rust subsystem to using drv-parts

DavHau:
@yusdacra I currently don't have much time to proceed with this. If you could take it on from here, it would be great.

- ripgrep already builds.
- I backported https://github.com/nix-community/dream2nix/pull/525 and created an example, but the build currently fails. Maybe it's just because a different builder is used, but I'm not sure.


yusdacra:
I ported the crane builder since it's the more complete and better working one, builds both of the rust examples right now. Still have to look at why buildRustPackage fails at building that one example but I think I can take a look at it later. Is there anything else specifically you want me to do? I still have to add the devshell back, and probably some options to the crane builder to make doing some stuff easier (mainly setting cargo flags and whatnot).

yusdacra:
added a few options for setting flags and profiles, and disabling tests

yusdacra:
i added two new options for configuring the main derivation and the dependencies derivation, i wanted to use the `mkDerivation` module for those options but i couldnt get it to work properly so just went with a module with a freeform type declared for now.

DavHau:
I pushed some changes to make the crane module use mkDerivation for the top level and dependency derivation. Does this look alright to you?

yusdacra:
> I pushed some changes to make the crane module use mkDerivation for the top level and dependency derivation. Does this look alright to you?

Yes, thats what I've wanted to do aswell. I'll test it soon and see if it's working as intended.

Issue #601, opened 2023-07-26
Consider pex instead of pip for locking


phaer:
We are currently using `pip install --dry-install --report` in `fetchPipMetadata` to create our python lock files. While this works better than earlier FOD-based approaches for many use-cases, it has (at least) to major issues:

* cross-platform locking: We can't currently use that approach to e.g. create aarch64-darwin lock files on x86_64-linux, because we can't easily fake platform environment markers, see https://github.com/pypa/pip/issues/11664#issue-1500189934
* evaluation of extras & environment markers currently happens at lock-time, not at build-time. While this approach was easier to implement because we could to it in python, it means that
  * the lock file becomes much larger as we need to lock each platform-extra combination ahead of time
  * it can result in incorrect cross-platform locks for legacy sdists which use setup.py. This is because we would need to prepare metadata by executing the setup.py on the build-platform. If the code in setup.py introspects the interpreter to determine the platform that could result in wrong metadata being output.

https://github.com/nix-community/dream2nix/issues/583#issuecomment-1650443589 recommended https://pex.readthedocs.io as an alternative to pip to tackle the first problem (and support cyclic dependencies, as that's what the linked issue was about). 

`pex` lock files include unevaluated enviroment markers,same as `pip`s installation report, i.e. `"brotli>=1.0.9; platform_python_implementation == \"CPython\" and extra == \"brotli\"",`. So we would need to evaluate those in nix. Prior work for that includes `poetry2nix` [pep425.nix](https://github.com/nix-community/poetry2nix/blob/master/pep425.nix).

What do you think, is that something we should try? Would obsolete my work in https://github.com/phaer/untangled_snakes/ but I am happy about that as long as it improves our python support. 

@DavHau @chaoflow

jsirois:
> pex lock files include unevaluated enviroment markers,same as pips installation report, i.e. "brotli>=1.0.9; platform_python_implementation == \"CPython\" and extra == \"brotli\"",. So we would need to evaluate those in nix.

What is the transform you do at a high level? From lock file to ... venv? sys.path entry (flat directory of installed wheels) or something else? Pex has alot of tools to work with a lock file.

yajo:
Would it still work with the max-date setting that's currently supported by the pip locker? That's very handy!

phaer:
> What is the transform you do at a high level? From lock file to ... venv? sys.path entry (flat directory of installed wheels) or something else?

[...]

PR #611, opened 2023-07-31
python builder, based on pex lock files and pyproject.nix

phaer:
Hi, 

This is WIP in order to keep my work transparent as it progresses and to collect feedback from various people here interested in the python support of dream2nix.

With the known issues and limitations of the current approach of `fetchPipMetadata` regarding platform-independent lock files and evaluation of optional dependencies - and after the discussion in #601 about pex lock files and the recent release of https://adisbladis.github.io/pyproject.nix,  I got curious whether somewhat different approach based on those tools combined could work better.

Please see [modules/drv-parts/pex/README.md](https://github.com/nix-community/dream2nix/blob/pex/modules/drv-parts/pex/README.md) for details and a checklist of open issues.


chaoflow:
@phaer I love the initiative!

1. Do you see a bridge for projects that are using requirements.txt like currently supported?
2. How does the approach deal with environments consisting of multiple python distributions like https://github.com/chaoflow/dream2nix-pyenv vs single distribution projects?

phaer:
> Do you see a bridge for projects that are using requirements.txt like currently supported?

Yes. I just started out with ´requirementsList` but pretty much the whole set should be supported overtime. Regular requirements not starting with '-' should already work okay-ish? That is: no `-r`, no `-e` yet, but environment markers, extras and version constraints. The parsing is done by pyproject-nix.lib.pep508.parseString which has a unit-test suite that can expanded as needed.
For your specific use case: `--hash` isn't covered yet in pyproject-nix, that might be a good contribution! And we would probably need to do a tiny bit of cleanup (removing newlines, comments). Happy to hack together on that next time!

> How does the approach deal with environments consisting of multiple python distributions like https://github.com/chaoflow/dream2nix-pyenv vs single distribution projects?

By "multiple" vs "single distribution", do you mean to distinguish between an environment of packages as in requirements.txt files vs a clear top-level projects as in pyproject.toml files? 
If so, that's still something we need to explore bit more api-wise and probably build something on or similar to nixpkgs for a similar problem (i.e. toPythonApplication/Module). I've tried to cover that in the pex readme.

chaoflow:
@phaer For me the line is not between requirements.txt and pyproject.toml. Some nomenclature:

- **python distribution** defined by pyproject.toml or legacy setup.cfg/setup.py contains python modules or packages
- **python module** a file ending in ``.py``
- **python package** a directory containing a ``__init__.py`` file

## Single distribution

- pyproject.toml for build dependencies, install dependencies and extra dependencies. These may have constrains, like ``>=3.0`` but usually do not specify exact versions
- optional requirements.txt files recursively compiled from dependencies defined in pyproject.toml, one for each type of dependency: build, "default", each extra - note https://github.com/jazzband/pip-tools/pull/1681

For *production* we want to create a *default python environment* from this and install the package in it. For development there is usually a ``dev`` extra for which we want a *development python environment* and install the package in *editable* mode.


## Multi distribution

Several python distributions as described in "Single distribution" contained in different folders.

On top of the two or more flavours above, corresponding to the default set and extra sets of dependencies, we want to have:
- environments for individual distribution to run unit tests
- environments for combinations to run integration tests
- environment for all packages for development and production

In each of these cases, packages can be installed in editable or normal mode.


## No distribution

A python environment may also be defined by a requirements.txt/.in file without a pyproject.toml, in that case we only create an environment, but have no package to install in it.


Do you already see some API for that? Happy to do a session together to sketch this out.

phaer:
> Do you already see some API for that? Happy to do a session together to sketch this out.

Thanks for the detailed response, @chaoflow! Also happy to quickly discuss this either in todays weekly or in another call :)

I think our terminologies do match. The common unit from nix POV are still derivations, often called "packages", in our case created by "drv-parts modules".

In the context of the python subsystem, we create one "package/module (our terms) for each python "distribution" (python term). We don't deal with pythons "packages" or "modules" here, from nix points of view.

In addition to one module per distribution we also need something to bundle all of them up in a shared environment, manage the lock file and so on; And I believe thats the focal point of our discussion? We are trying two approaches here atm:

`pip` has the notion of a "top-level package", or top-level distribution. The same module that manages the lock file also calls `buildPythonPackage` and yields a single package. 
This works conceptually with "single distributions", but obviously not with multiple distributions.

`pex` in its current wip iteration uses `python.buildEnv` instead of `buildPythonPackage` in order to explore exactly the question how we should handle this question api-wise.

My current thinking is that both pip and pex should work with a unified struct representing the package set and expose options or high-level functions to then either build a pythonEnvironment from that or pick one-per-submodule to pass to `buildPythonPackage`.

webknjaz:
> **python package** a directory containing a `__init__.py` file

This is not really correct as of today. It used to be a requirement in Python 2 but in Python 3, there's implicit namespace packages. So any directory with `*.py` modules is an importable package. In fact, if you create an empty directory, you can still import it — Python 3 will not complain.

chaoflow:
> > **python package** a directory containing a `__init__.py` file
> 
> This is not really correct as of today. It used to be a requirement in Python 2 but in Python 3, there's implicit namespace packages. So any directory with `*.py` modules is an importable package. In fact, if you create an empty directory, you can still import it — Python 3 will not complain.

@webknjaz Thank you for pointing this out, I updated my comment.

phaer:
I won't continue on this for the time being as, I believe that there are no more benefits compared to PDM-based implementations. Let me know if you still have a use case for pex in this context.

PR #613, opened 2023-08-05
fetchPipMetadata: fixes

fufexan:
- ~~fetchPipMetadata: avoid deprecation warning~~
- fetchPipMetadata: compute hash for non-FOD paths

### DIFF
--- a/pkgs/fetchPipMetadata/src/fetch_pip_metadata/lock_file_from_report.py
+++ b/pkgs/fetchPipMetadata/src/fetch_pip_metadata/lock_file_from_report.py
@@ -79,10 +79,20 @@ def lock_info_from_path(full_path):
     if drv_json:
         return lock_info_from_fod(store_path, drv_json)
     else:
-        raise Exception(
-            f"fatal: requirement '{full_path}' refers to something we "
-            "can't understand"
+        proc = subprocess.run(
+            ["nix", "hash", "path", store_path],
+            check=False,
+            text=True,
+            stdout=subprocess.PIPE,
         )
+        if proc.returncode == 0:
+            sha256 = proc.stdout
+            return "", sha256  # need to find a way to get the URL
+        else:
+            raise Exception(
+                f"fatal: requirement '{full_path}' refers to something we "
+                "can't understand"
+            )
 
 
 def lock_entry_from_report_entry(install):


phaer (changes requested):
Fod changes look good thanks. 

But the show-derivation "change breaks for nix <2.15 which we still want to support, given thats true for nix stable.

fufexan:
Oh, good point. I'll remove that change then.

phaer (approved):

phaer:
Are, we still need to integrate the merge queue and hercules-ci somehow, it seems. PRs from external forks still don't trigger ci builds yet. cc @DavHau who could overrule this and has full perms.

DavHau:
Thinking about this, I notice that we should probably better not lock URL's and hashes of store paths at all, because their presence in the store indicates that the URL and hash are already defined somewhere else and we don't want to have a second source of truth.

I remember, I was running into this problem with the dream-lock.json format where whenever there was a store path in the lock we intentionally filtered it out when parsing the lock, to not run into some weird problems.

DavHau:
Anyways, as it fixes a bug currently, I'll merge it.

DavHau:
> PRs from external forks still don't trigger ci builds yet

Is that only for first time contributors or every external fork?

phaer:
> Thinking about this, I notice that we should probably better not lock URL's and hashes of store paths at all, because their presence in the store indicates that the URL and hash are already defined somewhere else and we don't want to have a second source of truth.

Yes, that's a good point. @fufexan what was your use case here?

> Anyways, as it fixes a bug currently, I'll merge it.

Don't see how, tbh? This looks like a new "feature" insofar as that it would correctly error out without this change, no?

> Is that only for first time contributors or every external fork?

Afaiu its coupled to write permissions to this repo for security reasons, so whether it's a first-time contributions or not should not make a difference?

fufexan:
@phaer without this change, using a flake input as the `src` for `mkDerivation` errors out. @DavHau instructed me to make this change in the Matrix room.

DavHau:
see https://matrix.to/#/!PrIEtcffTLLzUWvJOS:matrix.org/$9Rv9hoW0fIdNa7p3jBK8DdbEJsrFRBiWcgh-WvYYnfM?via=nixos.org&via=matrix.org&via=nixos.dev

PR #620, opened 2023-08-21
Support git repositories in fetchPipMetadata

VincentVanlaer:
This was more or less an experiment to see if I could get this working, I'm not sure how useful this will be if the intent is to move to pex. Should there still be interest in support for git repositories in the pip builder, I'm happy to update this PR with whatever changes/tests/... are required. Feel free to close otherwise!


DavHau:
Hey that looks pretty cool. We don't know if and when the pex builder will lead to the deprecation of the pip module. I think we should continue the development of the pip moduele and your PR is very welcome.

Could you add a test for the git dependency feature? Otherwise we won't notice if we ever break the feature.

VincentVanlaer:
Is there a minimum python version this project is targeting? In the current version of the lock file script I use a match statement, but this is only available since python 3.10.

As for the test, I assume you mean adding a test to https://github.com/nix-community/dream2nix/tree/main/pkgs/fetchPipMetadata/src/tests?

DavHau:
Depending on python 3.10 is fine I think since the script can run with its own interpreter distinct from the one that is used for the pip fetching.
The test location you mentioned is the right one.

phaer (approved):

VincentVanlaer:
Added a test and rebased on main

phaer (approved):
LGTM, Thanks!

DavHau:
reverted because `nix flake check` was failing.

PR #621, opened 2023-08-22
Python: fixes and improvements

DavHau:
- adds new option flattenDependnecies (bool), to allow for dependency specifications without root (like a requirements.txt file).
- raise an informative error if flattenedDependencies is used the wrong way (like when it's set but a root with the current package name actually exists)
- adds new option ignoredDependencies to ignore dependencies. It cotains `wheel` by default, as building wheel using buildPythonPackage usually crashes due to conflicts
- pins the packaging library used to parse pip's report to the exact version that pip uses by refering to `pip._internal`
- prevents crashes when there is a cycle in the report -> instead remove the cycle

Other stuff:
- fix lib.evalModules
- fix devshell for the pip module python stuff

### DIFF
--- a/examples/dream2nix-packages-python/pyproject-dependencies/default.nix
+++ b/examples/dream2nix-packages-python/pyproject-dependencies/default.nix
@@ -1,0 +1,37 @@
+# An example package with dependencies defined via pyproject.toml
+{
+  config,
+  lib,
+  dream2nix,
+  ...
+}: let
+  l = lib // builtins;
+  pyproject = lib.pipe (config.mkDerivation.src + /pyproject.toml) [
+    builtins.readFile
+    builtins.fromTOML
+  ];
+in {
+  imports = [
+    dream2nix.modules.drv-parts.pip
+  ];
+
+  name = "pyproject-dependencies";
+  version = pyproject.project.version;
+
+  mkDerivation = {
+    src = ./.;
+  };
+
+  buildPythonPackage = {
+    format = lib.mkForce "pyproject";
+    pythonImportsCheck = [
+      "my_tool"
+    ];
+  };
+
+  pip = {
+    pypiSnapshotDate = "2023-08-27";
+    requirementsList = ["setuptools"] ++ pyproject.project.dependencies;
+    flattenDependencies = true;
+  };
+}
--- a/examples/dream2nix-packages-python/pyproject-dependencies/my_tool/__init__.py
+++ b/examples/dream2nix-packages-python/pyproject-dependencies/my_tool/__init__.py
@@ -1,0 +1,4 @@
+import requests
+
+
+print("Hello World!")
--- a/examples/dream2nix-packages-python/pyproject-dependencies/pyproject.toml
+++ b/examples/dream2nix-packages-python/pyproject-dependencies/pyproject.toml
@@ -1,0 +1,11 @@
+[build-system]
+requires = [ "setuptools" ]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "my-tool"
+description = "my tool"
+version = "1.0.0"
+dependencies = [
+  "requests"
+]
--- a/flake.lock
+++ b/flake.lock
@@ -70,6 +70,52 @@
         "type": "github"
       }
     },
+    "nix-github-actions": {
+      "inputs": {
+        "nixpkgs": [
+          "nix-unit",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1688870561,
+        "narHash": "sha256-4UYkifnPEw1nAzqqPOTL2MvWtm3sNGw1UTYTalkTcGY=",
+        "owner": "nix-community",
+        "repo": "nix-github-actions",
+        "rev": "165b1650b753316aa7f1787f3005a8d2da0f5301",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "nix-github-actions",
+        "type": "github"
+      }
+    },
+    "nix-unit": {
+      "inputs": {
+        "flake-parts": [
+          "flake-parts"
+        ],
+        "nix-github-actions": "nix-github-actions",
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "treefmt-nix": "treefmt-nix"
+      },
+      "locked": {
+        "lastModified": 1690289081,
+        "narHash": "sha256-PCXQAQt8+i2pkUym9P1JY4JGoeZJLzzxWBhprHDdItM=",
+        "owner": "adisbladis",
+        "repo": "nix-unit",
+        "rev": "a9d6f33e50d4dcd9cfc0c92253340437bbae282b",
+        "type": "github"
+      },
+      "original": {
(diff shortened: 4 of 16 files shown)


phaer (commented):
While those are undoubtedly pragmatic and well-meant changes to get *something* to build, I personally still believe it's worth the effort to get the primitives right in the medium and long-term; before getting too specific with our solutions.

I think a pip-specific mechanism to ignore dependencies and flatten them is a bit too specific. Ideally we get the pex builder to a point. where we can share *some* interface between our two concepts of "python package set"?
 
It's just another layer which you have to understand before you can use it, so ideally we could split out the logic to do things like this into functions to keep the pip module lean?
 Requiring it for standard-stuff like wheel also seems to hint at a wrong abstraction and I'd prefer not to merge this stuff just yet.

I think a "flattening" option might be confusing ux-wise, as it describes it from our perspective? I planned to  differentiate between "buildPythonEnv" and "buildPythonApp" - from a users perspective. 

I hope pex can soon replace the packaging vendoring, as it exposes the libraries in question and vendors pip itself to circumvent the explicit lack of a pip API.

cycle-breaking is a good idea

DavHau:
For me, the primary goal right now is to make things work somehow, then later care about doing it nicely. If you think that this stands in the way for your plan for the python subsystem, we could potentially fork the module and develop on it independently?

dream2nix is currently broken for python projects that define their dependencies via a list of requirements, as there will not be a root in the lock file.
We've been shifting this issue already for quite some time and I think we need a fix urgently. 

Working out a better and more generic interface for dependency management is nice, but I don't think that it should block us fixing problems like that quickly.
Once we come up with a generic solution, we can still change the module to use it. After all, our API is unstable for a reason.

There is also a risk in getting the abstraction wrong if we generalize things too early. I made the mistake once before of investing in abstractions too early when designing the legacy dream-lock interface. I don't want to go there again. I think it's much better to have a few modules implementing similar things in slightly different ways instead of making an abstraction that is slightly wrong for all of them. A sloppy abstraction can be way more painful than having some code duplication. And it's going to be sloppy if w don't have at least a couple of modules that require the same thing and give us confidence that this is the right way for all of them. 

I'd really like us to go fast, experiment with stuff, break stuff fast and fix things fast instead of investing into finding the best abstractions just to notice later we have to re-invent them anyways. I think the right way to go is implementing solutions specifically for individual modules first, like pip. Then, once we notice similarities between different modules, we can start to generalize and share code. Currently pip is the only module that needs the flattening and ignoring, therefore, I think it should not belong anywhere else yet.

> Requiring it for standard-stuff like wheel also seems to hint at a wrong abstraction 

The underlying nixpkgs builder simply crashes when you try to build the `wheel` package. It's a bug in dream2nix that it tries to build wheel at all, when it's already part of the build env and I want to fix it somehow. Should I remove the ignore option and instead hardcode ignoring the wheel package?

> I think a "flattening" option might be confusing ux-wise, as it describes it from our perspective? I planned to differentiate between "buildPythonEnv" and "buildPythonApp" - from a users perspective.

If flattening is required or not doesn't depend on the user building an `app` or an `env`. It just depends on the way the requirements are defined. I simply want to fix the case where there is no root defined for the dependency tree which is a common problem, like for example when a `requirements.txt` is used.

We can change the interface of the pip module in some other way to cover that possibility, but I think we should cover it somehow.

phaer:
> dream2nix is currently broken for python projects that define their dependencies via a list of requirements, as there will not be a root in the lock file.

Yes, but that has been true for a very long time - in fact I don't think it has ever supported this use-case correctly so far, that's where my preference to take it slow here comes from :)

>  We've been shifting this issue already for quite some time and I think we need a fix urgently.

Where does the urgency stem from? Neither repo nor docs claim that anything of this is ready yet and much of the delay came from the (IMO ongoing) overhaul of drv-parts APIs and the switch from pex. The latter is just a few weeks old.

> There is also a risk in getting the abstraction wrong if we generalize things too early. I made the mistake once before of investing in abstractions too early when designing the legacy dream-lock interface.

Answering the whole paragraph here: I agree on that starting with an abstract interface would be the wrong approach, but I don't think that's the case here: We've got the very same concept of python packages & package sets both in the pex and in the pip builder; They already share much of the same logic and and are much more specific than the dream2nix lock file thingy.

> I'd really like us to go fast, experiment with stuff, break stuff fast and fix things fast instead of investing into finding the best abstractions just to notice later we have to re-invent them anyways.

Fine with that in principle, as long as it does not bloat the implementation with ad-hoc fixes which are not nicely composable in the current API. If you can split them in a separate module and just include e.g. a "python-hotfixes" module or so, I am fine with that. If you keep it in sync with the pex builder, that could also work. 

> If flattening is required or not doesn't depend on the user building an app or an env. It just depends on the way the requirements are defined.

That's pretty much the same thing in my understanding. An "app" here would be a specific package you care about, where the dependencies are just baggage you need to run that package; An "env" here would be a flat list of packages, without a top-level file as in e.g. requirements.txt. 



> We can change the interface of the pip module in some other way to cover that possibility, but I think we should cover it somehow.

I also agree, else I would not have thought about how to implement that properly.

DavHau:
> in fact I don't think it has ever supported this use-case correctly so far

It did work fine with the mach-nix-xs module. I built complex projects like nixified-ai with it, but it broke when I updated my expression to use the new pip module.

> Where does the urgency stem from?

I'd like to use dream2nix to build packages successfully. 

> Neither repo nor docs claim that anything of this is ready yet

It claims the software is unstable, which doesn't necessarily mean it is broken. I think ensuring support for common use cases should be a priority.

> We've got the very same concept of python packages & package sets both in the pex and in the pip builder; They already share much of the same logic

Could you merge your pex module already (maybe named as `WIP-pex/default.nix`)? Or alternatively merge at least the shared abstractions that you intend to have between the pip and pex. This way I could nicely test and see if and how my changes interact with your changes and it would become obvious to me if a fix is specific to some module or not.

> Fine with that in principle, as long as it does not bloat the implementation with ad-hoc fixes which are not nicely composable in the current API

From the perspective of the current version of dream2nix, pip is the only module for python and there is no composability to break. Any change to the pip module isn't more or less ad-hoc than any other code that is already in the pip module before that change.

> If you can split them in a separate module and just include e.g. a "python-hotfixes" module or so

Currently those fixes are not used anywhere outside of the pip module, therefore putting it anywhere else than in the pip module seems odd to me. It could simply be factored out in the future once it is actually used somewhere else.

Anyways I just pushed some changes where I put the logic into an extra module `pip-hotfixes` hoping that this will make your work easier. I also added an example package using the feature and added nix unit tests, to allow for safe refactoring later.

>  If you keep it in sync with the pex builder, that could also work.

I'm happy to keep stuff in sync that is in main. Testing my changes against other WIP branches is too time consuming and also hard to automate.

> An "app" here would be a specific package you care about, where the dependencies are just baggage you need to run that package; An "env" here would be a flat list of packages, without a top-level file as in e.g. requirements.txt.

Having an env only module could make sense, but still there are apps that use the `env-style` way of defining dependencies, therefore we cannot strictly separate these concepts. They need to be combined somehow.

I'll merge this now but am happy to have the approach changed once a better solution comes up.

PR #637, opened 2023-09-03
Support git repositories in fetchPipMetadata; take 2

VincentVanlaer:
See #620 and #630 

Changes since the previous version:

- Ensured that the pip lock script uses the python version from nixpkgs, not the python version the user wants to lock the packages for
- A fix for the htop package failing to build (unrelated to these changes)
- Lockfiles for the various example repositories have been updated. This is what caused the CI to fail after merge.

I checked these changes with `nix flake check` which did not report any errors, but that took ~24GB of peak memory usage to complete. Is that something that is known/expected, or is something odd with my config?


DavHau:
Thanks a lot for this!

PR #649, opened 2023-09-05
feat(rust-crane): add devshell to packages

yusdacra:
- feature: add devshell to rust-crane packages
- fix: add patch workspace deps to buildRustPackage

### DIFF
--- a/modules/dream2nix/buildRustPackage/patch-workspace-deps.jq
+++ b/modules/dream2nix/buildRustPackage/patch-workspace-deps.jq
@@ -1,0 +1,48 @@
+def normalizeWorkspaceDep:
+  if ($workspaceDependencies."\(.key)" | type) == "object"
+  then [.value, $workspaceDependencies."\(.key)"] | add
+  else [.value, {"version":$workspaceDependencies."\(.key)"}] | add
+  end
+  # remove workspace option from the dependency
+  | del(.workspace)
+;
+
+# normalizes workspace inherited dependencies for one list
+def mapWorkspaceDepsFor(name):
+  if has(name)
+  then
+    ."\(name)" = (
+      ."\(name)"
+      | to_entries
+      | map(
+        if (.value | type) == "object" and .value.workspace == true
+        then .value = (. | normalizeWorkspaceDep)
+        else .
+        end
+      )
+      | from_entries
+    )
+  else .
+  end
+;
+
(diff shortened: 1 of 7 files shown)


modules/dream2nix/rust-crane/devshell.nix
```diff
@@ -76,6 +78,7 @@
     // {
       inherit name;
       passthru.env = _shellEnv;
+      nativeBuildInputs = _shellEnv.nativeBuildInputs ++ [cargo];
     };
 in
-  (mkShell.override {stdenv = (l.head drvs).stdenv;}) shellEnv
+  (mkShell.override {stdenv = (l.head drvs).config.mkDerivation.stdenv or stdenv;}) shellEnv
```
DavHau, line 84: I think the option `config.mkDerivation.stdenv` does not exist.

modules/dream2nix/rust-crane/default.nix
```diff
@@ -207,6 +206,12 @@ in {
   package-func.args = l.mkMerge [common buildArgs];
 
   public = {
+    devShell = import ./devshell.nix {
+      name = "${pname}-devshell";
+      drvs = [cfg.depsDrv.public config.public];
```
DavHau, line 211: Ich think it would be better to pass `mainDrv` and `depsDrv` as separate arguments instead of `drvs`;

modules/dream2nix/buildRustPackage/patch-workspace-deps.jq
```diff
@@ -0,0 +1,48 @@
+def normalizeWorkspaceDep:
```
DavHau, line 1: Where is this file used?

yusdacra: in `vendor.nix`. this should share the same file in the `rust-crane` folder but i didn't know where to put shared files between modules, so i kept it like this

modules/dream2nix/rust-crane/default.nix
yusdacra: done

modules/dream2nix/rust-crane/devshell.nix
yusdacra: done, i am pulling it from `mainDrv.out.stdenv` now instead, i missed that when i first wrote that i think

PR #666, opened 2023-09-11
feat(pip): allow pypiSnapshotDate to be null

DavHau:
For many users not having a snapshot date is a valid use case.
Also many users will expect their lock file to be updated to the latest dependency versions via a nix run .#package.config.lock.refresh. so it can be the default behavior.


phaer:
Thanks! This has been overdue in hindsight. Agree that it's better UX this way :)

PR #670, opened 2023-09-11
feat: community overrides

DavHau:
- Add community overrides accessible via dream2nix.overrides.{lang}.{pname}
- make the pip module use the overrides for dependencies

### DIFF
--- a/modules/flake-parts/overrides.nix
+++ b/modules/flake-parts/overrides.nix
@@ -1,0 +1,13 @@
+{self, ...}: let
+  overridesDir = self + "/overrides";
+in {
+  flake.overrides =
+    builtins.mapAttrs
+    (
+      category: _type:
+        builtins.mapAttrs
+        (name: _type: overridesDir + "/${category}/${name}")
+        (builtins.readDir (overridesDir + "/${category}"))
+    )
+    (builtins.readDir overridesDir);
+}
--- a/examples/packages/single-language/python-machine-learning/default.nix
+++ b/examples/packages/single-language/python-machine-learning/default.nix
@@ -3,11 +3,7 @@
   config,
   lib,
   ...
-}: let
-  pyproject =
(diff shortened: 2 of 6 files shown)


modules/flake-parts/overrides.nix
```diff
@@ -0,0 +1,13 @@
+{self, ...}: let
+  overridesDir = self + "/overrides";
+in {
+  flake.overrides =
+    builtins.mapAttrs
+    (
+      category: _type:
+        builtins.mapAttrs
+        (name: _type: overridesDir + "/${category}/${name}")
```
phaer, line 9: What would you think of adding a string for version constraints? can be a noop for the moment, i.e. "category.name.version". Version could be "*" for now, and could later parse pythons package constraints (implemented in pyproject.nix)

DavHau: I've been playing with similar ideas with @adisbladis at nixcon as well.
We figured that everything can be achieved with mkIf, for example:
```nix
{config, lib, ...}:
lib.mkMerge [
  (mkIf (lib.matchesSemVer config.version "^2.0.0") {
    mkDerivation.patches = [patch1];
  })
  (mkIf (lib.matchesSemver config.version "^3.0.0") {
    mkDerivation.patches = [patch2];
  })
]
```
This approach offers good flexibility. We could also use other package attributes for the conditions.
Currently don't see many benefits justifying a special structure specifically around version constraints. It might be slightly more performant (if even), but is less flexible

DavHau: It's not even strictly  necessary to use mkIf. This would also work:
```nix
{config, lib, ...}: {
  mkDerivation.patches =
    if lib.matchesSemVer config.version "^2.0.0"
    then [patch1]
    else if lib.matchesSemVer config.version "^3.0.0"
    then [patch2]
}
```

PR #674, opened 2023-09-12
refactor(rust-crane): separate main derivation config into mainDrv option

yusdacra:
- separates main derivation config to `config.rust-crane.mainDrv` option
- remove `mkDerivation` options from top level
- introduce `config.rust-crane.source` for setting the source root (this also helps differentiate `mkDerivation.src` since this can also hold workspace source)

### DIFF
--- a/modules/dream2nix/rust-cargo-lock/interface.nix
+++ b/modules/dream2nix/rust-cargo-lock/interface.nix
@@ -15,7 +15,8 @@ in {
     source = {
       type = t.either t.path t.package;
       description = "Source of the package";
-      default = config.mkDerivation.src;
+      # pull default either from rust-crane source or mkDerivation source (mkDerivation is what buildRustPackage uses)
+      default = config.rust-crane.source or config.mkDerivation.src;
     };
   };
 }
--- a/examples/packages/single-language/rust-package/default.nix
+++ b/examples/packages/single-language/rust-package/default.nix
@@ -16,21 +16,26 @@
   name = lib.mkForce "ripgrep";
   version = lib.mkForce "13.0.0";
 
-  mkDerivation = {
+  rust-crane = {
     # define the source root that contains the package we want to build.
-    src = config.deps.fetchFromGitHub {
+    source = config.deps.fetchFromGitHub {
       owner = "BurntSushi";
       repo = "ripgrep";
       rev = config.version;
       sha256 = "sha256-udEh+Re2PeO3DnX4fQThsaT1Y3MBHFfrX5Q5EN2XrF0=";
     };
-    # note: any more options defined here will be applied to both
-    # `config.mkDerivation` and `config.depsDrv.mkDerivation`.
-  };
-
-  rust-crane = {
+    # define additional options, such as build profile and flags
+    # these can modify both derivations (main and dependencies)
     buildProfile = "dev";
     buildFlags = ["--verbose"];
     runTests = false;
+    # options defined here will only apply to the resulting (`config.public`) derivation (main derivation).
+    mainDrv = {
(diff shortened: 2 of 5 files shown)


yusdacra:
I'm not sure why but the `docs` fail to build cause of a `rust-crane.depsDrv.deps.stdenv is used but not defined`. The `docs` of `depsDrv` and `mainDrv` build fine separately however. I tried adding `stdenv` to that but it did not fix anything. Afterwards I tried adding a `stdenv` to the `deps` in `core/docs/default.nix` and that for some reason gets rid of the error (but another error pops up). Do you have any idea what's wrong with this @DavHau ?

modules/dream2nix/rust-cargo-lock/interface.nix
```diff
@@ -15,7 +15,8 @@ in {
     source = {
       type = t.either t.path t.package;
       description = "Source of the package";
-      default = config.mkDerivation.src;
+      # pull default either from rust-crane source or mkDerivation source (mkDerivation is what buildRustPackage uses)
+      default = config.rust-crane.source or config.mkDerivation.src;
```
DavHau, line 19: default values that depend on other config values might be problematic when generating documentation. Could you move this to rust-cargo-lock/default.nix and use `mkDefault` instead?

DavHau (commented):
> separates main derivation config to config.rust-crane.mainDrv option

Why is that better? With all other modules, the top-level config defines the top-level derivation. What's special about rust that requires this change?

DavHau:
> I'm not sure why but the `docs` fail to build cause of a `rust-crane.depsDrv.deps.stdenv is used but not defined`. The `docs` of `depsDrv` and `mainDrv` build fine separately however. I tried adding `stdenv` to that but it did not fix anything. Afterwards I tried adding a `stdenv` to the `deps` in `core/docs/default.nix` and that for some reason gets rid of the error (but another error pops up). Do you have any idea what's wrong with this @DavHau ?

I don't know why exactly right now. But I know that the whole dependency management is kind of annoying and I plan to refactor this soon.
It's still weird that it would still throw the error even when you set `rust-crane.depsDrv.deps.stdenv` directly.

yusdacra:
> > separates main derivation config to config.rust-crane.mainDrv option
> 
> 
> 
> Why is that better? With all other modules, the top-level config defines the top-level derivation. What's special about rust that requires this change?

The reason I want it to have it's own option is just for clarity and nothing more, IMO it's more self-explanatory this way as opposed to documenting this. This would prevent people from getting confused about why their config isn't applying properly as the top-level config would only apply to the main derivation and not the dependencies derivation, potentially causing issues (and I have seen this happen first-hand so that is also why I'm more inclined towards this).

DavHau:
> This would prevent people from getting confused

OK, seems reasonable. We can try it that way. It's just that its going to be out of sync with the other dream2nix modules. It introduces an inconsistency where some dream2nix modules have their main derivations options at the top-level and some not.

But I get the point that is is generally not clear if the top-level settings affect all dependencies or just the main one.
Maybe it will become clearer as soon as we have an `overrides` option, where you would override packages by name, for example:

```nix
{
  overrides.{package-name}.mkDerivation = ...;
}
```

package-name could either be the main package name or the name of a dependency.
Would that make it any better?

yusdacra:
That would be great actually. This problem could would also apply when we introduce a more granular builder for Rust (or some other languages too I suppose) people would want to configure each of the package dependencies separately. That would be a good solution. For this module in specific I imagine it would export the deps derivation as `${package-name}-deps` and then you could configure it via the `overrides`, and the main derivation in the same way.

DavHau:
Closing, as we plan to solve this with overrides

PR #677, opened 2023-09-13
fix: make python3 dep just a default

yajo:
When trying to build a package that had `python3 = nixpkgs.python311`, I got an error similar to:

```
       Definition values:
       - In `/nix/store/...-source/nix/modules/dream2nix/default': <derivation python3-3.11.4>
       - In `/nix/store/...-source/modules/dream2nix/core/paths': <derivation python3-3.10.12>
       Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.»; }
```

This seems like a sane usage, so I'm expressing here that what dream2nix adds is just a default.

### DIFF
--- a/modules/dream2nix/core/paths/default.nix
+++ b/modules/dream2nix/core/paths/default.nix
@@ -8,7 +8,7 @@
     ../deps
   ];
   deps = {nixpkgs, ...}: {
-    python3 = nixpkgs.python3;
+    python3 = lib.mkDefault nixpkgs.python3;
     substituteAll = nixpkgs.substituteAll;
   };
   paths = {


yajo:
Oh, it seems I got confused and should have added `python` as the dependency, not `python3`: https://github.com/nix-community/dream2nix/blob/77eeb06927e94bc196444fcee2d08a20514443a3/modules/dream2nix/pip/default.nix#L9

Maybe this could be closed then.

DavHau:
Hm, maybe we should refactor the whole thing to use `python3`, not `python` what do you think?

yajo:
I think it's better to use `python`, not `python3`, because:
- You may want to build a python2 package.
- Some day, you may want to build a python4 package.

So, the fact you use `python` for the pip backend is better IMHO.

The fact you're using `python3` for path module also makes sense, because it's the supported version: https://github.com/nix-community/dream2nix/blob/5a729cb75ae914b2e97114e1c0e9280953b7b488/modules/dream2nix/core/paths/find-root.py#L1

So, IMHO the better fix would be to package `find-root.py` separately, so you add *that* as a dependency, and not its python interpreter directly. Then we only have one `python` variable around, and it's the one for building python packages.

DavHau:
@mergify rebase

DavHau:
I can't rebase your PR's. Mind giving me permissions?

yajo:
Sorry, I can't. Github doesn't allow that when the fork belongs to a team, like this one. Let me rebase it.

phaer:
I think this is fixed?

On current main branch, python3 gets a mkOverride-priority of 1003, which is even a bit higher than mkDefault - so can be easily overridden with a simple setting (with or without mkDefault/mkForce)

PR #681, opened 2023-09-14
refactor(lock): introduce mkLazy

DavHau:
This seems like a cleaner way to make the value for lock.content more lazy and not trigger errors because paths.XXX is not set in unit tests


phaer (approved):
LGTM

PR #691, opened 2023-09-18
fix: give an appropriate hint to refresh lock with flakes

yajo:


DavHau:
Since https://github.com/nix-community/dream2nix/pull/671, simply using `.lock` should be sufficient. Can you confirm?

yajo:
Yikes, you're right. I was running `.#whatever.config.lock`, not `.#whatever.lock`. Sorry! 🙈

DavHau:
No worries and thanks for your contributions.

PR #692, opened 2023-09-18
fix: improve handling of local python dependencies

yajo:
The pip locker [sometimes created a URL without hash][1] to reflect that there was a local source.

This produced the following error when trying to evaluate that derivation:

       … from call site

         at /nix/store/qdzdi6qvrqxfqxhi3j70d5dsl9f5jpby-source/modules/dream2nix/pip/default.nix:86:28:

           85|       mkDerivation = {
           86|         src = l.mkDefault (fetchers.${metadata.sources.${config.name}.type} metadata.sources.${config.name});
             |                            ^
           87|         doCheck = l.mkDefault false;

       … while calling 'url'

         at /nix/store/qdzdi6qvrqxfqxhi3j70d5dsl9f5jpby-source/modules/dream2nix/pip/default.nix:57:11:

           56|   fetchers = {
           57|     url = info: l.fetchurl {inherit (info) url sha256;};
             |           ^
           58|     git = info: config.deps.fetchgit {inherit (info) url sha256 rev;};

       error: value is null while a string was expected

Of course, one cannot call `builtins.fetchurl` without a `sha256` argument.

Now, local dependencies are specified exactly like that in the lock file. Now, we'll fetch nothing when dealing with local dependencies.

Also, their evaluation when generating `preFixup` attribute is delayed until build time. This was a bigger problem due to the previous bug, but in any case it should still help to reduce evaluation overhead.

FWIW, sometimes local sources might not really exit in the derivation source tree until build time. For example, when building aggregated sources with meta-repo management tools such as Mr. Chef.

[1]: https://github.com/nix-community/dream2nix/blob/40b65e4598e249af6cc285259c1c0afefbb0f420/pkgs/fetchPipMetadata/src/fetch_pip_metadata/lock_file_from_report.py#L72C28-L72C28

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 result
 interpreter
 __pycache__
+*.egg-info
--- a/examples/packages/single-language/python-local-dependencies/default.nix
+++ b/examples/packages/single-language/python-local-dependencies/default.nix
@@ -1,0 +1,39 @@
+{
+  config,
+  lib,
+  dream2nix,
+  ...
+}: let
+  pyproject = lib.importTOML ./subpkg1/pyproject.toml;
+  buildWithSetuptools = {
+    buildPythonPackage.format = "pyproject";
+    mkDerivation.buildInputs = [config.deps.python.pkgs.setuptools];
+  };
+in rec {
+  imports = [
+    dream2nix.modules.dream2nix.pip
+    buildWithSetuptools
+  ];
+
+  inherit (pyproject.project) name version;
+
+  mkDerivation.src = lib.concatStringsSep "/" [
+    config.paths.projectRoot
+    config.paths.package
+    "subpkg1"
+  ];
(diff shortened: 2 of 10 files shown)


DavHau:
Oh I remember having an issue because of that missing hash.
Do you think you could add an eval test under `./tests/nix-unit/` that reproduces the issue you had and succeeds because of the fix?

yajo:
I added it as an example because it was simpler to me to just build a full derivation and assert it succeeds than doing a expr/expected test. But this one should be ready IMHO.

DavHau:
OK, looks good. Could you run `nix flake check` and make sure all errors are fixed? Thanks

yajo:
I wasn't able to finish checking because I'd need 2 extra hard drives 😆.

However, until where I reached, it was working now.

PR #699, opened 2023-09-22
feat: allow specifying env for fetchPipMetadata

yajo:
This helps for configuring pip with environment variables. See https://github.com/nix-community/dream2nix/discussions/697 for more context.

### DIFF
--- a/modules/dream2nix/pip/default.nix
+++ b/modules/dream2nix/pip/default.nix
@@ -117,12 +117,11 @@ in {
     l.mapAttrs (_: l.mkOverride 1002) {
       # This is imported directly instead of depending on dream2nix.packages
       # with the intention to keep modules independent.
-      fetchPipMetadataScript = import ../../../pkgs/fetchPipMetadata/script.nix {
-        inherit lib;
-        inherit (cfg) pypiSnapshotDate pipFlags pipVersion requirementsList requirementsFiles nativeBuildInputs;
-        inherit (config.deps) writePureShellScript nix;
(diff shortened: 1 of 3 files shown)


modules/dream2nix/pip/default.nix
```diff
@@ -117,12 +117,11 @@ in {
     l.mapAttrs (_: l.mkOverride 1002) {
       # This is imported directly instead of depending on dream2nix.packages
       # with the intention to keep modules independent.
-      fetchPipMetadataScript = import ../../../pkgs/fetchPipMetadata/script.nix {
-        inherit lib;
+      fetchPipMetadataFunction = nixpkgs.callPackage ../../../pkgs/fetchPipMetadata/script.nix;
+      fetchPipMetadataArgs = {
```
yajo, line 121: This split between function and args is because I've not found a way to get the "super" definition of a dep.

It seem due to the modular nature of dream2nix, but something like this yields an infinite recursion:

```nix
{config, ...}: {
  deps = {nixpkgs, ...}:
    fetchPipMetadataScript = config.deps.fetchPipMetadataScript.override {env.PIP_FIND_LINKS = "${config.deps.setuptools-odoo.out.dist}";};
  };
}
```

If there were a way to do that, just wrapping the script with `callPackage` would be enough.

DavHau:
Thanks I fixed this up and merged it over here: https://github.com/nix-community/dream2nix/pull/726

I'm worried that using callPackage will make it harder to override dependencies. We should go through `deps` as much as possible.

PR #709, opened 2023-09-29
fix: nixpkgs follows d2n

waalge:
hows this?

#nixcamp


DavHau:
We don't want flake.lock files in our examples, so I removed it. Thanks for the fix.

PR #712, opened 2023-09-29
Spago support

aciceri:


modules/dream2nix/WIP-spago/default.nix
```diff
@@ -0,0 +1,137 @@
...
+      config.deps.yq-go
+    ];
+    buildInputs = [
+      config.deps.nodejs
+    ];
+    buildPhase = ''
+      export HOME="$(realpath .)"
+      mkdir -p "$HOME/.cache/spago-nodejs"
+      ln -s ${registry} "$HOME/.cache/spago-nodejs/registry"
+      ln -s ${registry-index} "$HOME/.cache/spago-nodejs/registry-index"
```
waalge, line 85: Maybe we can get rid of those

DavHau: I checked and we cannot

PR #714, opened 2023-10-01
pdm: add non-default groups as well

FRidh:


DavHau:
I think for the integration test to use multiple groups we'd need something to consume those groups like a top-level package or a dev shell. Alternatively we could just add unit tests similar to `tests/nix-unit/test_groups` testing if the pdm module populates stuff correctly. Maybe that's cleaner after all.

DavHau (approved):

DavHau:
Oh and you can build the current pdm example via:
`nix build .#checks.x86_64-linux.example-repo-flake-pdm -L`
and it is also built via `nix flake check`

PR #717, opened 2023-10-05
Fix small typos relating to pdm

purepani:
Fixes typo in the override type of pdm.
Fixes lack of pyproject.nix input, which the pdm module uses.

### DIFF
--- a/examples/dream2nix-repo-flake-pdm/flake.lock
+++ b/examples/dream2nix-repo-flake-pdm/flake.lock
@@ -1,0 +1,143 @@
+{
+  "nodes": {
+    "dream2nix": {
+      "inputs": {
+        "flake-compat": "flake-compat",
+        "nixpkgs": "nixpkgs",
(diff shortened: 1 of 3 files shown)


DavHau:
Thanks, I merged the type fix via https://github.com/nix-community/dream2nix/pull/725.
The flake input in the top-level should not be required, as the `/modules` have their own flake

purepani:
Thanks for merging it quickly; I just didn't notice.
The reason I added the input in the top level was because I was getting an error saying that pyproject-nix wasn't available. ```
 error: attribute 'pyproject-nix' missing

       at /nix/store/60q3vj251hb1p5k7dkxyxh4hp0akpzdk-source/modules/dream2nix/WIP-python-pdm/default.nix:11:26:

           10|
           11|   libpyproject = import (dream2nix.inputs.pyproject-nix + "/lib") {inherit lib;};
             |                          ^
           12|

```
I suppose the issue is with the import at that line then?

PR #727, opened 2023-10-09
Fix spago lock script

aciceri:
@DavHau I picked this up again and realized that recreating the lockfile then the build failed because now [package-set](https://github.com/purescript/package-sets/) is too new (we were always fetching `master`).
Unfortunately that repo doesn't have tags that includes package set versions (like `42.1.0`) so now the entire repo is fetched instead of the tarball and then I checkout the correct branch which is discovered checking commit messages.

There could be better ways but I believe it's enough for the lock script.

### DIFF
--- a/modules/dream2nix/WIP-spago/default.nix
+++ b/modules/dream2nix/WIP-spago/default.nix
@@ -113,9 +113,11 @@ in {
     ]
     ''
       set -euo pipefail
-      mkdir $TMPDIR/package-sets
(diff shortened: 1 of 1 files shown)


DavHau:
@mergify rebase

DavHau:
How about just downloading the required revision via tarball?
```
curl -fL https://github.com/purescript/package-sets/archive/refs/heads/$rev.tar.gz | tar xz --strip-components=1
```

aciceri:
Yeah unfortunately that solution isn't possible here.

PR #737, opened 2023-10-12
Feat/graph utils

hsjobeki:
Adds some graph utilities.
(Renamed `findCycles.nix` into `graphUtils.nix`)  

- SanitizeGraph returns a nix buildable dependency tree. (Even if the previous graph has cycles)
- Compat functions that allow backwards compatibility for potential use cases in V1 (node granularBuilder)
- 

Still TODO: Fix the old cycle remover, so it supports 

Multiple connected cycles in the graph 

e.g.

```
a->c
b->c
c->d
d->a
d->b
```


DavHau:
@mergify rebase

DavHau:
@mergify rebase

PR #742, opened 2023-10-16
fix(pdm): Make package dependencies lowercase to be consistent with pdm lock

purepani:
Package dependencies in the lock file can have uppercase letters, while all pdm.lock entries are lowercase. The package that I found this in was flask. 
This PR fixes this by making all package names lowercase it `getDepsRecursively`.


purepani:
I've realized that this patch is a terrible solution, and this should probably be fixed in upstream for https://github.com/adisbladis/pyproject.nix/, so I'll close this, and just open an issue.

PR #746, opened 2023-10-17
feat: allow pip drvs to access config from their parent drv

yajo:
I'm doing custom builds from editable installs in a meta-repo.

To get the dirty editable path of a dependency drv, I'd need this patch, to be able to do something like:

```nix
{config, ...}:
let
  inherit (config.mkDerivation.passthru) topConfig;
  fullRelPath = topConfig.lock.content.fetchPipMetadata.sources.${config.name}.path;
in ...
```

@moduon MT-1075

### DIFF
--- a/modules/dream2nix/pip/default.nix
+++ b/modules/dream2nix/pip/default.nix
@@ -106,6 +106,8 @@
           deps = defaultDeps ++ (l.concatLists (l.map depsByExtra cfg.buildExtras));
         in
           l.map (name: cfg.drvs.${name}.public.out) deps;
+
+        passthru.topConfig = config;
       };
     };
   };


DavHau:
Why don't you just access the `config` attribute of the parent directly from your nix expression?

yajo:
Yeah, I can do that, and that's currently my workaround.

However, my parent nix epression is using pip and the value needed to pass in `pip.drvs.*` is quite complex, so I refactored it out to different files that define reusable drv-parts.

I can call those parts, but it seemed more ergonomic to just declare a module and be able to access its parent config directly.

DavHau:
Hm, I wonder if there is a cleaner solution to this than passing the full parent config to the child module. Maybe we could support the editable use case directly or alternatively pass the path of the package via meta.relPath or such?

I'm happy to look into this. It would help if you could provide a reproducer of your use case so I can plan around with it.

PR #760, opened 2023-10-23
groups: change structure

DavHau:
renames:
- packages.foo."1.0.0" -> packages.foo."1.0.0".conf
- packagesEval.foo."1.0.0" -> packages.foo."1.0.0".evaluated
- public.packages.foo."1.0.0" -> packages.foo."1.0.0".public

### DIFF
--- a/modules/dream2nix/WIP-python-pdm/default.nix
+++ b/modules/dream2nix/WIP-python-pdm/default.nix
@@ -64,9 +64,9 @@ in {
   mkDerivation = {
     propagatedBuildInputs =
       map
-      (x: lib.head (lib.attrValues x))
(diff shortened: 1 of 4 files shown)


FRidh:
Any reason to use `conf` instead of say `config` which is also used in NixOS? Or is it because they're not exactly the same?

DavHau:
They are not exactly the same. `conf` is a module which itself can have `config` and `options`. We could name it `module` instead but I thought `conf` is better.

Or would it be better to rename `conf` to `module`?

FRidh:
Yes, module is more clear I think. Regardless, being explicit like this about the different attributes is very helpful.

DavHau:
done in https://github.com/nix-community/dream2nix/pull/761

PR #777, opened 2023-11-02
pdm: Use pyproject.nix for wheel selection & fetching

adisbladis:
There is a lot more from `lib.nix` that could be ported over to use functions from https://nix-community.github.io/pyproject.nix/lib/pypa.html.

This is a minimal effort to use the pyproject.nix implementations.


modules/dream2nix/WIP-python-pdm/default.nix
```diff
@@ -9,14 +9,20 @@
...
 
   libpyproject = import (dream2nix.inputs.pyproject-nix + "/lib") {inherit lib;};
-
-  selectWheel = import ../../flake-parts/lib/internal/python/selectWheel.nix {
+  libpyproject-fetchers = import (dream2nix.inputs.pyproject-nix + "/fetchers") {
     inherit lib;
-    pep599 = libpyproject.pep599;
-    python = config.deps.python3;
-    stdenv = config.deps.stdenv;
+    pkgs = config.deps;
```
DavHau, line 14: I think if we pass the whoel config.deps here, this will make the error in case of missing attributes more obscure. I created a patch for pyproject.nix here: https://github.com/nix-community/pyproject.nix/pull/29/files

DavHau (commented):
Thanks, that's great! I opened a PR to fix some issue: https://github.com/nix-community/pyproject.nix/pull/29/files

PR #793, opened 2023-11-18
pip: disable cache for modified response

VincentVanlaer:
The pypi proxy filters packages, which can end up corrupting the pip cache. For an example of what this would look like, see https://github.com/pypi/warehouse/issues/14457


DavHau:
This seems to impose a significant performance bottleneck. With large closures re-fetching everything all over again on each lock update is quite expensive considering that some python packages are several 100MB in size.

While I'm aware that fixing the bug is more important than performance, I wonder if there are alternatives. Could we just remove the versions from the response as well and fix the inconsitency thereby?

VincentVanlaer:
Did some digging in the pip source code, found a way to disable the cache only on the modified reponses.

PR #814, opened 2024-01-02
Add git to the npm lock fetch script path

raboof:
Testing with pisignage-server, I found that it failed to create `lock.json` when I ran `flake run .#pisignage-server.lock`, presumably because https://github.com/colloqi/pisignage-server/blob/59d9fb6b238c00a2ba6d166cd58c444f738416bf/package-lock.json#L3108C4-L3108C4 requires git to fetch.

With this change, it now creates a `lock.json` - however without an `integrity` section:

```
      "node_modules/919.socket.io": {
        "version": "0.9.19",
        "resolved": "git+ssh://git@github.com/colloqi/socket.io.git#7534db0e49b8310047527cc46e3c3342d3986be3",
        "dependencies": {
          "base64id": "0.1.0",
          "policyfile": "0.0.4",
          "socket.io-client": "0.9.16"
        },
        "engines": {
          "node": ">= 0.4.0"
        },
        "optionalDependencies": {
          "redis": "0.7.3"
        }
      },
```

... and even if it would have an `integrity` field, it fails to download:

```
trying git+ssh://git@github.com/colloqi/socket.io.git#7534db0e49b8310047527cc46e3c3342d3986be3
curl: (1) Protocol "git+ssh" not supported or disabled in libcurl
error: cannot download socket.io.git-7534db0e49b8310047527cc46e3c3342d3986be3 from any mirror
```

... so it seems further explicit support for fetching would be needed?

Interestingly the 'upstream' package-lock.json doesn't have an `integrity` field for this package either

### DIFF
--- a/modules/dream2nix/nodejs-package-json-v3/default.nix
+++ b/modules/dream2nix/nodejs-package-json-v3/default.nix
@@ -33,6 +33,7 @@ in {
           bash
           coreutils
           gawk
+          git
           path
           writeScript
           writeScriptBin
@@ -45,6 +46,7 @@ in {
       [
         config.deps.coreutils
         config.deps.npm
+        config.deps.git
       ]
       ''
         source=${cfg.source}
--- a/modules/dream2nix/nodejs-package-json/default.nix
+++ b/modules/dream2nix/nodejs-package-json/default.nix
@@ -38,6 +38,7 @@ in {
           bash
           coreutils
           gawk
+          git
           path
           writeScript
           writeScriptBin
@@ -48,6 +49,7 @@ in {
       writers.writePureShellScript
       [
         config.deps.coreutils
+        config.deps.git
         npm
       ]
       ''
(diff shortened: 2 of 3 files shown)


DavHau:
> Interestingly the 'upstream' package-lock.json doesn't have an integrity field for this package either

That's expected and OK. The url `git+ssh://git@github.com/colloqi/socket.io.git#7534db0e49b8310047527cc46e3c3342d3986be3` contains the git revision after the `#` symbol, which can be used in builtins.fetchGit to verify integrity.

Basically just use `builtins.fetchGit` instead of `pkgs.fetchgit` and just pass `url` and `rev`. It should be able to deal with the `ssh://` kind of URls.

And thanks a lot for the fix!

raboof:
> > Interestingly the 'upstream' package-lock.json doesn't have an integrity field for this package either
> 
> That's expected and OK. The url `git+ssh://git@github.com/colloqi/socket.io.git#7534db0e49b8310047527cc46e3c3342d3986be3` contains the git revision after the `#` symbol, which can be used in builtins.fetchGit to verify integrity.

Gotcha, thanks for the confirmation!

> Basically just use `builtins.fetchGit` instead of `pkgs.fetchgit` and just pass `url` and `rev`. It should be able to deal with the `ssh://` kind of URls.

Aah, that's the ticket - updated the PR!

Issue #829, opened 2024-01-12
[pip] Error: A definition for option `lock.extraScripts` is not of type `list of path`


yukkop:
Hi,

I am trying to use dream2nix to set up my Python development environment. I want to use requirements.txt, therefore I am using  [pip.requirementsFiles](https://nix-community.github.io/dream2nix/options/pip.html?highlight=requirementsList#opt-pip.requirementsFiles).

In the default [python-project](https://github.com/nix-community/dream2nix/tree/main/examples/packages/single-language/python-project) example, I only changed this part:

`default.nix`
```nix
  # ...
  pip = {
    pypiSnapshotDate = "2023-08-27";
    requirementsFiles = [ "./requirements.txt" ];
    flattenDependencies = true;
  };
  # ...
```

`requirements.txt`
```ignore
requests==2.25.1
python-telegram-bot==13.1
pyOpenSSL==20.0.1
python-i18n[YAML]==0.3.9
Flask==2.0.2
Pillow==8.4.0
argparse==1.4.0
colorlog==6.7.0
```

 I get the following error:
 ```ignore
       error: A definition for option `lock.extraScripts' is not of type `list of path'. Definition values:
       - In `/nix/store/b2c9v0629dkajlmizwaddv7ivn2wmjvm-source/modules/dream2nix/core/lock/interface.nix': ""
 ```

yukkop:
This is a minimal [repository](https://github.com/yukkop/dream2nix.extraScripts.error-example) with an example of the error

DavHau:
Fixed by #831

PR #831, opened 2024-01-14
Fix type error in default value provided for option lock.extraScripts

grybiena:
**Problem**: The default is an empty string but this violates the type constraint for the value which expects a list.

**Fix**: change empty string to empty array.

This bug is mentioned by @yukkop in this issue https://github.com/nix-community/dream2nix/issues/829


DavHau:
Thanks

PR #912, opened 2024-03-27
Adds editable package to pdm devshell 

purepani:
This pr uses a modified version of https://github.com/nix-community/poetry2nix/blob/master/editable.nix to add the local package to the devShell.

This is very unsophisticated so feel free to improve it(Ex. it only looks in ./src for packages).

Note: This uses the $PWD env variable as a reference to the current directory in order to make this work with flakes. This isn't ideal, but since flakes are designed to be fully hermetic, this is impossible to get around as far as I can tell.

### DIFF
--- a/modules/dream2nix/WIP-python-pdm/default.nix
+++ b/modules/dream2nix/WIP-python-pdm/default.nix
@@ -119,6 +119,33 @@ in {
       config.public.pyEnv
       config.deps.pdm
     ];
+    shellHook = let
+      name = libpyproject.pypa.normalizePackageName pyproject.pyproject.project.name;
+      pkgInfoFields = {
+        Metadata-Version = "2.1";
+        Name = name;
+        Version = pyproject.pyproject.project.version;
+        Summary = pyproject.pyproject.project.description;
+      };
+      pkgInfo = lib.concatStringsSep "\n" (lib.mapAttrsToList (key: value: "${key}: ${value}") pkgInfoFields);
+    in ''
+      DREAM_SITE=$PWD/.dream2nix/pdm/site-packages
+         	#Prepare .dream2nix/pdm/site-packages
+      mkdir -p $DREAM_SITE
+      export PYTHONPATH=$DREAM_SITE:$PYTHONPATH
+      touch $DREAM_SITE/sitecustomize.py
+      echo "import site; site.addsitedir('$DREAM_SITE')" > $DREAM_SITE/sitecustomize.py
+      touch $DREAM_SITE/pdm-editable.pth
+
+      #Add PTH locations
+      echo "$PWD/src" > $DREAM_SITE/pdm-editable.pth
+      echo "$PWD" >> $DREAM_SITE/pdm-editable.pth
+
+      #Add editable packages
+      mkdir -p $DREAM_SITE/${name}.egg-info
+      touch $DREAM_SITE/${name}.egg-info/PKG-INFO
+      echo "${pkgInfo}" > $DREAM_SITE/${name}.egg-info/PKG-INFO
+    '';
   };
 
   groups = let


purepani:
Also, the first commit just changes the order pdm and the interpreter package in buildInputs for the shell; this was done because otherwise the python interpreter for pdm would come before the interpreter for the environment.

modules/dream2nix/WIP-python-pdm/default.nix
```diff
@@ -108,12 +108,23 @@ in {
...
       ps:
         config.mkDerivation.propagatedBuildInputs
+        ++ [
+          (import ./editable.nix {
+            inherit (config.deps) runCommand python;
+            inherit (pyproject) pyproject;
+            inherit libpyproject;
+            inherit lib;
+            editablePackageSources = {
+              "${pyproject.pyproject.project.name}" = /${builtins.getEnv "PWD"}/src;
```
phaer, line 118: * We probably shouldn't hardcode  `/src` here, as not all project use a "src layout", see e.g. https://backend.pdm-project.org/build_config/.

* Using `builtins.getEnv "PWD"` should work fine in impure mode, but maybe we could show a nicer error such as "Error: editable installs are only supported in impure mode." or so?

phaer (commented):
Thank you @purepani, I think something like this would be a very nice start to improve the python devshell experience in dream2nix! :tada: 

* This doesn't handle entrypoints yet, right? I think at least console_script support would be nice to have and to be expected from an editable install in python?

* Coincidentally, I just tried to build something similar - outside of dream2nix for the time being, but somewhat compatible - on PEP 660 and by calling the packages build-backend. It's still missing a few features to be useful, see inline comments: https://github.com/phaer/entangled_snakes/blob/66bef9aad901d9ed7382b07ef88b356983dd6e3d/lib/editable.nix#L33
But maybe its good enough for inspiration

purepani:
> Thank you @purepani, I think something like this would be a very nice start to improve the python devshell experience in dream2nix! 🎉
> 
>     * This doesn't handle entrypoints yet, right? I think at least console_script support would be nice to have and to be expected from an editable install in python?
> 
>     * Coincidentally, I just tried to build something similar - outside of dream2nix for the time being, but somewhat compatible - on PEP 660 and by calling the packages build-backend. It's still missing a few features to be useful, see inline comments: https://github.com/phaer/entangled_snakes/blob/66bef9aad901d9ed7382b07ef88b356983dd6e3d/lib/editable.nix#L33
>       But maybe its good enough for inspiration

Thanks,
I'll put this in draft mode while I continue to work on it. I'll take a look at your implementation and also try to make it compliant with PEP660

DavHau:
Hey, thanks for taking a look at this.

I cherry-picked your first commit now via https://github.com/nix-community/dream2nix/pull/917

I think requiring `--impure` for the dev-shell should be avoided add all costs as it negatively impacts UX.
The good new is, it looks like that can be fixed by moving this logic into the shellHook instead of having it as a separated derivation.

The challenge with the shellHook is to make the python interpreter accept the `.pth` file. The problem is that by default, it doesn't scan things like `PYTHONPATH` for `.pth` files. But there is a workaround [described here](https://stackoverflow.com/questions/15208615/using-pth-files).

After experimenting with that work around, I come o the conclusion, that something like this should work within the shellHook:
- create a hidden directory at the root of the project (something like `.dream2nix/pdm/site-packages`) and add it to PYTHONPATH
- put a [sitecustomize.py](https://docs.python.org/3/library/site.html#module-sitecustomize) inside that directory which executes `import site; site.addsitedir(".dream2nix/pdm/site-packages'')`
- since the temp dir is now a site dir, any `.pth` files inside it will be loaded once the interpreter starts
- now generate the editable packages into `.dream2nix/pdm/site-packages`

Let's do this step by step.
If we can get something working that doesn't require impure eval and achieves that the currently developed package can be imported, I'd be happy to merge it, as this fixes the most pressing issue.
Supporting additional stuff like scripts is also important, but can be done in a second PR.

purepani:
Thanks for pointing this out. I didn't think it was possible at first until I realized shellHook just has direct access to $PWD, which is the only mechanism that makes this possible. I'll get to work on this

purepani:
The only thing I'm worried about in this case is the polluting of PYTHONPATH: if another program tries to run that uses the same package that you're developing, wouldn't that potentially cause issues?

DavHau:
> The only thing I'm worried about in this case is the polluting of PYTHONPATH: if another program tries to run that uses the same package that you're developing, wouldn't that potentially cause issues?

Potentially yes, but I wouldn't put too much effort into this edge case yet. We can still improve the slution later.

phaer:
> put a [sitecustomize.py](https://docs.python.org/3/library/site.html#module-sitecustomize) inside that directory which executes import site; site.addsitedir(".dream2nix/pdm/site-packages'')

I was about to write that this would be somewhat messy because every nixpkgs python env ships its own sitecustomize.py which we would shadow in a naive approach, so we would need to patch it for something like this to work.

But turns out, this changed since I tried something similar about a month ago. sitecustomize.py in nixpkgs has been removed in https://github.com/NixOS/nixpkgs/commit/234bb31f611f43f8b744b305ab82035de637aaca :tada:

purepani:
@phaer @DavHau This implements the suggestion from @DavHau earlier. The only package it adds at the moment is the one which is defined by the name in pyproject.nix.

purepani:
Unfortunately the fix in nixpkgs that would get rid of the sitecustomize.py was reverted due to breaking of a bunch of packages.

phaer:
@purepani Thanks for the notice - I would otherwise have missed https://github.com/NixOS/nixpkgs/pull/302385. Unfortunate that python envs from nixpkgs will be non-conformant again because there the workaround is too entrenched. Will have a look at how to fix this upstream. Maybe we could just remove/vendor the included sitecustomize? :thinking: 

Anyways, I had a go at a devshell implementation in https://github.com/nix-community/dream2nix/compare/main...phaer:dream2nix:devshells?expand=1, but just notice that I hadn't posted it here yet (discussed it in the weekly meeting with david).

Two open things I want to add is integration with the pdm submodule (uses the pip one atm) and wrapping of the binaries. Shouldn't be hard, I just lacked time last week after a family incident.

purepani:
Oh nice! Thanks for the quick notice: I was planning on working on making it PEP660 compliant after I woke up, but it looks like you have finished nearly everything! This'll be a great improvement to the python modules once it's ready!

purepani:
Closed due to being superceded by https://github.com/nix-community/dream2nix/pull/951

Issue #930, opened 2024-04-14
realpath: command not found


ahihi:
hi, dream2nix beginner here. i am trying to build a python application with pip dependencies, on macOS 10.14.

<details>
<summary>flake.nix</summary>

```nix
{
  inputs = {
    dream2nix.url = "github:nix-community/dream2nix";
    nixpkgs.follows = "dream2nix/nixpkgs";
    flake-utils.url = "github:numtide/flake-utils";
  };
  
  outputs = { self, dream2nix, nixpkgs, flake-utils, ... }:
    {
      overlays.default = final: prev: {
        musiclib5 = final.callPackage ./default.nix {};
      };
    } //
    (flake-utils.lib.eachDefaultSystem (system:
      let        
        pkgs = import nixpkgs {
          inherit system;
          overlays = [self.overlays.default];
        };
      in {
        packages.default = dream2nix.lib.evalModules {
          packageSets.nixpkgs = dream2nix.inputs.nixpkgs.legacyPackages.${system};
          modules = [
            ./default.nix
            {
              paths.projectRoot = ./.;
              # can be changed to ".git" or "flake.nix" to get rid of .project-root
              paths.projectRootFile = "flake.nix";
              paths.package = ./.;
            }
          ];
        };
      }
    ));
}
```
</details>

<details>
<summary>default.nix</summary>

```nix
{ config, dream2nix, lib, ... }:
{
  imports = [
    dream2nix.modules.dream2nix.pip
  ];
  
  name = "musiclib5";
  version = "0.1";

  deps = { nixpkgs, ... }: {
    python = nixpkgs.python312;
  };

  buildPythonPackage = {
    pythonImportsCheck = [
      config.name
    ];
  };

  pip = {
    pypiSnapshotDate = "2024-04-14";
    requirementsList = [
      "${config.name}==${config.version}"
    ];
  };
}
```
</details>

if i understand correctly, i need to generate a lockfile with `nix run .#default.lock`. but this fails with:

```
/nix/store/133dx9z930qzx06gd38vwx4s0grba2gz-refresh/bin/refresh: line 6: realpath: command not found
```

it seems `realpath` [was added to macOS 13](https://apple.stackexchange.com/questions/450035/is-the-unix-realpath-command-distributed-with-macos-ventura), but is not in older versions. i imagine it is available in `coreutils` — is it possible to make the lockfile creation depend on that?

DavHau:
Yes for sure. Making it depend on coreutils sounds like the right fix.

PR #931, opened 2024-04-15
core.lock: Add coreutils dependency

phaer:
...and use realpath from there.

Should close https://github.com/nix-community/dream2nix/issues/930


DavHau:
Thanks

PR #939, opened 2024-04-25
build(direnv): explicitly allow IFD for devshell

yajo:
Otherwise, if your Nix is configured to block IFDs, direnv will fail always.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,1 @@
-use flake ./dev-flake
+use flake ./dev-flake --allow-import-from-derivation


phaer:
The devshell shouldn't require IFD and works for me on current main without allow-import-from-derivation. Maybe that problem is solved since: https://github.com/nix-community/dream2nix/commit/ef046832220655543b105949787a9d0e6a24211e ?

Closing this for those reasons. Feel free to re-open if you think I am wrong!

PR #948, opened 2024-04-30
feat(buildPythonPackage): support new pyproject format

yajo:
https://github.com/NixOS/nixpkgs/pull/271597 implemented a new format for python packages that is starting to get traction in nixpkgs/unstable (soon to become NixOS 24.05).

That is a progress towards https://github.com/NixOS/nixpkgs/issues/272178, which would be a massive improvement for dream2nix once completed.

This first step just makes sure that `buildPythonPackage` supports the new arguments properly. Without this change, many dream2nix python derivations fail to build on nixos-unstable right now.

Probably other python auto-builders should leverage these new options when available for the d2n user. That can be added later.

@moduon MT-1075

### DIFF
--- a/modules/dream2nix/buildPythonPackage/interface.nix
+++ b/modules/dream2nix/buildPythonPackage/interface.nix
@@ -87,11 +87,57 @@ in {
         '';
       };
 
+    pyproject = l.mkOption {
+      type = t.nullOr t.bool;
+      default = null;
+      description = ''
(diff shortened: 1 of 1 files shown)


yajo:
So, ready to merge?

phaer:
> So, ready to merge?
Thought so during my review, but lets see what @DavHau has to say about it :)

PR #951, opened 2024-05-01
Python DevShells (for pip)

phaer:
Hello,

This is a PR which implements devshells with support for console scripts and richer metadata then the approach in https://github.com/nix-community/dream2nix/pull/912. Its currently only implemented for the pip builder, but the idea should be portable to PDM as well. Planning to look into this soon.

*  users need to declare an the root package and/or dependencies of a given python project as editable. Consider a devshell for a package named `my-tool`:
```nix
{
  pip.editables.charset-normalizer = "/home/my-user/src/charset-normalizer";
  pip.editables.click = null;
  pip.editables.my-tool = null;
}
```
The first expression would link the editable `charset-normalizer` from the given absolute path, to `.dream2nix/editables/charset-normalizer` and install it.
The second line would default to copying `click` from the nix store to `.dream2nix/editables/click` and install it.
As the first one matches the root package, we would default to link our project root to `.dream2nix/editables/my-tool` and install it.

* We also implicitly adds dependencies which are "direct installs" in pips report to editables, e.g. "click @ git+https://github.com/pallets/click.git@main" for convenience.

* "Installing" here only affects the devshell, we don't touch the results of `nix build`, but instead run a python script (`editable.py`) in a shellHook to get the right PATH and PYTHONPATH for the shell setup and install wrapper scripts for console_scripts of our dependencies.

* We do change public.pyEnv by moving its upstream sitecustomize.py to _sitecustomize.py. It was recently re-introduced into nixpkgs as its removal caused a few pkgs to break, but it did work for the overwhelming majority without issues. Moving it, deactivates it by default but it can easily be re-included if needed. We do so in the devshell just in case in https://github.com/nix-community/dream2nix/compare/main...phaer:dream2nix:devshells?expand=1#diff-4f29741e77180e7c500ebe90ecaa8af18d8aede267e142bf2f83888073079c9eR213

* This PR includes commits from https://github.com/nix-community/dream2nix/pull/932. I can drop them if unwanted.
* The history includes my first attempt as a shell script. Should I squash this?

# Give it a try

Just add some `pip.editables` as above to your package and include 
`${your-pkg.config.pip.editablesShellHook} ` in your shellHook or see
[examples/packages/languages/python-local-development/default.nix](https://github.com/nix-community/dream2nix/compare/main...phaer:dream2nix:devshells?expand=1#diff-6f11abbf9307b968b92e57cb6cc4e6dbb316b730dc95440846a6e1ccc604bb5a) and the devShell there.

### DIFF
--- a/modules/dream2nix/pip/editable.nix
+++ b/modules/dream2nix/pip/editable.nix
@@ -6,17 +6,11 @@
   pyEnv,
   editables,
   rootName,
-  drvs,
+  sources,
 }: let
   args = writeText "args" (builtins.toJSON {
-    inherit findRoot unzip rootName pyEnv editables;
+    inherit findRoot unzip rootName pyEnv editables sources;
     inherit (pyEnv) sitePackages;
-    drvs =
-      lib.mapAttrs (n: v: {
-        inherit (v.public) out;
-        inherit (v.mkDerivation) src;
-      })
-      drvs;
   });
 in ''
   source <(${pyEnv}/bin/python ${./editable.py} ${args})
--- a/modules/dream2nix/pip/editable.nix
+++ b/modules/dream2nix/pip/editable.nix
@@ -1,6 +1,7 @@
 {
   lib,
   findRoot,
+  unzip,
   pyEnv,
   editables,
   drvs,
@@ -17,14 +18,13 @@
   mkEditable = name: path: let
     drv = drvs.${name};
   in ''
+    name="${lib.replaceStrings ["-"] ["_"] name}"
     source="${
       if path != null
       then path
       else drv.mkDerivation.src
     }"
-    editable_dir="$editables_dir/${name}"
-    mkdir -p "$editable_dir"
-
+    editable_dir="$editables_dir/$name"
     # Realize the non-editable variant of this package,
     # if it's not in /nix/store already. We need its
     # .dist-info directory and might need it's unpackaged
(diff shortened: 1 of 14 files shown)


phaer:
Regarding pdm support: @DavHau mentioned that he might want to re-do the groups feature of the pdm module. For the editables we need a) a python environment and b) an attrset with one drv per package.

Another area which might need a bit more work, is garbage collection in .dream2nix. We probably don't want to drop .dream2nix/editables for each shell as it might contain actual worktrees. So we need to diff the declared editables and the existing ones.

modules/dream2nix/pip/editable.nix
```diff
@@ -6,17 +6,11 @@
...
   args = writeText "args" (builtins.toJSON {
     inherit findRoot unzip rootName pyEnv editables;
     inherit (pyEnv) sitePackages;
-    drvs =
-      lib.mapAttrs (n: v: {
-        inherit (v.public) out;
-        inherit (v.mkDerivation) src;
-      })
-      drvs;
+    inherit (builtins) storeDir;
```
phaer, line 13: left-over? Doesn't seem to end up used?

modules/dream2nix/pip/editable.py
```diff
@@ -29,69 +43,30 @@ def make_editable(
...
+    root_dir,
+    store_dir,
 ):
     normalized_name = name.replace("-", "_")
     editable_dir = editables_dir / name
-    drv_out = Path(drvs[name]["out"])
+    full_path = path if path.is_absolute() else root_dir / path
     if editable_dir.exists():
-        print(f"Skipping existing editable source in {editable_dir}", file=sys.stderr)
+        relative_editable_dir = os.path.relpath(editable_dir, root_dir)
```
phaer, line 55:
```suggestion
        relative_editable_dir = editable_dir.relative_to(root_dir)
```

lets stick to pathlibs api, as we do use that in the lines above and below

modules/dream2nix/pip/editable.py
```diff
@@ -29,69 +43,30 @@ def make_editable(
...
-        shutil.copytree(source, editable_dir, symlinks=True)
-        run(["chmod", "-R", "u+w", editable_dir])
-    else:
-        print(
-            f"Linking editable source from {source} to {editable_dir}",
-            file=sys.stderr,
-        )
-        run(["ln", "-sf", source, editable_dir])
+        exit(1)
+    # link_editable_source(editable_dir, site_dir, normalized_name, full_path)
```
phaer, line 64: maybe this can be removed? Looks like a left-over as well?

modules/dream2nix/pip/editable.py
```diff
@@ -166,6 +166,36 @@ def make_entrypoints(python_environment, bin_dir, dist_info):
         os.chmod(bin_dir / name, 0o755)
 
 
+def is_state_valid(args, dream2nix_python_dir):
```
phaer, line 169:
```suggestion
def has_state_changed(args, dream2nix_python_dir):
```

maybe? It doesn't really seem to be about validity :)

PR #969, opened 2024-05-27
Pdm editables

purepani:
A very literal copy paste of the pip editables from https://github.com/nix-community/dream2nix/pull/951
@DavHau Should I factor this out into a separate `editables` module?


modules/dream2nix/WIP-python-pdm/default.nix
```diff
@@ -106,19 +123,41 @@ in {
...
 
   public.devShell = config.deps.mkShell {
+    shellHook = config.public.shellHook;
     packages = [
       config.public.pyEnv
       config.deps.pdm
     ];
+    buildInputs =
+      [
+        config.deps.python.pkgs.tomli
```
phaer, line 149: I think we might have to check whether our package already depends on (another version of) tomli? If so, we can just use that one and leave deps.pythonpkgs.tomli out in order to avoid conflicts.

phaer (commented):
Great work @purepani! 🎉 

Yes, I think it would be best to refactor a single editables module for both pip and pdm to share as much of the implementation as possible between them.

modules/dream2nix/WIP-python-pdm/default.nix
purepani: Oh I think I changed that bc pdm has overrides in a completely different module. I'll add that back in!

purepani: With regards to this, should we make the pip module use the overrides module instead of pip.overrides? Then we can just put all of that logic in the new editables module, and it also makes more sense.

purepani: Oh I think David wants the opposite: 
`overrides` and `overridesAll` to `pdm.overrides` and `pdm.overridesAll`

purepani:
This feels a bit messy but it's factored into a separate module, and functional for both pdm and pip (I have some notes to put in, which I will do later). Let me know if anything is broken(I was doing some git stuff that I am pretty sure didn't break anything, but it may have). 
I also just realized I put the commits in the wrong order, so only the last one is functional, since the other's assume there's the editables module lol.

purepani:
Notes of changes made:
Override options for the pdm module were added into pdm.overrides and pdm.overridesAll, but at the moment, they just inherit from top level overrides directly. 
I'm only check the default group for any overrides and build inputs in pdm at the moment, and the dev shells are currently made in the module individually, since they are located in different places, and it just seemed simpler to do it like that. It might be possible to put the devshell in the new module if the apis for the python module are made to match

purepani:
I will put it into draft mode temporarily because i'm planning on reorganizing the commits since the current ones suck; I'm not planning on changing the code though, so this is still reviewable in it's current state

purepani:
Commits should all work independently now.

modules/dream2nix/WIP-python-pdm/default.nix
purepani: I'll tentatively resolve this since I added the overrides back in.

modules/dream2nix/WIP-python-pdm/default.nix
```diff
@@ -24,6 +24,8 @@
   lock_data = lib.importTOML config.pdm.lockfile;
   environ = libpyproject.pep508.mkEnviron config.deps.python;
 
+  #editables = lib.filterAttrs (_name: path: path != false) config.pdm.editables;
```
phaer, line 27: Is this a left-over?

modules/dream2nix/pip/default.nix
```diff
@@ -15,7 +15,7 @@
   isBuildInput = drv: cfg.buildDependencies.${drv.name} or false;
 
   # actually wanted editables (minus the ones set to false)
-  editables = lib.filterAttrs (_name: path: path != false) config.pip.editables;
+  #editables = lib.filterAttrs (_name: path: path != false) config.pip.editables;
```
phaer, line 18: Same as above

modules/dream2nix/python-editables/interface.nix
```diff
@@ -0,0 +1,40 @@
...
+
+    name = lib.mkOption {
+      type = t.str;
+      #default = {};
+      internal = true;
+    };
+
+    paths = lib.mkOption {
+      type = t.attrsOf t.str;
+      #default = {};
```
phaer, line 31: I think this default should not be commented?

modules/dream2nix/python-editables/interface.nix
```diff
@@ -0,0 +1,40 @@
...
+    };
+
+    editablesDevShell = lib.mkOption {
+      type = t.package;
+      readOnly = true;
+    };
+
+    name = lib.mkOption {
+      type = t.str;
+      #default = {};
```
phaer, line 25: I think this comment could be removed?

phaer (approved):
A few lines of commented-out code seem to be oversights? Otherwise lgtm! Thanks again!

purepani:
Yep those were leftover. Just fixed them as needed!

DavHau:
Good job! Thank you

PR #985, opened 2024-06-13
pip: use pkgs.fetchurl instead of builtins.fetchurl

yorickvP:
This resolves
error: store path 'nypj2b7h41zsrmyfsiww97zb6va445gn-torch-2.0.1%2Bcpu-cp310-cp310-linux_x86_64.whl' contains illegal character '%'

It also decreases evaluation times by moving the fetch into the build process.


phaer:
LGTM, thank you!

PR #986, opened 2024-06-13
filter-pypi-responses: stream package requests that hit the proxy

yorickvP:
mitmproxy was buffering all requests (except pythonhosted.*), causing pip to time out when getting big packages from custom URLs.

Setting flow.response.stream in the responseheaders hook fixes this.


phaer:
Good catch, thanks @yorickvP! 

In many cases where it isn't important to make the locking itself reproducible, as repeatable builds from the lockfiles are sufficient it's also possible to just disable to disable the proxy alltogether with pypiSnapshotDate = null.

PR #987, opened 2024-06-13
fix: first chapter was skipped in a nested set of chapters

yorickvP:


phaer:
Thank you. I sadly missed this before and we meanwhile finally switched the documentation from mdbook to mkdocs in current main. Module rendering could still be improved, but my next priority here are guides.

Let me know if e.g. https://nix-community.github.io/dream2nix/reference/pip/ is complete here :)

yorickvP:
Funfact, it's not :)

It's missing e.g. `paths.cacheFile` and `pip.overrideAll.buildPythonPackage.build-system`

Something's wrong with `preprocess_options` that causes it to generate
```
       "uv": {
         "children": {"extraArgs":…},
         "enable": {"declarations":…}
       }
```

instead of

```
      "uv": {
        "children": {
          "extraArgs": ...,
          "enable": {"declarations": ...}
        }
      },
```

phaer:
@yorickvP You are right, thanks a lot for the report & example. I believe I found my, bit ridiculous, mistake and fixed it in https://github.com/nix-community/dream2nix/pull/1009/files

PR #995, opened 2024-06-23
Fix python ediables

purepani:
`editables.py` doesn't export the correct environment variables when `.dream2nix` doesn't need updating, so editables don't get imported until the script thinks that there needs to be an update. 

The first commit isn't needed, but it made it a bit easier to test since without it, `mytool` is always importable unless you change directories first. @phaer


phaer (approved):
Thanks for testing, finding and fixing! :)

PR #1009, opened 2024-07-04
docs: fix incorrect option rendering for first children of nested options

phaer:
thanks @yorickvP for reporting this in
https://github.com/nix-community/dream2nix/pull/987


phaer:
nix-community buildbot seems to have issues atm. I restarted the job, but it's still hanging during flake evaluation.

yorickvP:
Thanks for the fix!

PR #1013, opened 2024-07-11
pip guides, docs devshell fix

phaer:
- [X] pillow example
- [X] python-local-development example
- [x] devshell with editables
- [x] editables options docs
- [ ] pdm example & comparision
- [x] more elaborate flattenDependencies option docs


phaer:
@purepani For some reason, I can't select you as reviewer here, but this might be of interest to you :)

docs/src/guides/pip.md
```diff
@@ -0,0 +1,257 @@
...
+      "aarch64-linux"
+      "x86_64-darwin"
+      "x86_64-linux"
+    ];
+  in {
+  packages = eachSystem (system: {
+    default = dream2nix.lib.evalModules { # (1)
+      packageSets.nixpkgs = nixpkgs.legacyPackages.${system};
+      modules = [
+        ./default.nix (2)
```
purepani, line 44: Missing comment before `(2)`

docs/src/guides/pip.md
```diff
@@ -0,0 +1,257 @@
...
+  });
+}
+```
+
+1. We call our package attribute `default` here...
+2. ...and the nix file `default.nix` here.
+
+## Example: Pillow
+
+Things get a bit more interesting in `default.nix` where we define a package module which fetches Pillow from pypi and builds it with minial features - just JPEG support. Click the plus to expand any code annotation below for details.
```
purepani, line 61: Minimal*

docs/src/guides/pip.md
```diff
@@ -0,0 +1,257 @@
...
+2. Declare external dependencies, like the python interpreter to use and libraries from nixpkgs. We use whatever the latest `python3` in nixpkgs is as our python.
+3. Declare which build tools we need to pull from nixpkgs for use in `mkDerivation` below.
+4. Declare name and version of our package. Those will also be used for `pip.requirementsList` below.
+5. Set dependencies, `pkg-config` is only required
+during build-time, while the libraries should be propagated. We use `config.deps` instead of a conventional `pkg` here to be able to "override" inputs via the [module system](../modules.md).
+6. Tell the [buildPythonPackage module](../reference/buildPythonPackage/index.md) to verify that it can import the given python module from our package after a build.
+7. Tell the [pip module](../reference/pip/index.md) which dependencies to lock using the same syntax as 
+a `requirements.txt` file. Here: `pillow==10.4.0`.
+8. `pip` uses binary wheel files if available by default, we well it not to do so in order to ensure
+a build from source.
```
purepani, line 126: > we well it not to do so in order to ensure
Probably needs to be something like
> 8. `pip` uses binary wheel files if available by default. We will not do so in order to ensure
a build from source.

docs/src/guides/pip.md
```diff
@@ -0,0 +1,257 @@
...
+    pipFlags = [ # (8)
+      "--no-binary"
+      ":all:"
+    ];
+  };
+}
+```
+
+1. Import the dream2nix [pip module](../reference/pip/index.md) into our module.
+2. Declare external dependencies, like the python interpreter to use and libraries from nixpkgs. We use whatever the latest `python3` in nixpkgs is as our python.
```
purepani, line 117: The dream module is set specifically to python311, so this isn't true.

docs/src/guides/pip.md
```diff
@@ -0,0 +1,257 @@
...
+  mkDerivation = {
+    src = lib.cleanSourceWith { # (5)
+      src = lib.cleanSource ./.;
+      filter = name: type:
+        !(builtins.any (x: x) [
+          (lib.hasSuffix ".nix" name)
+          (lib.hasPrefix "." (builtins.baseNameOf name))
+          (lib.hasSuffix "flake.lock" name)
+        ]);
+    };
```
purepani, line 202: I'm not sure if we should include the filtering of the source in this example, since it's not really needed in a "getting started" guide, and most people would probably take this example and copy paste it.

purepani (changes requested):
I'm generally of the opinion that "getting started" guides should have a super simple example that people generally copy and build off of, so maybe we could simplify the "mytool" example a bit, though I'd maybe just push this as it is for now since that is fixable later. 

Also I'm not exactly sure what the name of the second example is. I remember changing it from "my-tool" to "mytool" or something since the hyphen caused issues, so we should make sure that is consistent.

docs/src/guides/pip.md
phaer: Very good catch, thanks! Copied this a bit too quickly when changing examples.

docs/src/guides/pip.md
phaer: Yes, it's a good question. The exaple here is, as you probably noticed, a verbatim copy of the `python-local-development` one in ./examples. If we remove it here, I'd prefer to remove it over there as well for consistency.

On one hand i can see that it opens too many topics at once for people, on the other hand getting an idea of whats copied to the nix store when is important to have any change of understanding why a rebuild happens - even more so with devshells.

I see if i manage to do a better explanation on that, or otherwise remove it for now.

docs/theme/partials/toc-item.html
```diff
@@ -1,16 +1,56 @@
+<!--
```
DavHau, line 1: This reverts the recent changes and re-introduces the issue that the TOC lacks detail about the option path. Let's keep experimentation with the TOC separate form this PR.

DavHau (commented):
Looks good, except the one thing I commented on.

docs/theme/partials/toc-item.html
phaer: Ah right - I had added that again before we discussed this as I found it useful for the pip docs (and still do ;)). But you are right that it shouldn't be in this PR. Dropped it.

docs/src/guides/pip.md
```diff
@@ -0,0 +1,336 @@
+---
+title: Build a python project with pip
+---
+
+!!! info
+
+    We recommend reading our [Getting Started](./getting-started.md) guide first if you have not done so yet!
+
+this guide we are going to take a look at two annotated examples using the [pip module](../reference/pip/index.md):
```
purepani, line 9: Hate to point this out right after you pushed, but I think this missing an "In" at the beginning of the sentence!

phaer: Haha, right. No worries, I'll fix it with the next PR, probably later "today" (I am in CEST).

PR #1016, opened 2024-07-11
docs: improve table of contents for options

DavHau:
- increase width of toc
- show full option path in toc
- no nesting

This simplifies navigation on option pages a lot.


phaer:
Can we reconsider this? The removal of the nested rendering makes the toc pretty useless for an overview. If one knows what the option is called, the search works well enough IMO.

PR #1020, opened 2024-07-18
fetchPipMetadata: set PKG_CONFIG_PATH...

phaer:
to nativeBuildInputs. We still need to run a setup.py for many packages at locking time. Those often try to find inputs via PKG_CONFIG_PATH so i think it's a reasonable default.

### DIFF
--- a/pkgs/fetchPipMetadata/script.nix
+++ b/pkgs/fetchPipMetadata/script.nix
@@ -55,6 +55,12 @@
       ;
   });
 
+  env' =
+    {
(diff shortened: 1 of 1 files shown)


DavHau (commented):
Grrat idea, but doesn't pkg-config requirr the inputs for the target atchitecture, eg. buildInputs instead of nativeBuildInputs?

phaer:
Yes, nevermind - I was misinterpreting the output of a failed mesonpy build; but the locking environment per se was correct after all. Sorry for the noise!

PR #1022, opened 2024-07-18
fixup: pip default version in pip module

phaer:


phaer:
Took me a while that i needed a flake lock --override-input dream2nix <this-branch> ofc, as the invalidationHash does include pipVersion.

PR #1024, opened 2024-07-18
reference-docs: fix option rendering issue

DavHau:
remove tree-style preprocessing since it is not needed anymore


phaer:
@DavHau As written in https://github.com/nix-community/dream2nix/pull/1016#issuecomment-2223499760 i'd like to at least discuss this - as I think the navigation is really hard to navigate without some form of nesting.

Ideally active elements (active through scrolling, clicking or deep-linking) would un-collapse automatically. I had started doing that, but then didn't want to spent on the customization. Yet i still think: a flat list of everything is of limited use.

PR #1036, opened 2024-08-15
fix: python torch override for mac - no cuda on mac

JonBoyleCoding:
Adding torch to a project on mac failed as there are no cudaPackages.

The fix should only enable cuda if Linux is detected.

### DIFF
--- a/overrides/python/torch/default.nix
+++ b/overrides/python/torch/default.nix
@@ -9,12 +9,14 @@
   # use the autoAddOpenGLRunpathHook to add /run/opengl-driver/lib to the RPATH
   #   of all ELF files
   deps = {nixpkgs, ...}: {
-    inherit (nixpkgs.cudaPackages) autoAddOpenGLRunpathHook;
+    inherit (nixpkgs.stdenv) isLinux;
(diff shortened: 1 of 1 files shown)


phaer (approved):
Thanks!

overrides/python/torch/default.nix
```diff
@@ -9,12 +9,14 @@
   # use the autoAddOpenGLRunpathHook to add /run/opengl-driver/lib to the RPATH
   #   of all ELF files
   deps = {nixpkgs, ...}: {
-    inherit (nixpkgs.cudaPackages) autoAddOpenGLRunpathHook;
+    inherit (nixpkgs.stdenv) isLinux;
+    autoAddOpenGLRunpathHook = lib.optionalAttribute nixpkgs.stdenv.isLinux nixpkgs.cudaPackages.autoAddOpenGLRunpathHook;
```
phaer, line 13: I don't think `lib.optionalAttribute` is a thing?

JonBoyleCoding: I don't know how I messed that up! I will look into and fix.

PR #1044, opened 2024-09-18
Enable building python reuse from source

chaoflow:


chaoflow:
@DavHau This works for me locally and is needed for python-3.12 for which no wheel is available on pypi.

PR #1067, opened 2024-10-26
Adds uv resolver for pdm

purepani:
PDM now experimentally supports a [uv resolver](https://github.com/pdm-project/pdm/pull/3144). This is significantly faster(even on the examples with very few dependencies, resolve times go from a few seconds to a few hundred milliseconds). This PR enables the use of that resolver. 

note that nixpkgs needs to be updated due to only the newer versions of pdm supporting it.

@DavHau @phaer

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,11 +18,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1728538411,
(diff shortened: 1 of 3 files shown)


phaer:
Nice speedup almost for free, what's not to like? :)

purepani:
@rupurt see https://github.com/nix-community/dream2nix/pull/1084

rupurt:
Amazing. Thank you @purepani :raised_hands:

PR #1084, opened 2024-12-25
pdm: Makes uv resolver optional

purepani:
Makes use of uv resolver optional per https://github.com/nix-community/dream2nix/pull/1067#issuecomment-2552461148
@phaer


phaer:
Formatting seems off, and it's a nit but we don't use snake case elsewhere for options I believe, so maybe call it uvResolver or even useUvResolver?

Otherwise LGTM, thank you!

purepani:
Should be good now!

PR #1085, opened 2024-12-25
Update pyproject.nix

purepani:
Moves pyproject.nix to more recent version(since the current one is quite old). Also switched to using `fetchPypiLegacy` from nixpkgs since pyproject.nix removed it. 
This shouldn't be merged right now since the current version of `pdm` in nixpkgs can't lock due to https://github.com/pdm-project/pdm/issues/3264

Once https://github.com/NixOS/nixpkgs/pull/353056 is merged we should be ok to merge here.


purepani:
@DavHau @phaer https://github.com/NixOS/nixpkgs/pull/368557 is merged and nixpkgs-unstable is updated so this can be reviewed now!

phaer:
Ah, we haven't updated nixpkgs here in dream2nix for a while and CI seems currently broken: https://github.com/nix-community/dream2nix/pull/1065. So I guess we have to do that first?

PR #1091, opened 2025-03-04
package-func: select first output in outputs for multi-output derivations

polykernel:
This commit modifies the package-func module to use the first output in
`package-func.outputs` or the default output (if the attribute `outputSpecified`
is true) as the top-level output in the final package instead of throwing
an error for multi-output derivations. Documentation for structure of the
assembled package and the selection criterion for the top-level output added
to the README.md under the module directory. In addition, a changelog entry
was appended to CHANGELOG.md to record this change.

### DIFF
--- a/modules/dream2nix/package-func/default.nix
+++ b/modules/dream2nix/package-func/default.nix
@@ -8,13 +8,15 @@
 }: let
   l = lib // builtins;
 
+  rawPackage = config.package-func.result;
+
   # outputs needed to assemble a package as proposed in
   #   https://github.com/NixOS/nix/issues/6507
   outputs = l.unique config.package-func.outputs;
 
   outputDrvs =
     l.genAttrs outputs
-    (output: config.package-func.result.${output});
+    (output: rawPackage.${output});
 
   outputPaths = l.mapAttrs (_: drv: "${drv}") outputDrvs;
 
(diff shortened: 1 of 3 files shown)


modules/dream2nix/package-func/default.nix
```diff
@@ -41,9 +27,7 @@
     // {
       inherit outputs;
       inherit config extendModules;
-      drvPath = throwIfMultiDrvOr outputDrvs.out.drvPath;
-      outPath = throwIfMultiDrvOr outputDrvs.out.outPath;
-      outputName = throwIfMultiDrvOr outputDrvs.out.outputName;
```
DavHau, line 46: Shouldn't we keep these checks to prevent people from accidentally mishandling multi-derivation packages?

polykernel: I thought about this for a bit. We can do that with the error demoted to a warning, but it would be hard to remove this warning in the future without introducing surprises. My thinking goes that since the module errors on multi-output derivations as of the status quo anyways, we can document the behavior and do without the guardrails. The change shouldn't affect any current code that depends on this behavior since they have had to explicitly reference a particular output in order to evaluate.

DavHau:
Thank you!

PR #1095, opened 2025-03-14
feat(rust-crane): add testFlagsExtra to interface

snpschaaf:
I extended the interface to access cranes setting `cargoTestExtraArgs` . Users can do so now via `testFlagsExtra` (named like `testFlags` does for `cargoTestFlags`).


modules/dream2nix/rust-crane/interface.nix
```diff
@@ -65,6 +65,11 @@ in {
       description = "Flags to add when testing the crate";
       default = [];
     };
+    testFlagsExtra = {
+      type = t.str;
+      description = "additional flags to be passed in the cargoTestCommand invocation";
+      default = "--no-run";
```
snpschaaf, line 71: TODO:FIXME: this default should not be streamlined. 

crane expect a string here. Without default everyone needs to set it. With some default some crane default would be overwritten. Make this interface optional.

Issue #1093, opened 2025-03-07
PDM - exclude certain group


h0nIg:
Hi,

we currently use the PDM default group and would like to enhance our lock file with different additional groups (one of them being a dev group).

We were wondering what the desired approach without code duplication would look like. We would like to have 1 full-fledged pdm module including default+dev and an additional one just with the dev group being removed. Maybe do we search on the wrong examples?

PR #1106, opened 2025-06-13
pdm: group name

h0nIg:
the advantage of PDM and its single lockfile is a single point of truth, for multiple architectures and multiple use cases (groups).

The groups feature should get leveraged, in order to have different types of python, e.g. python used for tests, python used for runtime only.

Therefore the group which should get used should be configurable. Closes https://github.com/nix-community/dream2nix/issues/1093

modules/dream2nix/python-pdm/tests/packages/python-local-development:

image


DavHau:
Thanks
If you find the time it would be lovely if you could add some docs (aka, description field in mkOption).

PR #1112, opened 2025-06-26
pdm: add example to create full python env

h0nIg:


DavHau (commented):
Looks good! Should we just replace the `python-local-development-pdm` with this one? I guess one `better` example is better than two.

Also we should remove the flake.lock before merging this. (I know some of the example have one,  but they should not. We still need to add a CI check for that).

> the extension via extendModules

Did you intend to make that part of your example? Because I don't see it. (Just asking)

PR #1121, opened 2025-07-18
mkderivation: add runtimedependencies for autopatchelfhook

h0nIg:


h0nIg:
can get set through env.runtimeDependencies, not required anymore

PR #1130, opened 2025-09-11
fix(rust-crane): update crane to v0.21.0

NeuralModder:
For some reason with crane v0.19.0 at the "stripping references to cargoVendorDir" step, sed just kept gobbling up memory until OOM killer ended it. And for some reason this fixed it.


NeuralModder:
@DavHau Any chance this can get merged? Idk if there's anything else in the repo that needs to be changed for this

NeuralModder:
This actually breaks the examples and I simply cannot be arsed to figure out why. Plus, users can just pass in their own version of crane, so they can just update it themselves

